PR #1543, opened 2015-01-05
Use Filters As Returned By GeoServer

tojofo:
This PR links portal to GeoServer's that have the layer filter extension installed.

Filters are configured in GeoServer and requested directly by portal.


danfruehauf:
I merged/modified/incorporated @tojofo 's changes in the original no_db branch which is maintained. So closing this for now.

PR #1563, opened 2015-01-21
Revert "Return of snazo UI"

anguss00:
Reverts aodn/aodn-portal#1562


julian1:
Goodbye snazo. sniff..

anguss00:
arnold

Issue #1517, opened 2014-11-30
Non ncWMS Filters do not load if Step 1 or 3 accessed before they load


kereid:
Instance name: IMOS
            App version: 3.36.3
            Build date: Thu Nov 27 14:31:24 EST 2014
            Build number: 282
            Git ref: 96ed7befea521587e2a1631957f9d8445da26a2d

Steps to reproduce:

Add any non-ncWMS collection.

AS soon as taken to step 2 click back to step 1 or forward to step 3.

Then go back to step 2

What should happen:

Spatial and Temporal extents are successfully loaded.

What does happen:

Just see a blank space where they should be

pblain:
@pmbohm agreed.  I'll add the accordion feature to the backlog for hopeful inclusion next iteration.  I'll close this bug because we are not going to fix it directly.

pblain:
Accordion feature alone will not fix. Re-opening.

PR #1564, opened 2015-01-23
#1517 - Only layout the filters panel when it is displayed

jonescc:


danfruehauf:
I'll look at this.

danfruehauf:
Works a treat. I think though we can add a test here. Something along the lines that `doLayout` is called only if `isDisplayed()` is true. What do you think?

jonescc:
So you want me to test the if statement works - isn't that a given.

PR #1571, opened 2015-01-29
FacetFilter is a better desciption of this filter, term doesn't really m...

jonescc:
...ake sense

Preparatory work for adding a category selection tree panel


danfruehauf:
Is it just a search & replace thing?

jonescc:
Yes.

PR #1578, opened 2015-02-04
Version CSS files to force re-download with new versions

### LINKED ISSUE #1577: CSS file imports should be versioned like our javascript files are
**Steps to reproduce**
Visit a Portal which has been updated to 'Snazo'.

**What happens**
If your browser has cached the CSS files then the Portal will not appear correctly (e.g. solid grey background on Step 1)

**What should happen**
CSS includes should be versioned the same way we version JS files so that a new copy is always downloaded when we release a new Portal.

dnahodil:
Fixes #1577


pmbohm:
Nice

PR #1588, opened 2015-02-06
Fix new search issue introduced by drilldown panel refactor

jonescc:
Need to use dolayout if changing components on a visible panel


danfruehauf:
I want to click the merge button, but something stops me. Can you show me where this situation happens?

danfruehauf:
Every time someone uses `doLayout` god kills a kitten.

Issue #1591, opened 2015-02-16
Faceted search - Don't show + when the only options will be those you've already selected


dnahodil:
**Steps to reproduce**
On step 1 select a facet from the Parameter
Click on the `+` icon to select from that facet again
Keep adding more new selections to the facet
Click on the `+` icon to select from that facet again

**What does happen**
You can keep adding to the facet even though the only option(s) is the one you've already selected

**What should happen**
You shouldn't be able to add more selections to the facet when you'll only be able to select options you have already selected

danfruehauf:
I think that shouldn't be tagged as a bug, but rather implemented correctly at the first place, no?

dnahodil:
An idea:
Should we also not include any selection(s) you've already made from the list when adding a new selection. e.g. if you select 'Temperature of the Water Body' and then add a new selection to the facet then you shouldn't see 'Temperature of the Water Body' in the second selection list.

dnahodil:
@danfruehauf It's a bit unclear. The work meets the acceptance criteria, but we've found a problem which needs to be fixed.

jonescc:
How about I remove all useless options (drilldowns which won't result in a narrowing of the search) and only display the add button when there are useful options?

jonescc:
Refer https://github.com/aodn/aodn-portal/compare/remove-useless-options?expand=1 for how this would work.

PR #1592, opened 2015-02-18
#1591 - Remove previously selected drilldowns on new drilldown panels.  ...

jonescc:
...Disable add button if no drilldowns will be available if selected

### DIFF
--- a/web-app/js/portal/prototypes/Ext.js
+++ b/web-app/js/portal/prototypes/Ext.js
@@ -15,7 +15,7 @@ Ext.tree.TreeNode.prototype.toValueHierarchy = function() {
             values.push(encodeURIComponent(p.attributes.value));
         }
         p = p.parentNode;
-    } while (p)
+    } while (p);
 
     return values.reverse().filter(function(n) {
         return n;
@@ -34,5 +34,16 @@ Ext.tree.TreeNode.prototype.clone = function(recursive) {
     }
     
     return clone;
-}
+};
 
+Ext.tree.TreeNode.prototype.getHierarchy = function(attribute) {
+    var values = [];
+    var currentNode = this;
+    do {
+        if (currentNode.attributes[attribute]) {
+            values.unshift(currentNode.attributes[attribute]);
+        }
+        currentNode = currentNode.parentNode;
+    } while (currentNode);
+    return values;
+};
--- a/web-app/js/portal/search/FacetDrilldownPanel.js
+++ b/web-app/js/portal/search/FacetDrilldownPanel.js
@@ -112,6 +112,7 @@ Portal.search.FacetDrilldownPanel = Ext.extend(Ext.tree.TreePanel, {
         this._hideSiblings(checkedNodes);
         this._expand(checkedNodes);
         this._removeCount(checkedNodes);
+        this._hidePreviousDrilldowns();
     },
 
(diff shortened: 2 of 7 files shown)


danfruehauf:
I'm reviewing.

web-app/js/portal/search/FacetDrilldownPanel.js
```diff
@@ -169,5 +170,20 @@ Portal.search.FacetDrilldownPanel = Ext.extend(Ext.tree.TreePanel, {
                 node.setText(node.attributes.value);
             }
         });
+    },
+
+    _hidePreviousDrilldowns: function() {
+        var recordCount = this.searcher.getCount();
```
jonescc: This is a hangover from a previous implementation - it should be removed along with other getCount related code

web-app/js/portal/prototypes/Ext.js
```diff
@@ -34,5 +34,16 @@ Ext.tree.TreeNode.prototype.clone = function(recursive) {
     }
     
     return clone;
-}
+};
 
+Ext.tree.TreeNode.prototype.getHierarchy = function(attribute) {
```
danfruehauf: @jonescc I'm still looking at the code, however I wanted to ask why is this overridden at the Ext level. Wouldn't it be better to inherit from that class and then override?

This means that if TreeNode is used anywhere else - you get this non-standard behaviour as well.

web-app/js/portal/service/CatalogSearcher.js
```diff
@@ -87,6 +87,14 @@ Portal.service.CatalogSearcher = Ext.extend(Ext.util.Observable, {
         }
     },
 
+    getCount: function() {
+        var summaryNode = this._searchResultRootNode.findChildBy(function(node) {
+            return node.attributes.tagName == 'summary';
+        }, this, true);
+
+        return summaryNode?summaryNode.attributes.count:0;
```
danfruehauf: Please space a bit this ternary operator...

danfruehauf: `return summaryNode ? summaryNode.attributes.count : 0;`

web-app/js/portal/prototypes/Ext.js
jonescc: I think a subclass of TreeNode would be cleaner from the treenode's point of view  but overriding the loader createNode method to create the subclassed node is not very clean (would need to expose the default treeloader processing).

This is adding an additional 'getter' style method that won't mutate the object so I don't think this is a big issue.

danfruehauf: Cool with me. Mind if we invite a couple of others to take a look at before final decision?

Pretty sure @dnahodil and @jkburges will have an opinion. Guys?

jkburges: Sorry to disappoint, but for once, I do not have an opinion on this :-D

dnahodil: Yeah, it's an interesting one. Because the behaviour is generic (so it could be meaningful somewhere else), and because things get messy if you did want to isolate it to a subclass, I'm not worried about adding this new behaviour to TreeNode.

:+1:

danfruehauf: @jkburges @dnahodil Thanks guys.

PR #1594, opened 2015-02-18
Ncwms - dates will be updated as soon as user types a valid date

pmbohm:
Updates are done as soon as a valid date is typed into the date field requiring the enableKeyEvents = true flag set. The timefield remains as is, as its not editable, only selectable.


jonescc:
I'll review

jonescc:
Causes some anomalies when editing dates manually.

For example, select a from date from the picker, and then start backspacing from the end.  Once the month my changes are replaced with another date.

pmbohm:
That is current behaviour also, although it requires the user to to click outside the box to see the behaviour. @jonescc

pmbohm:
Closing as checking on each change causes more problems than it solves

PR #1601, opened 2015-02-26
No db filters prototype

jkburges:
**DO NOT MERGE** - just PRifying so that I can add comments.


grails-app/conf/Config.groovy
```diff
@@ -86,7 +86,7 @@ downloadAuth {
 }
 
 featureToggles {
-    // someFeature = <boolean>
+    dynamicGeoserverFilters = false
```
jkburges: Do we need a toggle?  I'd say no so long as all the geoservers are updated before the portal (which is likely).

src/groovy/au/org/emii/portal/wms/GeoserverServer.groovy
```diff
@@ -66,6 +72,48 @@ class GeoserverServer extends WmsServer {
...
+
+        try {
+            def xml = new XmlSlurper().parseText(_getFilterValuesXml(server, layer, filter))
+
+            if (xml.value.size() > 0) {
+                values = xml.value.collect() { it -> it.text() }
+            }
+        }
+        catch (e) {
+            log.error "Unable to parse filters values for server '${server}', layer '${layer}', filter '${filter}': '${e}'"
```
jkburges: There's a method for logging exception and msg as two separate params, i.e.:
```groovy
log.error('some msg', e)
```

danfruehauf: That's a comment for me I think, on the `no_db` branch. I'll fix it.

web-app/js/portal/filter/BooleanFilterPanel.js
```diff
@@ -41,8 +41,7 @@ Portal.filter.BooleanFilterPanel = Ext.extend(Portal.filter.BaseFilterPanel, {
     },
 
     _formatBoxLabel: function() {
-        var label = this.filter.label.split('_').join(' ').toTitleCase();
-        return label;
+        return this.filter.getDisplayLabel();
```
jkburges: I think that `getCQL` (below) should be moved from the `FilterPanel` classes to the `Filter` class. The `FilterPanel` would need to somehow update the `Filter.value` appropriately.

web-app/js/portal/filter/BooleanFilterPanel.js
```diff
@@ -83,6 +82,10 @@ Portal.filter.BooleanFilterPanel = Ext.extend(Portal.filter.BaseFilterPanel, {
         this.checkbox.setValue(false);
     },
 
+    needsFilterRange: function() {
+        return false;
+    },
+
     _setExistingFilters: function() {
         this.re = new RegExp(this.filter.name + " = (.*?)( |$)");
```
jkburges: What the actual **** are we trying to do here?

web-app/js/portal/filter/CollectionFilters.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+Ext.namespace('Portal.filter');
+
+Portal.filter.CollectionFilters = Ext.extend(Object, {
+
+    constructor: function(collection) {
+
+        this.collection = collection;
+        this.filterObjects = [];
```
jkburges: I'd prefer to just call this `filters`, rather than `filterObjects`.

web-app/js/portal/filter/CollectionFilters.js
```diff
@@ -0,0 +1,50 @@
...
+
+    loadFilters: function(callbackFunction, callbackScope) {
+        alert("bloody collection filters fuck off");
+
+        var filterService  = new Portal.filter.FilterService();
+
+        filterService.getFilters(this.collection, this._callback, this);
+    },
+
+    _callback: function(scope, filterDetails) {
```
jkburges: `onFiltersLoaded`?   `_callback` is not very descriptive.

web-app/js/portal/filter/Filter.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+Ext.namespace('Portal.filter');
+
+Portal.filter.Filter = Ext.extend(Object, {
+
+    constructor: function(filterDetail, layer) {
+
+        this.layer = layer;
+        this.filterName = filterDetail.name;
```
jkburges: What is filterName - would `id` be more descriptive?  Also, can we drop `filter` from this (just have `name`) since it's obvious it's a filter name from the context.

web-app/js/portal/filter/Filter.js
```diff
@@ -0,0 +1,64 @@
...
+        this.displayLabel = filterDetail.label;
+        this.filterType = filterDetail.type;
```
jkburges: Same here, and method names below.

web-app/js/portal/filter/BaseFilterPanel.js
```diff
@@ -84,6 +84,14 @@ Portal.filter.BaseFilterPanel = Ext.extend(Ext.Panel, {
     },
 
     _setExistingFilters: function() {
+    },
+
+    needsFilterRange: function() {
```
jkburges: I'm a bit unsure about this... I would've thought that the range/set of possible values is a property of the `Filter` itself?  Obviously this has implications for the UI, but can it be derived from the Filter rather than coded here?

dnahodil: I see this the other way around. We currently have one usage of filter possible values - the UI. So we ask the UI "Do you need to know the range of values?". If the answer is 'no' then we don't need to bother getting them. If we had some other component that might need the possible values then we could ask that, too. That way we only fetch the possible values (which is an expensive operation) if something needs them.

dnahodil: Oh, maybe this is unclear. The filter UI is notified using the setFilterRange(range) after he range is loaded. The needsFilterRange() method just lets you know whether the UI is going to use the range or not.

jkburges: I see your point.

grails-app/conf/Config.groovy
dnahodil: Do we maybe need to have a configuration per server saying whether or not it has this plugin installed? That way we'd only send requests to servers we know can respond to them.

web-app/js/portal/filter/BooleanFilterPanel.js
dnahodil: This is old code for parsing values out of CQL. This will be killed with the advent of the new Filter model.

grails-app/conf/Config.groovy
danfruehauf: Per server is the way to go IMO. I'll fix that on the `no_db` branch.

However we might also want to enable filters from plain simple files, it can still be useful!

web-app/js/portal/filter/BooleanFilterPanel.js
dnahodil: I'm actually thinking that there will be somewhere else that has a function that takes a Filter and turns it into CQL. I've been thinking about Filters as a data structure rather than as an object with behaviour/business logic.

jkburges: It's a good idea to have a class separate to `Filter` for constructing CQL.  Anyway, I think we both agree it definitely doesn't belong in the UI.

dnahodil: Yep

web-app/js/portal/filter/CollectionFilters.js
dnahodil: We should have got rid of this. This class is actually no longer being used.

web-app/js/portal/filter/Filter.js
dnahodil: I agree that all these things can lose the 'filter' prefix.

GeoServer calls them "Layer Name" in its UI. But I like that "id" lets you know this is the unique identifier ("name" doesn't).

dnahodil: We've actually been a little inconsistent in both JS and Groovy land. We should tidy this up as we do this work.

src/groovy/au/org/emii/portal/wms/GeoserverServer.groovy
```diff
@@ -10,6 +10,12 @@ package au.org.emii.portal.wms
 import au.org.emii.portal.proxying.ExternalRequest
 
 class GeoserverServer extends WmsServer {
+    def dynamicFilters
+
+    GeoserverServer(dynamicFilters) {
+        this.dynamicFilters = dynamicFilters
```
dnahodil: Yeah we'll configure GeoServer filter plugins per server and get rid of this here.

web-app/js/portal/filter/BooleanFilterPanel.js
dnahodil: I also envisage a similar class which takes a Filter and returns html for a human-readable representation for Step 3.

jkburges: ... or serialises a filter in to a URL key/value param - so that we can then construct links which includes filters....

dnahodil:
I'm just going to close this PR to avoid any confusion. Our work on https://github.com/aodn/aodn-portal/tree/filters_from_geoserver replaces this prototype.

anguss00:
:+1:

PR #1605, opened 2015-03-04
CSV output format for downloads is now configured with a string.

jkburges:
This enables the possibility of using different output formats for different geoservers.

Depends on https://github.com/aodn/chef-private-sample/pull/8 and https://github.com/aodn/chef-private/pull/1348

**After** this change is deployed, then https://github.com/aodn/chef-private-sample/blob/csv_output_config/data_bags/imos_webapps_portal_known_servers/geoserver_123_aodn.json#L7 and likewise can be deleted.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -198,7 +198,7 @@ knownServers = [
         uri: '', // The server URL (required) -- I would prefer to see that is url rather than uri but I have it as uri for now to match the field in the Server domain class.
         wmsVersion: '', // The version number as a string e.g. "1.1.1", "1.3.0" (not sure if we should make this required or not)
         type: '', // Identifying the specific server software. e.g. "nvWMS", "GeoServer". If omitted or blank then it is considered to just be a general WMS server without any extra functionality. Should be case insensitive.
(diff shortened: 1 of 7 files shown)


danfruehauf:
I'll have a look shortly.

danfruehauf:
@jkburges I did merge this, but noticed a problem with that afterwards - the size estimation ceased to work. And that is quite obvious - it's because it's asking for `csv-with-metadata-header` when estimating size. This download format doesn't have the URL field. So... breaks :/

jkburges:
Thanks for picking this bug up @danfruehauf - I've reverted this (https://github.com/aodn/aodn-portal/commit/1d0943463bca5bdbf038c8d621be3f0b14fa2727) and will re-attempt this change.

jkburges:
@danfruehauf or someone else, it's been a while since I've done any portal development...and things have changed a bit.   What is a suggested handful of collections to test against to check that I haven't broken anything in this area?

danfruehauf:
@jkburges Anything with CSV - so... argo et al

jkburges:
I'm more worried about breaking gogoduck or some such...  I've tested CSV and it seems ok to me.

danfruehauf:
Experiment with an ACORN layer and gogoduck then. But I don't think it should break...

PR #1606, opened 2015-03-04
Issue 1603

### LINKED ISSUE #1603: interface unresponsive with ncwms layers
Steps to repeat:
 * Select a NcWMS layer from CSIRO (srs for instance)
 * Change dates

What should happen:
 * Interface is responsive (user can interact with other controls) while dates are being fetched

What happens:
 * Interface is not responsive

NOTE: This is apparent because the latency between our portal and CSIRO is not amazing. Happens as well when choosing for instance ACORN layers, but it's not as apparent.

danfruehauf:
Fixes #1603

The issue became apparent as SRS layers have many days defined (1992-2015). In particular the functions `getFirstDay()` and `getLastDay()` were scanning a big hash. They were modified to return a fixed variables which is calculated every time a new date is being added. This is addressed at a6160c7.

Another problem was the inline addition of dates. When many dates are defined this proves to be a problem. This is now done in an interleaved manner which doesn't jam the UI. This is addressed in 924a871.

Altogether the UI experience is **much** snappier now when choosing NcWMS layers and it is very apparent.

### DIFF
--- a/src/test/javascript/portal/visualise/animations/TemporalExtentSpec.js
+++ b/src/test/javascript/portal/visualise/animations/TemporalExtentSpec.js
@@ -91,6 +91,26 @@ describe("Portal.visualise.animations.TemporalExtent", function() {
         });
     });
 
+    describe('_createDay', function() {
+        var aMomentInTime = moment.utc('2000-01-01T01:00:00.000');
+        var calledWith = null;
+
+        beforeEach(function() {
+            spyOn(temporalExtent, '_setFirstDay').andCallFake(function(date) { calledWith = date; });
+            spyOn(temporalExtent, '_setLastDay').andCallFake(function(date) { calledWith = date; });
+            temporalExtent._createDay(aMomentInTime);
+        });
+
+        it('calls _setFirstDay', function() {
+            expect(temporalExtent._setFirstDay).toHaveBeenCalled();
+            expect(aMomentInTime).toBeSame(calledWith);
+        });
+
+        it('calls _setLastDay', function() {
+            expect(temporalExtent._setLastDay).toHaveBeenCalled();
+            expect(aMomentInTime).toBeSame(calledWith);
+        });
+    });
 
     describe('getDay', function() {
         it('return null when empty', function() {
@@ -248,11 +268,12 @@ describe("Portal.visualise.animations.TemporalExtent", function() {
 
         it('returns all defined dates in array', function() {
             temporalExtent.parse(['2000-01-01T23:00:00.000/2000-01-02T01:00:00.000/PT30M']);
-            expect(temporalExtent.getExtentAsArray()[0].toISOString()).toBe('2000-01-01T23:00:00.000Z');
-            expect(temporalExtent.getExtentAsArray()[1].toISOString()).toBe('2000-01-01T23:30:00.000Z');
-            expect(temporalExtent.getExtentAsArray()[2].toISOString()).toBe('2000-01-02T00:00:00.000Z');
-            expect(temporalExtent.getExtentAsArray()[3].toISOString()).toBe('2000-01-02T00:30:00.000Z');
-            expect(temporalExtent.getExtentAsArray()[4].toISOString()).toBe('2000-01-02T01:00:00.000Z');
+            var flatArrayOfDates = temporalExtent.getExtentAsArray();
+            expect(flatArrayOfDates[0].toISOString()).toBe('2000-01-01T23:00:00.000Z');
+            expect(flatArrayOfDates[1].toISOString()).toBe('2000-01-01T23:30:00.000Z');
+            expect(flatArrayOfDates[2].toISOString()).toBe('2000-01-02T00:00:00.000Z');
+            expect(flatArrayOfDates[3].toISOString()).toBe('2000-01-02T00:30:00.000Z');
+            expect(flatArrayOfDates[4].toISOString()).toBe('2000-01-02T01:00:00.000Z');
         });
     });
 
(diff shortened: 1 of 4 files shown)


jkburges:
I'll take a look at this.

src/test/javascript/portal/visualise/animations/TemporalExtentSpec.js
```diff
@@ -91,6 +91,26 @@ describe("Portal.visualise.animations.TemporalExtent", function() {
...
+        });
+
+        it('calls _setFirstDay', function() {
+            expect(temporalExtent._setFirstDay).toHaveBeenCalled();
+            expect(aMomentInTime.toDate()).toEqual(calledWith.toDate());
+        });
+
+        it('calls _setLastDay', function() {
+            expect(temporalExtent._setLastDay).toHaveBeenCalled();
+            expect(aMomentInTime.toDate()).toEqual(calledWith.toDate());
```
jkburges: Could you use `expect(aMomentInTime).toBeSame(calledWith);` here?  `toBeSame()` is a bit vaguely named - it probably should be `toBeSameMoment()`

jkburges: Just noting also that this package name is not at all what you'd expect - `Portal.visualise.animations`, I think it should be `Portal.visualise.something` but not sure what the `something` should be.  What do you think?  (btw not expecting you to necessarily do anything about this in this PR).

danfruehauf: I'll change that test, yeah. Equality doesn't work for moments for some reason.

Speaking of the package names - yeah, messed as...

pmbohm: the moment objects may differ, but you could compare strings?
 expect(aMomentInTime.utc()).toBeSame(calledWith.utc())

danfruehauf:
@jkburges Addressed the comment there. Take a look at SRS, it's heaps snappier, move forward/backward in time slices.

web-app/js/portal/ui/openlayers/layer/NcWMS.js
```diff
@@ -102,9 +102,7 @@ OpenLayers.Layer.NcWMS = OpenLayers.Class(OpenLayers.Layer.WMS, {
         });
     },
 
-    _timeSeriesDatesLoaded: function(datesWithData) {
```
jkburges: It took me some time to figure out what all the similarly named functions do in this class, listed here:

```
_fetchTimeSeriesForDay
_getTimeSeriesUrl
_loadTimeSeriesDates
_timeSeriesDatesLoaded
_timeSeriesLoadedForDate
getTimeSeriesForDay
loadTimeSeriesForDay
```

Firstly, I just wonder if this could be made a bit more understandable by doing the following:

* change `timeSeriesDates` simply to `Days`, e.g. `_loadTimeSeriesDates` becomes `loadDays`.  
* similarly, change `timeSeriesForDay` to `timesForDay`... 

This is because I think `timeSeries` is a bit redundant in the names, and `Date` is a bit ambiguous, in this context.

Secondly, correct me if I'm wrong here, but the general algorithm is now:

1. Fetch all the days that the layer covers
2. Fetch all the times on the *first* and the *last* day

Could this algorithm be more clearly expressed by bringing in a new class which sits between `NcWms` and `TemporalExtent`, and deals solely with this stuff?   It's just a bit confusing atm, IMO.

src/test/javascript/portal/visualise/animations/TemporalExtentSpec.js
jkburges: What *would* be an appropriate package name though?  Basically we're dealing with collections that have a time dimension...  `Portal.visualise.temporal`?

jkburges:
I can verify the UI is much snappier now btw, I checked that first.  So, I'm happy that this fixes #1603.

web-app/js/portal/ui/openlayers/layer/NcWMS.js
danfruehauf: I actually think the function names make perfect sense. It is just a bit more complicated than the average filter as things are being done asynchronously + processing is mostly asynchronous. I had a second look at the function names and I could easily tell what each does. We can rename however `timeSeries` to just `times`, that can work.

I think we can do many things here. Generally speaking I'd like to see all of that factored into a normal filter, rather than this bastardization of NcWMS panel etc.

That, and the package name which I don't think is appropriate anymore.

On the bright side of things - it's stateless and relatively "bug free" so far.

danfruehauf: About decoupling the first/last day into a different class, I think it should be decoupled in such a way that the interface will decide what to fetch s.t. the interface can fetch all dates, then decide what days to fetch. It will not be a 5 minute job unfortunately...

jkburges: > I actually think the function names make perfect sense...

Well, that may be the case for someone familiar with the code, but just saying it took me about 30 minutes of reviewing this class to figure out what was going on here, so that suggests that some refactoring might be in order (more moving things in to a dedicated class rather than just renaming, I think).

As always, I'm ok to leave it as is, since it's a matter of opinion.

jkburges:
So, I'm good to merge @danfruehauf, you GTG?

web-app/js/portal/ui/openlayers/layer/NcWMS.js
danfruehauf: I'm with you. At the time of doing that work it was a bit of a stretch even getting done what was required - at that time frame.

Portal code needs refactoring all over the place - that's yet another one. Not sure when we can actually get to it though :/

danfruehauf:
Yeah, GTG. We'll refactor that another time :/

Issue #1121, opened 2014-05-16
Filters displayed/used twice in steps 2 and 3


anguss00:
Portal version: portal-rc.aodn.org.au, version 3.23.0-rc

Steps to reproduce: Add argo layer, draw bounding box, go to step 3

Expected behaviour: filter is displayed once on step 3 download page in the expected format.

Observed behaviour: See screenshot below, filter is displayed twice

![open_access_to_ocean_data](https://cloud.githubusercontent.com/assets/5964166/2992452/039148c4-dc92-11e3-873e-a232d9e29df2.png)

PR #1607, opened 2015-03-04
Filters from geoserver

dnahodil:
Ready for review.

This branch will be the collection of any Portal-side filters work which can't be safely merged directly into master. Work on the filters should be on a new branch branched from this and then a PR to merge back into this branch.


dnahodil:
We might need to be careful that #1121 doesn't come back. We've removed the code that fixed that, but it might be okay as we've changed some much other code around it. If it does come back the fix can be reapplied.

dnahodil:
FYI I've just rebased this onto master.

danfruehauf:
I can see the pain coming, rebasing no_db after this is merged...

dnahodil:
Unfortunately not all the commits on no_db were small enough that we could just cherry-pick the ones we needed.

danfruehauf:
@dnahodil Please cherry-pick also:
https://github.com/aodn/aodn-portal/commit/86e8dab647e8419ce4acc35eb662cab642791728

dnahodil:
Done @danfruehauf

danfruehauf:
@dnahodil Thanks!

I'll need help rebasing no_db after this is merged. It might be a bit epic :/

dnahodil:
Let's see how messy it gets, @danfruehauf. It we want to maintain the no_db branch then it might be easier cherry-picking commits from the current one onto a new branch created after this work is merged.

dnahodil:
This is ready for review now by somebody other than myself, @danfruehauf, and @anguss00 (who have all contributed or reviewed parts of this code, already).

dnahodil:
There is more refactoring happening on this work, currently, but the latest commits make it functionally complete.

jkburges:
What's the status of this PR? Is it mergeable, supposing that it's bug free and reviewer(s) is/are happy?  I can take a look at it.

dnahodil:
It is mergable. Just be aware that there is more refactoring happening (it can continue after this is merged) so things like the new Filter model have some more tweaks coming.

jkburges:
Here's a meta-comment:  this is obviously a **massive** PR, can we have a think about how this could possibly have been done in a few smaller chunks, and importantly, integrated back in to master along the way?  What caused it to become so big before being mergeable? Can we avoid the same mistakes (if you agree with me that it's a "mistake" to have big PRs)?

dnahodil:
I agree that's it's too big to really be able to be reviewed properly. There are a couple of problems as I see it.

The first is that we were reusing work that was done to replace the current behaviour (which we pulled in bits and pieces from the no_db branch) but what we're doing he is allowing both behaviours (with a feature toggle). So that made it messier than it needed to be.

Secondly, we're also paying the price for a bunch of code that has grown in many directions with lots of contributors. So there's a lot of tidy-up and moving related to that.

grails-app/conf/Config.groovy
```diff
@@ -128,7 +128,7 @@ openId {
 }
 
 featureToggles {
-    // someFeature = <boolean>
+    filterSource = 'database' // Todo - DN: This can go back to being a boolean when we've removed loading filters from the DB
```
jkburges: Non boolean "toggles" seems like a scary prospect.  What's the reason behind why this is a string value?

src/groovy/au/org/emii/portal/HostVerifier.groovy
```diff
@@ -16,91 +16,47 @@ class HostVerifier {
...
-    def retrieveResultsFromGeoNetwork(server) {
-        def catalog = grailsApplication.config.geonetwork.url
-
-        return new XmlParser().parse("$catalog/srv/eng/q?serverUrl=$server&fast=index&summaryOnly=true")
-    }
-
-    def extractServer(url) {
-        return url.protocol + "://" + url.host + url.path
+    def extractHost(url) {
+        return url.toURL().host
```
jkburges: so if it wasn't a url, was it a string?

src/groovy/au/org/emii/portal/wms/GeoserverServer.groovy
```diff
@@ -0,0 +1,217 @@
...
+        catch (e) {
+            log.error "Unable to parse filters for server '${server}', layer '${layer}'", e
+        }
+
+        return filters
+    }
+
+    def _getFiltersXml(server, layer) {
+
+        switch (filterSource) {
```
jkburges: Well, I guess this answers my question from above.  But, in this case, I don't think it's a feature toggle - it's just configuration.  A feature toggle is like a light switch - on or off.

src/groovy/au/org/emii/portal/wms/GeoserverServer.groovy
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+package au.org.emii.portal.wms
+
+import au.org.emii.portal.Layer
+import au.org.emii.portal.Server
+import au.org.emii.portal.proxying.ExternalRequest
+import groovy.xml.MarkupBuilder
+
+class GeoserverServer extends WmsServer {
```
jkburges: This class is a nice abstraction.

jkburges: Possibly the filter specific stuff can be extracted in to a separate class dealing with just that.  Then this class becomes a pretty high level interface and it's easier to see what can be done with it (without the low-level stuff providing a distraction).

grails-app/conf/Config.groovy
dnahodil: This will become a boolean when the database option is removed (currently a string because there are 3 options). But as discussed I will move it out of the feature toggles block as it is not really a feature toggle.

danfruehauf: I think leave it as is, even though it's not a toggle. no_db will clean this up, hopefully it's just "around the corner".

dnahodil: As I understand it no_db will still leave us with a boolean option, @danfruehauf. We wanted to be able to load from GeoServer of from XML files to allow more options for developing/troubleshooting. So we'll need a switch between those options.

danfruehauf: Correct, no_db should amend this to become a boolean rather than a string. Isn't that what you intended to do?

I mean, it can be factored to something like `dynamicFilters = true/false`. And then have (not as a toggle) `filterSource = 'database/files`, and `filterSource` would actually be used only if dyanmic filters is set to false - then you need "static" filters, either from database or files.

Either way, not a biggie. Even though @jkburges is concerned, I'm not very, because I know it'll be gone sooner than later.

jkburges: Just to clarify - I'd be more concerned with us doing something like this as a matter of habit, rather than with this particular occurrence.  I think we need to be very prudent with our use of feature toggles - especially w.r.t. to making them short lived, as @danfruehauf eludes to.

danfruehauf: It's a fair point. Lets clean it up now then.

web-app/js/portal/filter/Filter.js
```diff
@@ -0,0 +1,61 @@
...
+
+        this.name = filterDetail.name;
+        this.displayLabel = filterDetail.label;
+        this.type = filterDetail.type;
+        this.visualised = filterDetail.visualised;
+        this.wmsStartDateName = filterDetail.wmsStartDateName;
+        this.wmsEndDateName = filterDetail.wmsEndDateName;
+    },
+
+    setValue: function(value) {
```
jkburges: I recall discussing this @dnahodil.  I think getters are just fluff here.  We could put them back in if and when needed?

grails-app/conf/Config.groovy
dnahodil: I think we're all on the same page. I'd like to leave this as-is for this PR but when we combine master and no_db this will be changed to a general configuration option and not treated as a feature toggle.

web-app/js/portal/filter/Filter.js
dnahodil: I hated the repetition here so in a refactoring (which isn't on this branch) I pulled this all out. See https://github.com/aodn/aodn-portal/blob/ee69e19157de6b36b4d0bcedf57a265230343c5d/web-app/js/portal/filter/Filter.js#L14

So if we were to remove the accessor methods now there wouldn't be any indication of which fields are going to be present other than by searching other code for their usages. Do you think that would be a problem?

dnahodil:
@jkburges FYI I've made a review change commit (146788a) and reverted a mistake (3f58c04) and rebased the whole lot.

jkburges:
I will likely merge this, if and when https://github.com/aodn/aodn-portal/issues/1631 is resolved (which I'm about to start looking at).

As I have mentioned offline, I haven't reviewed the code very closely here, as there's so much changed. For this very reason, I'd like to get this merged sooner rather than later so that further refactors and fixes can be reviewed more thoroughly.

dnahodil:
Sounds like a plan

PR #1608, opened 2015-03-04
CSV output format for downloads is now configured with a string.

jkburges:
This enables the possibility of using different output formats for different geoservers.

It might be worth getting 2 reviewers on this PR, given that it has the potential to break some critical use cases.  I've tested it but that didn't stop me introducing major bug in the last attempt :-)


dnahodil:
I'll have a look at this, too.

PR #1610, opened 2015-03-05
Broken jasmine on geoserver filters

dnahodil:
@anguss00 this fixes a bunch of tests which we should have updated as we did our filter work.


dnahodil:
@anguss00 there are still 7 broken tests that I know about. I think all of them are about our changes to FilterGroupPanel. There's something else I need to look at now. Would you mind checking these out and seeing if you can see where the tests need to be updated to reflect our recent changes? I can have a look at this with you later in the afternoon if you have any trouble.

anguss00:
Fixed dem tests @dnahodil

dnahodil:
Good one. That all look okay, and it will get a good test as we continue our development on this branch.

PR #1611, opened 2015-03-05
Prototype for how chat could work, for users who we think haven't found...

jkburges:
... what they're looking for.

**DO NOT MERGE**

I think something like this could help to lower the bounce rate, i.e. users who come to the site but don't end up staying for long.

I reckon we could run an experiment with what's in this PR (I've only spent 30 mins on this so far), with a few small changes:

* decide on exactly when to trigger (e.g. based on timer, only trigger in business hours, only trigger once etc)
* use a dedicated IRC channel, e.g. "imos-support"
* a slightly better UI, e.g. a clickable link would be a good start :smile:
* configurable wording and chat URL

Of course, we would need to have a few people willing to monitor the channel...

Discuss.

### DIFF
--- a/web-app/js/portal/ui/Viewport.js
+++ b/web-app/js/portal/ui/Viewport.js
@@ -7,6 +7,21 @@
 
 Ext.namespace('Portal.ui');
 
+window.onbeforeunload = function() {
+
+    if (   Portal.config.isFeatureToggleEnabled('chatOnFailedSearch')
+        && (Portal.ui.MainPanel.maxStepReached < TAB_INDEX_VISUALISE)) {
+        // Replace this with a better UI...
+        return String.format(
+            "It looks like you didn't find what you were looking for. Want to chat - go to this link: {0}",
+            'http://webchat.freenode.net?channels=%23imos&uio=d4'
+        );
+    }
+    else {
+        return undefined;
+    }
+};
+
 Portal.ui.Viewport = Ext.extend(Ext.Viewport, {
     constructor: function(cfg) {
 
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -86,6 +86,8 @@ downloadAuth {
 }
 
 featureToggles {
+    // someFeature = <boolean>
+    chatOnFailedSearch = true
     dynamicFilters = true
 }
 
--- a/web-app/js/portal/config/PortalConfigLoader.js
+++ b/web-app/js/portal/config/PortalConfigLoader.js
@@ -1,4 +1,3 @@
-
 /*
  * Copyright 2013 IMOS
  *
@@ -7,6 +6,10 @@
  */
 
 Ext.namespace('Portal.config');
+
+Portal.config.isFeatureToggleEnabled = function(toggleName) {
+    return Portal.app.appConfig.featureToggles[toggleName];
+}
 
 // Loads the various configs the Portal needs and calls functions based on success or failure
 Portal.config.PortalConfigLoader = Ext.extend(Object, {
--- a/web-app/js/portal/ui/MainPanel.js
+++ b/web-app/js/portal/ui/MainPanel.js
@@ -60,6 +60,7 @@ Portal.ui.MainPanel = Ext.extend(Ext.Panel, {
     },
 
     setActiveTab: function(tabIndex) {
+        Portal.ui.MainPanel.maxStepReached = Math.max(tabIndex, Portal.ui.MainPanel.maxStepReached);
         this.layout.setActiveTab(tabIndex);
     },
 
@@ -85,3 +86,5 @@ Portal.ui.MainPanel = Ext.extend(Ext.Panel, {
         }
     }
 });
+
+Portal.ui.MainPanel.maxStepReached = TAB_INDEX_SEARCH;


web-app/js/portal/ui/Viewport.js
```diff
@@ -7,6 +7,21 @@
...
 Ext.namespace('Portal.ui');
 
+window.onbeforeunload = function() {
+
+    if (   Portal.config.isFeatureToggleEnabled('chatOnFailedSearch')
+        && (Portal.ui.MainPanel.maxStepReached < TAB_INDEX_VISUALISE)) {
+        // Replace this with a better UI...
+        return String.format(
+            "It looks like you didn't find what you were looking for. Want to chat - go to this link: {0}",
+            'http://webchat.freenode.net?channels=%23imos&uio=d4'
```
danfruehauf: What's uio=d4?

Can we also plant something like a random username? `portal_user_TOKEN`?

web-app/js/portal/ui/Viewport.js
```diff
@@ -7,6 +7,21 @@
 
 Ext.namespace('Portal.ui');
 
+window.onbeforeunload = function() {
+
+    if (   Portal.config.isFeatureToggleEnabled('chatOnFailedSearch')
+        && (Portal.ui.MainPanel.maxStepReached < TAB_INDEX_VISUALISE)) {
+        // Replace this with a better UI...
+        return String.format(
+            "It looks like you didn't find what you were looking for. Want to chat - go to this link: {0}",
```
danfruehauf: I like the first part of the sentence. but maybe for the second we can have:
"Would you like to chat with us? Follow this link."

web-app/js/portal/ui/Viewport.js
```diff
@@ -7,6 +7,21 @@
 
 Ext.namespace('Portal.ui');
 
+window.onbeforeunload = function() {
+
+    if (   Portal.config.isFeatureToggleEnabled('chatOnFailedSearch')
```
danfruehauf: Some many spaces here after the `(`

danfruehauf:
Good idea, I like!

About the phrasing, perhaps I'm not the best person to comment on that, but we can find something "friendlier".

web-app/js/portal/ui/Viewport.js
danfruehauf: Oh, and make this a label and a corresponding value in `lang/en.js`

web-app/js/portal/ui/Viewport.js
danfruehauf: This can come from portal server config, so we can set it on the server side.

jkburges: I think it's to do with the styling to give webchat... It's a generated link.

web-app/js/portal/ui/Viewport.js
jkburges: I should've noted in the OP - both the wording *and* the chat URL need to be made configurable.

jkburges: I've add this to the OP, just to be clear.

web-app/js/portal/ui/Viewport.js
jkburges: It's to align with the next line... deliberate, but perhaps not in line with the standard.

danfruehauf: Scratch that, then I'm fine with that. I do it myself quite a lot, was just a bit weird in the github viewer...

web-app/js/portal/ui/Viewport.js
danfruehauf: OK cool. I'm happy to do this work if you've fully spent your 10% time ;)

jkburges:
@pblain @kereid @jachope @smancini comments?

danfruehauf:
Note: works only on chrome :)

kereid:
As we haven't had the google analytics dojo yet, perhaps my understanding of bounce rate is incorrect... but according to GA "bounce Rate is the percentage of single-page visits (i.e. visits in which the person left your site from the entrance page without interacting with the page), therefore I have a couple of questions:

How do you catch someone that does nothing and clicks straight out? I would imagine you would still need the website open something similar to what is seen on Virgin/Jetstar or banking websites?

Will you have a time limit on when the question will be asked? Don't want every person to get a popup everytime they go on if they are pondering what they are going to search for a few seconds.

Also as we aren’t 24/7  it would need to only be “switched on” (is that even possible?) during business hours which probably is still only about 50% of the time users are on there anyway? 

This is a good idea (not sure if we do have the number of users that would warrant it, and perhaps having it will increase them) and I do understand this is a 10% time prototype but as it would directly affect me I'd be more comfortable with it going via JPSR for prioritisation and going through the whole design meeting/planning process as a new feature before going anywhere near systest/rc/production...

jkburges:
@kereid those are good questions - I have added "decide on exactly when to trigger (e.g. based on timer, only trigger in business hours, only trigger once etc)" to the OP because this is really important for this feature to be useful rather than an annoyance.  Currently, this notification is triggered simply when a user closes the window and they have not left step 1, but we could certainly implement a more intelligent algorithm here for deciding when to trigger.

I definitely think this would need to be approved by JPSR because aside from changes to the actual portal behaviour, it obviously impacts the team in other ways (e.g. having to monitor and respond to users' questions being the main one).

pblain:
@jkburges I think it's a great idea as long as we can come up with the man power to answer the inquiries. It would look pretty bad if no one answers.  As Kate says there are some issues JPSR would need to consider on that front.  But anything we can do help more users find and download data is worth considering. 

A more low tech idea would be to put our phone number on the portal.  This would have similar issues (ie who would answer the calls).  Another matter fro JPSR.  But having an obvious phone number is considered good practice if you are serious about good customer service.  Even if they don't call it, users like knowing that someone is available to help. Here's an excerpt from "Don't Make Me Think":

## Things that diminish goodwill
Here are a few of the things that tend to make users feel like the people
publishing a site don’t have their best interests at heart:

### Hiding information that I want. 
The most common things to hide are customer support phone numbers, 
shipping rates, and prices.

The whole point of hiding support phone numbers is to try to keep users
from calling, because each call costs money. The usual effect is to diminish
goodwill and ensure that they’ll be even more annoyed when they do find
the number and call. On the other hand, if the 800 number is in plain sight—
perhaps even on every page—somehow knowing that they can call if they
want to is often enough to keep people looking for the information on the site
longer, increasing the chances that they’ll solve the problem themselves.

dnahodil:
@jkburges FYI I rebased this onto master

danfruehauf:
I actually hoped to see that going in... :(

jkburges:
The work is tagged - https://github.com/aodn/aodn-portal/releases/tag/chat

PR #1612, opened 2015-03-05
Fixes #1597.

### LINKED ISSUE #1597: If no applicable top level parameters are set should not show in results
Instance name: IMOS
            App version: 3.41.2
            Build date: Thu Feb 19 09:38:07 AEDT 2015
            Build number: 357
            Git ref: e9cec741231a49f2774f7397860b0be6cbdbb353

Steps to reproduce:

Select biological platform. Look at list of results

What should happen:

If there is no applicable parameter set then the row should not be seen

What does happen:

Parameter symbol of that which is not set show but with no text

symbol only

jkburges:
If a collection does not have any measured parameters, show a message indicating such, as opposed to a blank string.

### DIFF
--- a/src/test/javascript/portal/search/FacetedSearchResultsDataViewSpec.js
+++ b/src/test/javascript/portal/search/FacetedSearchResultsDataViewSpec.js
@@ -114,4 +114,25 @@ describe("Portal.search.FacetedSearchResultsDataView", function() {
             expect(window.trackUsage).toHaveBeenCalledWith("Collection", "select", "Argo Australia Profiles");
         });
     });
+
+    describe('_getMeasuredParametersAsCommaSeparatedString', function() {
+
+        beforeEach(function() {
+            params = [];
+
+            facetedSearchDataView._getMeasuredParameters = function() {
+                return params;
+            };
+        });
+
+
+        it('with some parameters', function() {
+            params = ['temp', 'salinity'];
+            expect(facetedSearchDataView._getMeasuredParametersAsCommaSeparatedString()).toEqual('temp, salinity');
+        });
+
+        it('with no parameters', function() {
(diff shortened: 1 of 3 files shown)


src/test/javascript/portal/search/FacetedSearchResultsDataViewSpec.js
```diff
@@ -114,4 +114,25 @@ describe("Portal.search.FacetedSearchResultsDataView", function() {
             expect(window.trackUsage).toHaveBeenCalledWith("Collection", "select", "Argo Australia Profiles");
         });
     });
+
+    describe('_getMeasuredParametersAsCommaSeparatedString', function() {
+
+        beforeEach(function() {
+            params = [];
```
dnahodil: An accidental global variable. Needs 'var' keyword.

Same again, below.

web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -132,13 +132,24 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
...
                 "label": label,
-                "value": this.getMeasuredParameters(values)
+                "value": this._getMeasuredParametersAsCommaSeparatedString(values)
             });
         }
         return "";
     },
 
-    getMeasuredParameters: function(values) {
+    _getMeasuredParametersAsCommaSeparatedString: function(values) {
```
dnahodil: Maybe consider calling this `_getMeasuredParametersText()` or something? It's a touch shorter and we wouldn't have to change it if we changed symbols again (we used to separate with pipes '|').

dnahodil:
Changes look good @jkburges. Just a couple of comments.

jkburges:
Updated @dnahodil

PR #1614, opened 2015-03-06
Updating URL when collections are added/removed.

jkburges:
This is to make it easier for people to share links which will result in a portal with already loaded collections.

Some more 10% action - this *could* be extended further to include serialisation of filter state etc, but anyway, I think this would provide a bit of value to users.

### DIFF
--- a/grails-app/views/_js_includes.gsp
+++ b/grails-app/views/_js_includes.gsp
@@ -149,6 +149,7 @@
     <script type="text/javascript" src="${resource(dir: 'js/portal/data', file: 'GeoNetworkRecordFetcher.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/data', file: 'GeoNetworkRecordStore.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/data', file: 'ActiveGeoNetworkRecordStore.js')}"></script>
+    <script type="text/javascript" src="${resource(dir: 'js/portal/data', file: 'UrlUpdater.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/ui', file: 'EmptyDropZonePlaceholder.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/mainMap', file: 'map.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/common', file: 'GeoExt.ux.BaseLayerCombobox.js')}"></script>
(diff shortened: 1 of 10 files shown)


danfruehauf:
Works well, but a couple of things:
 * Mr. Travis is not happy
 * If I touch the URL in the address bar I can break the behavior of it, is that by design?

jkburges:
> Mr. Travis is not happy

I'll take a look at that - worked for me!

jkburges:
Btw, this feature will only work for HTML5 - is that going to be ok? Or should we make it at least not baulk for HTML4?

dnahodil:
I would say it needs to be HTML4 safe, even if they don't get the functionality. Any idea which versions of Chrome, FF, IE, and Safari are the earliest to have HTML5 support?

jkburges:
@dnahodil I'd say we're pretty safe actually: http://caniuse.com/#search=pushstate

dnahodil:
Yeah, that looks pretty good

jkburges:
Travis happy now.

jkburges:
... oh and bug fixed in 6065af3

danfruehauf:
Squash commits?

danfruehauf:
The fix doesn't quite work for me. If I play with the URL at the address bar it still doesn't behave properly (i.e. not modifying the address)

jkburges:
> The fix doesn't quite work for me. If I play with the URL at the address bar...

image

danfruehauf:
Works in chrome, in ffox I'm still able to break it, I can demonstrate how.

jkburges:
Ok, I'll check it tomorrow. Good pick up!

jkburges:
I can't break it in FF 28.0 - what version are you using?  What are the exact steps to break it (or demo to me)?

jkburges:
I asked the question about whether firefox should update the URL in the case that a user has edited the address but not hit return in #firefox IRC. 

<blockquote>
jkburges: considering a page which updates the URL using HTML 5's window.history.pushState(), ala https://gist.github.com/jkburges/17f6028059dfb65c3cf9 …  when a user edits the address in the address bar, *does not* hit enter but removes focus, should any subsequent changes to the URL via pushState() be reflected in the URL?  In chrome, changes *are* shown, but not in firefox.  One of my colleagues has asked me to explain what the correct behaviour should be.<br/><br/>
13:53	ChrisMorgan	jkburges: there is no such thing as “correct” there. If the user goes to the address bar again and presses escape, it will go back to reflecting the current URL, but there are arguments in favour of both approaches.
</blockquote>

I tend to agree with what Chris says - there's no right or wrong in this case.  Certainly chrome and firefox differ, but there's not much in that.

jkburges:
Commits squashed btw.

danfruehauf:
Where did you ask that?

Don't forget the beer.

jkburges:
> Where did you ask that?

https://client02.chat.mibbit.com/?server=irc.mozilla.org%3A%2B6697&channel=%23firefox

> Don't forget the beer.

I won't, but it'll have to be next week perhaps.

PR #1615, opened 2015-03-06
Geoserver filters feature toggle

dnahodil:
This adds a feature toggle to select between loading filters from the database and loading them from XML files/GeoServer. For now we use XML files (ie. not GeoServer) but that can be easily changed when the new GeoServer plugin changes have been made.

### DIFF
--- a/get_filters.rb
+++ b/get_filters.rb
@@ -1,4 +1,4 @@
-#!/usr/bin/ruby
+#!/usr/bin/env ruby
 
 require 'nokogiri'
 require 'open-uri'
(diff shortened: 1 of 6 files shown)


dnahodil:
Still has some println()s in there but I can grab them out later.

anguss00:
Looks good @dnahodil :+1:

anguss00:
Actually, I've just found an issue with filterPanels not being created correctly when accessing from xml files that I'll need to ask you about on tuesday @dnahodil

dnahodil:
@anguss00 I've pushed up a change to the get_filters.rb script which should fix the type problem you found. It will still be broken until we do the conversion from GeoServer -> Portal filter types. That work is already started on another branch so we should see it soon.

anguss00:
Ok no worries @dnahodil, I'll merge this when Trav is finished

PR #1617, opened 2015-03-10
New filter type mapping

dnahodil:
This change moves the choice of the correct UI component for a given GeoServer type to the client side. As it is it should be backwards compatible with the WFS scanner and server-side types which are used currently but which will disappear soon.

### DIFF
--- a/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
@@ -21,7 +21,7 @@ describe("Portal.filter.BaseFilterPanel", function() {
 
         it("should return undefined", function() {
 
-            filter.getType = function() { return null };
+            filter.getType = function() { return '' };
 
             panel = newFilterPanelFor({
                 layer: {},
--- a/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
@@ -47,8 +47,8 @@ describe("Portal.filter.BaseFilterPanel", function() {
             expectNewFilterPanelForString('BooleanFilterPanel', 'Boolean');
         });
 
-        it("should create BoundingBoxFilterPanel", function() {
-            expectNewFilterPanelForString('BoundingBoxFilterPanel', 'BoundingBox');
+        it("should create GeometryFilterPanel", function() {
+            expectNewFilterPanelForString('GeometryFilterPanel', 'BoundingBox');
         });
 
         it("should create NumberFilterPanel", function() {
--- a/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/BaseFilterPanelSpec.js
@@ -11,21 +11,14 @@ describe("Portal.filter.BaseFilterPanel", function() {
     describe("newFilterPanelFor()", function() {
 
         var newFilterPanelFor = Portal.filter.BaseFilterPanel.newFilterPanelFor;
-        var filter;
-        var panel;
-
-        beforeEach(function() {
-
-            filter = {};
-        });
 
         it("should return undefined", function() {
 
-            filter.getType = function() { return '' };
-
-            panel = newFilterPanelFor({
+            var panel = newFilterPanelFor({
                 layer: {},
-                filter: filter
+                filter: {
+                    getType: function() { return '' }
+                }
             });
 
             expect(panel).toBeUndefined();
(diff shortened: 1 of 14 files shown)


anguss00:
I'll review

src/test/javascript/portal/filter/BaseFilterPanelSpec.js
```diff
@@ -11,21 +11,14 @@ describe("Portal.filter.BaseFilterPanel", function() {
...
         it("should return undefined", function() {
 
-            filter.getType = function() { return null };
-
-            panel = newFilterPanelFor({
+            var panel = newFilterPanelFor({
                 layer: {},
-                filter: filter
+                filter: {
+                    getType: function() { return '' }
```
anguss00: Nice @dnahodil

web-app/js/portal/filter/BaseFilterPanel.js
```diff
@@ -101,28 +101,51 @@ Portal.filter.BaseFilterPanel = Ext.extend(Ext.Panel, {
...
+        type = 'datetime'
     }
-    else if (cfg.filter.getType() == "Date") {
-        newFilterPanel = new Portal.filter.DateFilterPanel(cfg);
+    else if (type === "boundingbox") {
+        type = 'geometrypropertytype'
     }
-    else if (cfg.filter.getType() == "DateRange") {
-        newFilterPanel = new Portal.filter.DateFilterPanel(cfg);
+    else if (type === "number") {
```
anguss00: Could be I'm being paranoid, but what do you think about using the == operator instead of the === operator? That way we get around any potential type comparison problems.

anguss00:
Looks great aside from that one comment @dnahodil

anguss00:
Also, another point... the sort order of filters appears to be an issue here. All filters are getting a sort order of -1. Will investigate

web-app/js/portal/filter/BaseFilterPanel.js
dnahodil: Yeah, because of line 104 we'll always have a String. It's definitely something to be wary of, though, particularly with numerical comparisons. Another tip is to start strict and loosen it up if necessary. That's essentially what we're doing by using `===`.

dnahodil:
Good spotting on that ordering problem. I'll get onto that.

web-app/js/portal/filter/BaseFilterPanel.js
anguss00: Yep, sounds like a good plan. I think it's pretty safe here, more of a question for my future reference ;-)

dnahodil:
Turns out the ordering of those UI components is complicated by some other stuff we're doing. We're grouping them and inserting spacers, etc. I'm working this out at the moment.

dnahodil:
Ready to go again, @anguss00

src/test/javascript/portal/filter/FilterGroupPanelSpec.js
```diff
@@ -80,65 +75,53 @@ describe("Portal.filter.FilterGroupPanel", function() {
...
-            }
+            spyOn(Portal.filter.BooleanFilterPanel.prototype, '_createField');
+            spyOn(Portal.filter.BooleanFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.NumberFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.NumberFilterPanel.prototype, 'setLayerAndFilter');
+            spyOn(Portal.filter.GeometryFilterPanel.prototype, '_createField');
+            spyOn(Portal.filter.GeometryFilterPanel.prototype, 'setLayerAndFilter');
+            spyOn(Portal.filter.DateFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.DateFilterPanel.prototype, '_createField');
+            spyOn(Portal.filter.ComboFilterPanel.prototype, '_setExistingFilters');
```
dnahodil: A number of these calls to `spyOn()` will hopefully be unnecessary after I've done some of the refactoring we have planned.

src/test/javascript/portal/filter/FilterGroupPanelSpec.js
```diff
@@ -292,14 +272,54 @@ describe("Portal.filter.FilterGroupPanel", function() {
...
+        var numComponentsPerGroup = 2;
+        var numFiltersExpected;
+
+        beforeEach(function() {
+            spyOn(Portal.filter.NumberFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.NumberFilterPanel.prototype, '_createField');
+            spyOn(Portal.filter.DateFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.DateFilterPanel.prototype, '_createField');
+            spyOn(Portal.filter.ComboFilterPanel.prototype, '_setExistingFilters');
+            spyOn(Portal.filter.ComboFilterPanel.prototype, '_createField');
```
dnahodil: These ones, too.

anguss00:
I'm on it @dnahodil

dnahodil:
Just spotted a small problem with filter group labels. Will address in a new commit.

dnahodil:
New commit added @anguss00.

Indentation issue fixed, also.

dnahodil:
Oh. I missed a test. Will fix that.

anguss00:
Nice one @dnahodil. Will merge this when Trav has finished his great work

PR #1618, opened 2015-03-11
Multiple substitutions for download URLs can now be configured.

jkburges:
This is a requirement for some SRS collections.

Depends on https://github.com/aodn/chef-private-sample/pull/9

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
@@ -25,7 +25,7 @@ class DownloadController extends RequestProxyingController {
 
     def urlListForLayer = {
 
-        def (fieldName, prefixToRemove, newUrlBase) = _loadCommonFields(params)
+        def (fieldName, urlSubstitutions) = _loadCommonFields(params)
 
         if (!fieldName) {
             render text: 'urlFieldName was not provided', status: 400
@@ -34,7 +34,7 @@ class DownloadController extends RequestProxyingController {
 
         _performProxying(
             requestSingleFieldParamProcessor(fieldName),
-            urlListStreamProcessor(fieldName, prefixToRemove, newUrlBase)
+            urlListStreamProcessor(fieldName, urlSubstitutions)
         )
     }
 
@@ -46,7 +46,7 @@ class DownloadController extends RequestProxyingController {
 
     def downloadNetCdfFilesForLayer = {
 
-        def (fieldName, prefixToRemove, newUrlBase) = _loadCommonFields(params)
+        def (fieldName, urlSubstitutions) = _loadCommonFields(params)
 
         if (!fieldName) {
             render text: 'urlFieldName was not provided', status: 400
@@ -61,7 +61,7 @@ class DownloadController extends RequestProxyingController {
         }
 
         def resultStream = new ByteArrayOutputStream()
-        def streamProcessor = urlListStreamProcessor(fieldName, prefixToRemove, newUrlBase)
+        def streamProcessor = urlListStreamProcessor(fieldName, urlSubstitutions)
 
         try {
             _executeExternalRequest url, streamProcessor, resultStream
@@ -145,7 +145,7 @@ class DownloadController extends RequestProxyingController {
         }
     }
 
-    def urlListStreamProcessor(fieldName, prefixToRemove, newUrlBase) {
+    def urlListStreamProcessor(fieldName, urlSubstitutions) {
 
         return { inputStream, outputStream ->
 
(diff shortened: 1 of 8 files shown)


grails-app/controllers/au/org/emii/portal/DownloadController.groovy
```diff
@@ -174,10 +174,14 @@ class DownloadController extends RequestProxyingController {
...
                     def rowValue = currentRow[fieldIndex].trim()
-                    rowValue = rowValue.replace(prefixToRemove, newUrlBase)
+                    urlSubstitutions.each {
+                        search, replace ->
+
+                        rowValue = rowValue.replaceAll(search, replace)
+                    }
 
                     if (rowValue && includedUrls.add(rowValue)) {
-                        outputWriter.print "$rowValue\n"
```
dnahodil: I think this used `print()` with a newline character for a reason. Maybe different platforms were using `\n\r` and that caused problems somewhere along the line (e.g. maybe GoGoDuck).

Happy to say that if we didn't have a test for it then it can't have been *that* important. But just something to keep in mind.

test/unit/au/org/emii/portal/DownloadControllerTests.groovy
```diff
@@ -235,12 +234,20 @@ http://data.imos.org.au/IMOS/Q9900542.nc\n\
...
 http://data.imos.org.au/IMOS/Q9900540.nc\n\
 http://data.imos.org.au/IMOS/Q9900541.nc\n\
+/some_path/bar/123.nc\n\
 """
 
         def inputStream = new ByteArrayInputStream(input.bytes)
         def outputStream = new ByteArrayOutputStream()
 
-        def sp = controller.urlListStreamProcessor('relativeFilePath', '/mnt/imos-t4/', 'http://data.imos.org.au/')
+        //def sp = controller.urlListStreamProcessor('relativeFilePath', '/mnt/imos-t4/', 'http://data.imos.org.au/')
```
dnahodil: I assume this is no longer needed

grails-app/domain/au/org/emii/portal/Server.groovy
```diff
@@ -223,4 +222,10 @@ class Server {
             operations << operation
         }
     }
+
+    def getUrlListDownloadSubstitutions() {
+        return grailsApplication?.config?.knownServers?.find {
+            this.uri.startsWith(it.uri)
+        }?.urlListDownloadSubstitutions ?: [:]
```
dnahodil: All those save navigation operators could be a smell (I would say so).

If you wanted to change this we could update the tests so that Server always has a (mocked) `grailsApplication` object. I think there are 3 tests that break, currently.

Alternatively, the following code is equivalent (assuming no-one removes the knownServers entry from their config) and doesn't use all those operators:
``` groovy
return grailsApplication ? grailsApplication.config.knownServers.find {
    this.uri.startsWith(it.uri)
}.urlListDownloadSubstitutions : [:]
```
No tests would need to be updated to make this change.

grails-app/migrations/20150310-JB-DropUrlSubstitution.groovy
```diff
@@ -0,0 +1,10 @@
+databaseChangeLog = {
+
+    changeSet(author: "jkburges", id: "1425952041-1") {
+        [ 'url_list_download_prefix_to_remove', 'url_list_download_prefix_to_substitue' ].each {
```
dnahodil: Good thinking to tidy these up.

grails-app/controllers/au/org/emii/portal/DownloadController.groovy
jkburges: Do you think I should change it back?  It first came in with https://github.com/aodn/aodn-portal/commit/f9c0ec81caa8c7e7fcf69be33198a48cb7a43c0e, which implies that it wasn't `println` previously, so I *think* this should be ok (better perhaps, since it's platform independent).

dnahodil: Nah, don't change it unless it turns out to be a problem. And if it is a problem we just fix it and make a comment in the git commit message, or on the line itself) why it is this way. And probably write a test, too.

grails-app/domain/au/org/emii/portal/Server.groovy
```diff
@@ -223,4 +222,10 @@ class Server {
             operations << operation
         }
     }
+
+    def getUrlListDownloadSubstitutions() {
+        return grailsApplication.config.knownServers.find {
+            this.uri.startsWith(it.uri)
+        }?.urlListDownloadSubstitutions ?: [:]
```
jkburges: Note that it's still necessary to have the `?` after find - since there's no guarantee that the server will be in the known servers config.  I think this is slightly complicated by the fact that as it stands currently, *some* server config is coming from the DB, some from the config file.

When everything is coming from the config file, then it would probably be impossible to get to this method unless the server had config... if that makes sense.

jkburges:
Updated @dnahodil.

grails-app/domain/au/org/emii/portal/Server.groovy
dnahodil: Good point about `find()` not necessarily finding anything. I was thinking about `findAll()` which returns an empty list if nothing is found.

Yep. Makes sense.

danfruehauf:
I'd just like to note that the second commit message says nothing about the commit is about. This is why squashing it would have made sense.

PR #1621, opened 2015-03-12
Allow filters to be loaded from DB, GeoServer, or filesystem

dnahodil:
This is for the time being until we get filters out of the DB. At that point we can change this back to a boolean setting to choose between GeoServer (normal behaviour) and loading from filesystem (for development, troubleshooting) if needed.


anguss00:
I'll review

PR #1622, opened 2015-03-12
Filters refactor

dnahodil:
This work is basically all moving and removing. Nothing new/changed.

### DIFF
--- a/web-app/js/portal/filter/GeometryFilterPanel.js
+++ b/web-app/js/portal/filter/GeometryFilterPanel.js
@@ -73,10 +73,6 @@ Portal.filter.GeometryFilterPanel = Ext.extend(Portal.filter.BaseFilterPanel, {
         trackFiltersUsage('filtersTrackingSpatialConstraintAction', OpenLayers.i18n('trackingValueCleared'));
     },
 
-    _setExistingFilters: function() {
-        // Never restored from an existing filter
-    },
-
     needsFilterRange: function() {
         return false;
     },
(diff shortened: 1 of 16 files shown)


anguss00:
I'll review

web-app/js/portal/filter/GeometryFilterPanel.js
```diff
@@ -37,8 +52,15 @@ Portal.filter.GeometryFilterPanel = Ext.extend(Portal.filter.BaseFilterPanel, {
...
-    isVisualised: function() {
-        return false;
+    getFilterData: function() {
+
+        return {
+            name: this.filter.getName(),
+            visualised: this.isVisualised(),
+            cql: this.getCQL(),
+            humanValue: this._getCQLHumanValue(),
+            type: "geom"
```
anguss00: Can we get filterType from the filter object itself here?

dnahodil: Good question, I hadn't noticed that.

That is something we will need to look at in further refactorings. The value 'geom' here is actually arbitrary, it's not a Portal FilterType and it's not any of the GeoServer types. I'm not sure what we'll do here as if we use the new Filter object's type then there are 5 types that count as geometries so the code (elsewhere where this value is used) would have to check for 5 possible values instead of just this one. We might be able to replace this with a nicely-named boolean (e.g. 'isGeometry: true') or soemthing.

dnahodil: For reference: https://github.com/aodn/aodn-portal/blob/3bca2fab04bd8b241a2d09f9039898b3848e3c4f/web-app/js/portal/prototypes/OpenLayers.js#L181

dnahodil:
I've just added 2 small commits, @anguss00

PR #1627, opened 2015-03-16
Pass time-range filter values from server to front-end

dnahodil:


web-app/js/portal/filter/Filter.js
```diff
@@ -15,6 +15,8 @@ Portal.filter.Filter = Ext.extend(Object, {
         this.displayLabel = filterDetail.label;
         this.type = filterDetail.type;
         this.visualised = filterDetail.visualised;
+        this.wmsStartDateName = filterDetail.wmsStartDateName;
+        this.wmsEndDateName = filterDetail.wmsEndDateName;
```
anguss00: Should we run getters here as per type and display label?

dnahodil: Yes, we should. I'll add those in.

dnahodil: Change made @anguss00. See below.

I didn't change any of the current usages (which currently use the property instead of the newly-added getter method) because they will basically all change in the refactoring.

anguss00: :+1: @dnahodil

PR #1632, opened 2015-03-18
fixes #1581

### LINKED ISSUE #1581: Time filter not applied on step 3 if user does not click out of box
Instance name: IMOS
            App version: 3.40.2
            Build date: Wed Feb 04 17:00:40 EST 2015
            Build number: 338
            Git ref: fa518e0bcace90ff148d31b9768925466bb0d433

It is already in prod too.

Steps to reproduce:

Select Parameter satellite, then select any srs collection to add.

Once on step 2 wait until time filter loads and then click in the from date box

Manually type to change the from date. Click next or download button to proceed to step 3

What should happen:

New from date is applied, to the possible download.

What does happen:

Date is still set at default,

NB if you then go back to step 2 and then forward to step 3 the correct date is then applied.

pmbohm:


pmbohm:
@dnahodil I made a mess of things rebasing, so here's a new clean branch

src/test/javascript/OpenLayers/Layer/NcWMSSpec.js
```diff
@@ -99,19 +99,11 @@ describe("OpenLayers.Layer.NcWMS", function() {
...
             cachedLayer.loadTimeSeriesForDay(moment.utc('2001-07-02T00:00:00Z'));
-
-            waitsFor(function() {
-                return functionCalled;
-            }, "_timeSeriesDatesLoaded not called", 1000);
-
-            runs(function() {
-                expect(cachedLayer._timeSeriesLoadedForDate).toHaveBeenCalled();
-            });
+            expect(cachedLayer._timeSeriesLoadedForDate).toHaveBeenCalled();
```
dnahodil: I'm not trying to be a pain @pmbohm but is this bit also unrelated?

I'm not trying to be difficult, it is really disconcerting seeing unrelated changes in a PR and makes it dangerous to press the merge button as you're not confident about what changes will go in. **Let me know if you want a hand sorting this out.**

dnahodil:
Good to go, @pmbohm. Nice one.

Issue #1613, opened 2015-03-05
Step 2 'Info' tab - Some links are useless


dnahodil:
**Steps to reproduce**
Open the Portal and add the Argo data collection
On step 2 open the 'Info' tab

**What does happen**
You will see the abstract and a number of links (see below)
![screenshot from 2015-03-05 16 16 24](https://cloud.githubusercontent.com/assets/2774917/6499652/548edf5e-c353-11e4-8812-7de0ec5512cb.png)
Some of these links (e.g. "This OGC WFS service returns filtered geographic information. ....") are useless. If you click on them you get a 4xx or 5xx error.

**What should happen**
Only meaningful links should be shown here.

PR #1637, opened 2015-03-19
Extract info panel link html logic to portal js, use configuration to de...

anguss00:
...termine list of online resources.

Fixes issue #1613

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -393,7 +393,7 @@ class LayerController {
         }
     }
 
-    def getFormattedMetadata = {
+    def getMetadataAbstract = {
 
         def responseText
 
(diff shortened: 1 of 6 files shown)


grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -408,44 +408,17 @@ class LayerController {
...
-                    html += "</ul>"
 
-                    responseText = html
+                    responseText = abstractText
                 }
             }
             catch (SAXException e) {
-                log.warn("Error getting formatted metadata, params: ${params}", e)
+                log.warn("Error getting metadata abstract, params: ${params}", e)
                 responseText = "<BR>The metadata record is not available at this time."
```
danfruehauf: If we're formatting stuff in the client side, perhaps returning here a HTTP error is adequate and then the client can deal with it?

grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -408,44 +408,17 @@ class LayerController {
...
             }
             catch (FileNotFoundException e) {
-                log.warn("Error getting formatted metadata, params: ${params}", e)
+                log.warn("Error getting metadata abstract, params: ${params}", e)
                 responseText = "<BR>The metadata record is not available at this time."
```
danfruehauf: Same here - with returning an error.

web-app/js/portal/details/InfoPanel.js
```diff
@@ -26,20 +26,54 @@ Portal.details.InfoPanel = Ext.extend(Ext.Container, {
...
         Ext.Ajax.request({
             url: metadataUrl,
             scope: this,
             success: function(resp, options) {
-                this.update(resp.responseText, false);
+
+                this._constructInfoTabHtml(resp.responseText, this.layer.parentGeoNetworkRecord.data.onlineResources);
             },
             failure: function(resp) {
                 this.update("<i>" + OpenLayers.i18n('noMetadataMessage') + "</i>", false);
```
danfruehauf: So this is where you should return the error to the user, in `noMetadataMessage`.

grails-app/controllers/au/org/emii/portal/LayerController.groovy
danfruehauf: In other words - if the client is already handling errors, and this is just a simple proxy request - let the client do it all the way.

anguss00: Agreed, should definitely be handled in the client side. Do you suggest adding logging to the failure clause in the Ajax request on the client?

danfruehauf: Probably yes, wouldn't hurt.

anguss00:
Changes made @danfruehauf

PR #1640, opened 2015-03-19
Filters error handling

dnahodil:
These changes cover three possibilities from filter loading.
1) Getting the list of filters for a layer fails - 'Filtering unavailable' message is displayed
2) Getting possible values for a filter fails - (Only currently applicable for the combo picker) drop-down is disabled with a message
3) Any filter type is unrecognised - The problem is logged but an empty UI component is used so the UI does not break

### DIFF
--- a/web-app/js/portal/filter/BaseFilterPanel.js
+++ b/web-app/js/portal/filter/BaseFilterPanel.js
@@ -126,7 +126,8 @@ Portal.filter.BaseFilterPanel.newFilterPanelFor = function(cfg) {
     }
     else {
         //Filter hasn't been defined
-        log.error("Could not create filter panel for type'" + type + "'");
+        log.error("Could not find correct filter panel class for type'" + type + "'");
+        newFilterPanel = new Portal.filter.EmptyFilterPanel(cfg);
     }
 
     return newFilterPanel;
--- a/web-app/js/portal/filter/BaseFilterPanel.js
+++ b/web-app/js/portal/filter/BaseFilterPanel.js
@@ -125,8 +125,7 @@ Portal.filter.BaseFilterPanel.newFilterPanelFor = function(cfg) {
         newFilterPanel = new Portal.filter.GeometryFilterPanel(cfg);
     }
     else {
-        //Filter hasn't been defined
-        log.error("Could not find correct filter panel class for type'" + type + "'");
+        log.error("Unhandled filter type '" + type + "' for filter '" + cfg.filter.getName() + "' on layer '" + cfg.layer.wmsName + "'");
         newFilterPanel = new Portal.filter.EmptyFilterPanel(cfg);
     }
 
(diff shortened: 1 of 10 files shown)


danfruehauf:
I'll take a look, managed to clean my table a bit...

web-app/js/portal/filter/EmptyFilterPanel.js
```diff
@@ -0,0 +1,32 @@
...
+        Portal.filter.EmptyFilterPanel.superclass.constructor.call(this, cfg);
+    },
+
+    _createField: function() {},
+
+    getFilterData: function() {
+
+        return {
+            visualised: false
+        }
```
danfruehauf: I think need `;` here.

danfruehauf:
What's the best way of *testing* that? Mock some filters that don't load?

dnahodil:
Yeah, basically I just added something like
``` groovy
render text: "", status: 500
return
```
At the start of getFiltersAsJSON and separately getFilterValuesAsJSON. That lets you test the handling of those cases. To test the unknown filter type you could change the xml files (if you're loading from filesystem) or I just changed the switch statement in BaseFilterPanel.js so that the type value sisn't match any of the handled values.

web-app/js/portal/filter/BaseFilterPanel.js
```diff
@@ -126,7 +126,8 @@ Portal.filter.BaseFilterPanel.newFilterPanelFor = function(cfg) {
     }
     else {
         //Filter hasn't been defined
-        log.error("Could not create filter panel for type'" + type + "'");
+        log.error("Could not find correct filter panel class for type'" + type + "'");
```
danfruehauf: Space missing after `type'`

danfruehauf: But maybe a more succinct error message can be used here, such as "Unknown filter type 'TYPE'". Perhaps also add the layer name if you can - will assist in fixing the problem.

dnahodil:
Thanks @danfruehauf 

Changes made

dnahodil:
Rebased off master

danfruehauf:
The rebase broke something. When drawing a bounding box I get:
```
TypeError: this.combo.enabled is not a function
if (this.combo.enabled() && this.combo.getValue()) {
ComboFi...anel.js (line 92, col 12)
```

dnahodil:
Thanks. I'll check it out on Monday

dnahodil:
Changes made @danfruehauf

danfruehauf:
Cool. I'll be having a look.

web-app/js/portal/filter/BaseFilterPanel.js
```diff
@@ -125,8 +125,8 @@ Portal.filter.BaseFilterPanel.newFilterPanelFor = function(cfg) {
         newFilterPanel = new Portal.filter.GeometryFilterPanel(cfg);
     }
     else {
-        //Filter hasn't been defined
-        log.error("Could not create filter panel for type'" + type + "'");
+        log.error("Unhandled filter type '" + type + "' for layer '" + cfg.layer.wmsName + "'");
```
danfruehauf: So when testing, I realized it'd be beneficial to have also the filter name here, but it's not a biggie because we have already the layer name. Up to you, not blocking me from merging this.

dnahodil:
Hang on, I'll fix that.

PR #1641, opened 2015-03-19
fixes #1604

### LINKED ISSUE #1604: Malformed included css file
### What happens

This incorrect url is being written into the portal pages
```
<link rel="stylesheet" type="text/css" href="/aodn-portal/home/css?v=1425423212840"/>
```

pmbohm:
Hides link to empty css file when PortalBranding is not used

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/HomeController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/HomeController.groovy
@@ -42,6 +42,6 @@ class HomeController {
     }
 
     def css = {
-        render text: portalBranding.css
+        render text: portalBranding.css, contentType: "text/css"
     }
 }


danfruehauf:
I'll handle this PR. Probably just verify the fix and add the new line.

danfruehauf:
Since I completely changed the fix here, a quick review by someone else should follow. Please note my latest comment on #1604:
https://github.com/aodn/aodn-portal/issues/1604#issuecomment-84737037

dnahodil:
Fixing the content type is good (I think this is the actual issue noticed here). I think it would be good to also prevent the CSS request when possible as Phil's commit did. Having both will make things better.

dnahodil:
Oh. Too slow, apparently.

danfruehauf:
As much as this can cause another 50ms delay, I'd much rather have one less branch in the code. A matter of preference I'd say.

PR #1643, opened 2015-03-20
Do not disable dates if get missing days returns empty array, Ext bug 1339

anguss00:
Ref http://www.sencha.com/forum/showthread.php?112648-OPEN-1339-Datefield-setDisabledDates-Bug-%28setting-empty-again%29

Fixes issue https://github.com/aodn/aodn-portal/issues/1644

Fixes Brewin Nanoplankton 2012 issue on production


dnahodil:
Looks good

anguss00:
:+1:

Issue #1638, opened 2015-03-19
UUID Tag held during refresh


kereid:
Instance name: IMOS
            App version: 3.42.1
            Build date: Wed Mar 18 10:53:36 AEDT 2015
            Build number: 374
            Git ref: fd0444aca561d4384989d0be1ef6bc5ae360d4bf

Steps to reproduce:

Add any collection to the map, proceed to step 3 or back to step 1, conduct hard refresh.

What should happen:

Portal is refreshed - no collection on map, user taken to step 1 to start again.

What does happen:

user taken to step 2 and collection re-loads.

pmbohm:
I agree this is a bug of sorts. I find the behaviour very weird as I am used refreshing the page as a developer, so i'm biased by this.
To fix this bug we either need a permlink button so that users can save/share state (reinstating previous behaviour) or we need a prominent clear/refresh button on all pages.
@pblain a ruling please?

dnahodil:
I agree that this would be less of an issue for people who use refresh to clear collections if clearing collections was easier through the Portal UI (which I believe we are working towards).

pblain:
I'd like to roll back Jon's query string feature for now.

Going forward we need a clear/refresh button on all three steps.

We also need, at a minimum, to include the step number in the query string so that refreshing leaves you on the same step.  You'll lose filtering, zooming, faceted search settings etc etc - but I think we can live with that.  But clicking refresh and jumping to step 2 is just too weird.

None of these enhancements are priorities though - which is why I want to roll it back.  This is coming up because a 10% feature made it's way into master without any of this having been considered.  A lot of good stuff comes out of 10% time (including this), but it can't just get merged without asking me.

jkburges:
This change has been reverted.  I agree that 10% features can't be released without @pblain's approval.  For this, I'm sorry, but luckily in this case, it hasn't been released to prod, and so I hope that a minimum of time has been spent on this.

Now, my arguments for this feature... :grin: 

Many websites do this, i.e. reflect the current state in the URL.  e.g. github (particularly issues), gmail, and wait for it, carsales: http://www.carsales.com.au/cars/results?sortby=~Price&offset=0&q=(Make%3D%5BHolden%5D%26Service%3D%5BCarsales%5D)&area=Stock&vertical=car&WT.z_srchsrcx=makemodel  Refreshing on any of these websites does not "reset", the page will look the same as it did, even though it does reload.

[...]

PR #1645, opened 2015-03-20
Revert "Updating URL when collections are added/removed."

dnahodil:
FYI @jkburges 3.42.2 is going to replace the version @kereid is currently testing. So I will move the revert commit here, too.

This reverts commit 989b22c39a2cfb3c7b5e13678d84514f622c1dee.

This change was not approved by @pblain. See: https://github.com/aodn/aodn-portal/issues/1638


jkburges:
I think the "to" branch should be "3.42.x"...

dnahodil:
Basically this branch is only going to be around long enough to build 3.42.2. I wasn't planning on having it around for other 3.x releases.

PR #1646, opened 2015-03-20
MARVL urlListForFeatureRequest endpoint accepts list of URL substitutions

dnahodil:
This brings it in line with BODAAC.

These tests didn't fail when the BODAAC changes were made because an important function was mocked out. So I have updated the test so that changes to that function (including to it's method signature) will now break the tests.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -311,8 +311,9 @@ portal {
 // MARVL project integration
 marvl {
     urlList {
-        prefixToRemove = "/mnt/imos-t3/"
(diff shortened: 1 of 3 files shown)


grails-app/conf/Config.groovy
```diff
@@ -311,8 +311,9 @@ portal {
 // MARVL project integration
 marvl {
     urlList {
-        prefixToRemove = "/mnt/imos-t3/"
-        newUrlBase = "http://data.aodn.org.au/"
+        substitutions = [
```
jkburges: There'll be corresponding chef changes also?

PR #1647, opened 2015-03-20
Update README for current features/requirements

dnahodil:
* Removed references to scanners and spatial search
* Removed listed 'features' that no longer exist
* Updated list of configuration overrides


danfruehauf:
Can still be further improved, but at least it removed outdated stuff.

dnahodil:
Yeah, I just asked Pete to review it when he looks at the AODN Cookbook. But I think at least it's technically accurate at the moment.

PR #1651, opened 2015-03-22
remove redundant config snippets

danfruehauf:
Trivial config removal


dnahodil:
Mostly good removals, @danfruehauf.

Except `indexedFile.fileSizeColumnName` which is still needed.

danfruehauf:
Oh, correct. My bad.

danfruehauf:
Good to go. Thanks for that!

PR #1656, opened 2015-03-23
Renamed layer controller methods for filters and styles

dnahodil:
getStylesAsJSON -> getStyles
getFiltersAsJSON -> getFilters
getFilterValuesAsJSON -> getFilterValues

I don't *think* anyone will object to this change. Including 'asJSON' in the name seems redundant. Maybe it was needed as a clarification at some point.

This will break get_filters.rb eventually (when this change makes it to prod) but I think that'll be easy to pick up and fix when it happen.

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -72,7 +72,7 @@ class LayerController {
         }
     }
 
-    def getStylesAsJSON = {
+    def getStyles = {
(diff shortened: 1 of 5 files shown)


anguss00:
I'll review

danfruehauf:
When this makes it into prod, we will get rid of `get_filters.rb` and those other shenanigans.

anguss00:
:+1:

danfruehauf:
Fine by me, can be merged, but @anguss00 wants to do it.

PR #1658, opened 2015-03-23
Determine filter source (GeoServer/files) from env var

dnahodil:
I've accidentally committed the `dynamicFilters` setting a couple of times already today. :S

I've set it up so it can be loaded from an environment variable to make it easier when you're working from filesystem for an extended period. Default is 'false' which is to read from GeoServer.

I also took it out of the feature toggles section as per a previous discussion with @jkburges and @danfruehauf.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -84,7 +84,7 @@ downloadAuth {
 }
 
 featureToggles {
-    dynamicFilters = true
+    // someFeature = <boolean>
 }
 
 // Google Analytics
@@ -92,6 +92,10 @@ googleAnalytics.trackingId = null
 
 // set per-environment serverURL stem for creating absolute links
 def env = System.getenv()
+
+filtering {
+    filePath = env['FILTERS_PATH']
+}
 
 environments {
 
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -68,7 +68,7 @@ class LayerController {
             return new NcwmsServer()
         }
         else {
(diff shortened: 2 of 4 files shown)


grails-app/conf/Config.groovy
```diff
@@ -93,6 +93,10 @@ googleAnalytics.trackingId = null
 // set per-environment serverURL stem for creating absolute links
 def env = System.getenv()
 
+filtering {
+    loadFromFiles = Boolean.parseBoolean(env['LOAD_FILTERS_FROM_FILES'])
```
dnahodil: Happy to take suggestions for the env var name. Although, I am happy enough with this.

danfruehauf: I'd say if you're already working on that - have it as a string which represents a path rather than a true/false toggle. This will be useful for IMAS for instance.

danfruehauf: About the name, I'd go with `FILTERS_PATH`, shorter and easy to remember.

dnahodil: Ok, I'll make changes so we can specify a file path.

danfruehauf: If the path is not specified - imply dynamic filtering.

dnahodil:
@danfruehauf as discussed I've changed the environment variable to be a path so that filters can be loaded from anywhere on a file system. It assumes the same structure within that directory.

Setting the environment variable to 'filters' will maintain the current filesystem behaviour.

I was also able to remove some duplication in a second commit.

danfruehauf:
Yes, that looks good. I trust you it works.

dnahodil:
I hope it does :)

PR #1659, opened 2015-03-23
Stacktraces in logging output

dnahodil:
@danfruehauf a suggestion for how we can clean up the stack traces in the GeoserverServer tests.

We don't have the logging output and we don't have to add an extra code path.


danfruehauf:
:+1:

PR #1663, opened 2015-03-24
Unique values from WMS

dnahodil:
With this PR the Portal now loads the unique values for a filter from the map layer (instead of the data layer). The map layer should contain all the unique values represented anyway but will be much more performant.

This also fixes a bug where we were determining the download layer from the wrong field in the GeoNetwork record. It worked most of the time but failed in some cases (noticed on IMOS - Australian National Mooring Network (ANMN) Facility - Current velocity time-series).

### DIFF
--- a/get_filters.rb
+++ b/get_filters.rb
@@ -144,7 +144,7 @@ def convert_filter_type(filter_type, filter_name)
   return return_filter_type
 end
 
-def write_filters(filters_dir, filters, opts)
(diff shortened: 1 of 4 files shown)


dnahodil:
This is now available. It is the change resulting from our meeting yesterday re: loading unique values form the map layer rather than the data layer.

anguss00:
I'll review this

PR #1664, opened 2015-03-24
Tidy up

dnahodil:
Remove _getWmsVersionString methods (there were three of them/it :worried:).
Remove references to server.type which is no longer supported.


anguss00:
I'll review

PR #1665, opened 2015-03-24
Add resources folder

dnahodil:
Moved these photoshop files out of the images directory (where they were being served publicly)


danfruehauf:
Is anything using that? No other references to that?

dnahodil:
Is anything using what? the resources folder or the photoshop files?

The photoshop files are like source files for parts of the UI for the Portal. They shouldn't be published in the app but I think they should be kept in the project.

PR #1668, opened 2015-03-26
remove unused instance name

danfruehauf:
* use default.css (can be overriden by portalBranding)
 * use IMOS_logo.png (can be overriden by portalBranding)
 * eliminate use of INSTANCE_NAME and defaultInstanceName

Work for @pblain after he struggled a bit with a fair few bits of unnecessary configuration we still have in `Config.groovy`. Should still be improved more.


grails-app/conf/Config.groovy
```diff
@@ -332,12 +323,7 @@ try {
     println "Loading external config from '$configurationPath'..."
 
     def startupConfig = new ConfigSlurper(grailsSettings.grailsEnv).parse(new File(configurationPath).toURI().toURL())
```
dnahodil: I think this line can be removed as well

danfruehauf: Yep, correct, no more references to that...

danfruehauf: Removed.

dnahodil:
This is nice, @danfruehauf. Good one.

PR #1669, opened 2015-03-26
local config at Config.local.groovy

danfruehauf:
allow loading of local config from grails-app/conf/Config.local.groovy if it
exists when in development mode. easier to amend config without needing to
touch the main Config.groovy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ plugins
 \#*\#
 .\#*
 *.swp
+grails-app/conf/Config.local.groovy
 
 # Include code style settings for IntelliJ IDEA projects
 .idea/*
--- a/README.md
+++ b/README.md
@@ -64,19 +64,6 @@ Read the [Getting Started guide](https://github.com/aodn/aodn-portal/wiki/Gettin
 We welcome contributions so please feel free to fork the project, address any issues or add features and submit
 a pull request.
 
-## Overriding config when developing
-Certain config items can be overridden by setting environment variables appropriately when running in development mode, e.g.:
-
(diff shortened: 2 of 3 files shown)


dnahodil:
With this change can we get rid of the loading from environment variables for some settings?

It's best to have as few ways of setting/changes config as possible and (if I understand correctly) this new functionality would supersede what we were doing previously.

danfruehauf:
I think it should be open for debate. Personally I like the ability to override a bunch of variables at a time (this is usually my use case). We also won't need to hold a table translating between a groovy config variable and an environment variable. So yes, I'd be happy to get rid of environment variables and use this method instead.

I can also do some more work to cleanup any references to environment variables, given we get a consensus about using this way.

Note that this is solely for dev mode though.

dnahodil:
One problem with environment variables is that you need to add a new one for every setting you want to override. This solution means you can override any (and every) config in one go.

I don't know if IMAS (or anyone using our Portal software) use the environment variables or not. But I would think that the preferred setup would be for them to have a config groovy file (like we do in prod).

danfruehauf:
IMAS wouldn't use environment variables - they just override everything in chef config.

So... I'll remove the environment variable overriding altogether, there shall be one way of overriding config (without touching the config file).

danfruehauf:
Would you prefer it on a different commit, or squash it with the original one?

danfruehauf:
@dnahodil So with the last commit there's no more `env[]` references.

danfruehauf:
(Closed by accident)

PR #1672, opened 2015-03-27
Update ResettableDate

dnahodil:
ResettableDate used to have hasValue() and getValue() functions. hasValue() took into account the fact that an empty date field had a value of the empty string. This meant that DateFilterPanel had to use hasValue() before it used getValue() which made code a bit messier than needed.

This suggested change removes hasValue() function and make getValue() return null instead of an empty string. This means that DateFilterPanel can have a cleaner interface because getValue() can be used in boolean conditionals as well.

### DIFF
--- a/web-app/js/portal/filter/ui/DateFilterPanel.js
+++ b/web-app/js/portal/filter/ui/DateFilterPanel.js
@@ -79,7 +79,7 @@ Portal.filter.ui.DateFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel, 
             emptyText: emptyText,
             listeners: {
                 scope: this,
-                change: this._applyDateFilterPanel
+                change: this._applyDateFilter
             }
         });
     },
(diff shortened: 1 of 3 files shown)


jonescc:
I'll look at this

jonescc:
Is this just a tidy up or was there some issue being addressed here as well?

dnahodil:
It was just changing the ResettableDate interface and update the DateFilterPanel to use that. There *shouldn't* be any functional changes.

dnahodil:
Oh. I tell a lie. The message logged to Google Analytics when some dates were changed was incorrect. So that got fixed. You can see the corresponding unit test here: https://github.com/aodn/aodn-portal/blob/resettable_date_no_empty_strings/src/test/javascript/portal/filter/DateFilterPanelSpec.js#L85

web-app/js/portal/filter/ui/DateFilterPanel.js
```diff
@@ -120,26 +120,16 @@ Portal.filter.ui.DateFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel,
 
     _applyDateFilterPanel: function(component) {
```
jonescc: You probably didn't name this but this method name threw me off.  Would _applyDateFilter make more sense?

jonescc: Perhaps we don't need to use the passed component variable either - just refer to the panel - that would make the code below even clearer without having to refer to where the code is called

jonescc: Forget that its obviously important

jonescc:
Ran this up locally, and the date doesn't appear to be resettable any more.  If I change it and then hit the clear subset button the dates aren't reset.

web-app/js/portal/filter/ui/DateFilterPanel.js
dnahodil: Yeah, definitely.

dnahodil:
That's pretty fundamental. Let me have a look.

jonescc:
OK, so thats a totally unrelated issue with ncwms subsetting dates.  I'll raise an issue for it.

dnahodil:
_applyDateFilterPanel -> _applyDateFilter rename has been made.

Issue #1623, opened 2015-03-13
No error message for failed downloads


jkburges:
Sometimes, after clicking "download as..." something, nothing happens at all.

**Steps to reproduce**
1. Configure a collection to include a layer belonging to a non-allowed geoserver.
2. Add collection in portal, download as List of URLs.

**What should happen**

An error message (or *some* form of feedback) is displayed to the user.

**What does happen**

![image](https://cloud.githubusercontent.com/assets/1849176/6631123/495a47fe-c975-11e4-93cd-54aa1f3307be.png)

Issue #1624, opened 2015-03-13
Empty file for BOADAAC with unknown geoserver layer


jkburges:
When downloaded as "List of URLs" for an incorrectly configured collection, an empty file is returned (rather than some form of error).

**Steps to reproduce**
1. Configure a collection to include a layer belonging to an allowed geoserver, but non-existent layer.
2. Add collection in portal, download as "List of URLs".

**What should happen**

An error message (or *some* form of feedback) is displayed to the user.

**What does happen**

1. a 500 error is returned from geoserver to the portal server (but this is not visible to the user).
2. an empty file is downloaded.

pblain:
Such a  misconfiguration shouldn't happen in the first place - because config changes should be tested in content-test.  And if it does happen, our AODN content monitor should detect it and hide the data collection.  But until we get our act together the portal should handle it more gracefully, and an alert should be raised.

Issue #1620, opened 2015-03-11
Large CSV downloads


pblain:
Steps to reproduce

1. On Step 1, select:  IMOS - Australian National Facility for Ocean Gliders (ANFOG) - delayed mode glider deployments (search by "Platform > Glider" facets)
2. On Step 2, filter by bounding box -30N, 150E, -70S, 100W and platform_type = seaglider
3. Go to step 3 and download as CSV

What should happen

1. It starts downloading
2. I get a really big file

What does happen

1. Download never starts

See it happen here (at the 18:50 mark) ...
https://youtu.be/xf1FnLnyhvc?t=18m50s

dnahodil:
Possibly related to #1623

jkburges:
The query is taking a while to return results.  The download does start...eventually.

Here's an example query:

```sql
SELECT "file_id","deployment_name","platform_type","platform_code","time_coverage_start","time_coverage_end","TIME","TIME_quality_control",
"LATITUDE","LATITUDE_quality_control","LONGITUDE","LONGITUDE_quality_control","DEPTH","DEPTH_quality_control","PRES","PRES_quality_control",
"TEMP","TEMP_quality_control","PSAL","PSAL_quality_control","DOXY","DOXY_units","DOXY_quality_control","CDOM","CDOM_quality_control",
"FLU2","FLU2_quality_control","UCUR","UCUR_quality_control","VCUR","VCUR_quality_control","CNDC","CNDC_quality_control","DOXY_raw","HEAD",
"HEAD_quality_control","UCUR_GPS","UCUR_GPS_quality_control","VCUR_GPS","VCUR_GPS_quality_control","VBSC","VBSC_quality_control","OCR443",
"OCR443_quality_control","OCR443_1","OCR443_1_quality_control","OCR443_42","OCR443_42_quality_control","OCR443_52","OCR443_52_quality_control",
"OCR443_8","OCR443_8_quality_control","OCR443_9","OCR443_9_quality_control","OCR455_9","OCR455_9_quality_control","OCR456_6",
"OCR456_6_quality_control","OCR470_3","OCR470_3_quality_control","OCR470_5","OCR470_5_quality_control","OCR489_1","OCR489_1_quality_control",
"OCR489_2","OCR489_2_quality_control","OCR489_4","OCR489_4_quality_control","OCR489_59","OCR489_59_quality_control","OCR489_64",
[...]

PR #1677, opened 2015-03-30
Validate download requests to external geoserver url before download

anguss00:
Provide response to user if download is not available.

Fixes issue #1623

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
@@ -99,7 +99,7 @@ class DownloadController extends RequestProxyingController {
         if (!hostVerifier.allowedHost(url)) {
 
             log.error "Host for address '$url' not allowed"
-            render text: "Host for address '$url' not allowed", contentType: "text/html", encoding: "UTF-8", status: HTTP_400_BAD_REQUEST
+            render text: "Host for address '$url' not allowed", contentType: "text/html", encoding: "UTF-8", status: HTTP_403_FORBIDDEN
         }
         else {
             render text: url, contentType: "text/html", encoding: "UTF-8", status: HTTP_200_OK
--- a/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/DownloadController.groovy
@@ -92,6 +92,21 @@ class DownloadController extends RequestProxyingController {
         }
     }
 
+    def validateRequest = {
+
+        def url = params.url
+
+        if (!hostVerifier.allowedHost(url)) {
+
+            log.error "Host for address '$url' not allowed"
+            render text: "Host for address '$url' not allowed", contentType: "text/html", encoding: "UTF-8", status: HTTP_400_BAD_REQUEST
+        }
+        else {
+            render text: url, contentType: "text/html", encoding: "UTF-8", status: HTTP_200_OK
+        }
+        return
+    }
+
     def estimateSizeForLayer = {
 
         def urlFieldName = params.urlFieldName
--- a/grails-app/utils/au/org/emii/portal/HttpUtils.groovy
+++ b/grails-app/utils/au/org/emii/portal/HttpUtils.groovy
@@ -13,6 +13,7 @@ final class HttpUtils {
         static final HTTP_200_OK = 200
         static final HTTP_400_BAD_REQUEST = 400
         static final HTTP_401_UNAUTHORISED = 401
+        static final HTTP_403_FORBIDDEN = 403
         static final HTTP_404_NOT_FOUND = 404
         static final HTTP_500_INTERNAL_SERVER_ERROR = 500
         static final HTTP_502_BAD_GATEWAY = 502
--- a/test/unit/au/org/emii/portal/DownloadControllerTests.groovy
+++ b/test/unit/au/org/emii/portal/DownloadControllerTests.groovy
@@ -108,6 +108,24 @@ class DownloadControllerTests extends ControllerUnitTestCase {
         assertEquals "Host for address 'http://www.example.com/?PROPERTYNAME=relativeFilePath' not allowed", mockResponse.contentAsString
     }
 
+    void testValidateRequestInvalidHost() {
+
+        _setHostShouldBeValid(false)
+
+        controller.validateRequest()
+
+        assertEquals "Host for address 'http://www.example.com/' not allowed", mockResponse.contentAsString
+    }
+
+    void testValidateRequestValidHost() {
+
+        _setHostShouldBeValid(true)
+
+        controller.validateRequest()
+
+        assertEquals "http://www.example.com/", mockResponse.contentAsString
+    }
+
     void testDownloadNetCdfFilesForLayer() {
 
         mockParams.downloadFilename = 'somedata.txt'
--- a/test/unit/au/org/emii/portal/DownloadControllerTests.groovy
+++ b/test/unit/au/org/emii/portal/DownloadControllerTests.groovy
@@ -10,6 +10,7 @@ package au.org.emii.portal
 import grails.test.ControllerUnitTestCase
 
 import static au.org.emii.portal.DownloadController.SIZE_ESTIMATE_ERROR
+import static au.org.emii.portal.HttpUtils.Status.*
 
 class DownloadControllerTests extends ControllerUnitTestCase {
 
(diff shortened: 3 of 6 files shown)


grails-app/controllers/au/org/emii/portal/DownloadController.groovy
```diff
@@ -92,6 +92,21 @@ class DownloadController extends RequestProxyingController {
...
     }
 
+    def validateRequest = {
+
+        def url = params.url
+
+        if (!hostVerifier.allowedHost(url)) {
+
+            log.error "Host for address '$url' not allowed"
+            render text: "Host for address '$url' not allowed", contentType: "text/html", encoding: "UTF-8", status: HTTP_400_BAD_REQUEST
```
danfruehauf: Go with 403 here, more appropriate I think.

test/unit/au/org/emii/portal/DownloadControllerTests.groovy
```diff
@@ -108,6 +108,24 @@ class DownloadControllerTests extends ControllerUnitTestCase {
         assertEquals "Host for address 'http://www.example.com/?PROPERTYNAME=relativeFilePath' not allowed", mockResponse.contentAsString
     }
 
+    void testValidateRequestInvalidHost() {
+
+        _setHostShouldBeValid(false)
+
+        controller.validateRequest()
+
+        assertEquals "Host for address 'http://www.example.com/' not allowed", mockResponse.contentAsString
```
danfruehauf: Validate also the return code, which is more important here.

test/unit/au/org/emii/portal/DownloadControllerTests.groovy
```diff
@@ -108,6 +108,24 @@ class DownloadControllerTests extends ControllerUnitTestCase {
...
+    }
+
+    void testValidateRequestValidHost() {
+
+        _setHostShouldBeValid(true)
+
+        controller.validateRequest()
+
+        assertEquals "http://www.example.com/", mockResponse.contentAsString
```
danfruehauf: Validate return code also here.

web-app/js/portal/cart/Downloader.js
```diff
@@ -84,6 +88,25 @@ Portal.cart.Downloader = Ext.extend(Object, {
...
+            $downloaderLink = $('<iframe>', { id:'downloaderLink', src: url }).hide().appendTo('body');
+        }
+    },
+
+    _onDownloadRequestFailure: function(resp) {
+        log.error("Synchronous download request has failed with the following error " + resp.responseText);
+        Ext.Msg.alert(
+            OpenLayers.i18n('errorDialogTitle'),
+            OpenLayers.i18n('downloadErrorText')
+        )
```
danfruehauf: Missing `;`

danfruehauf:
This in essence solves the bug, however if the URL behind it will return an error, what happens then? Is it failing properly?

anguss00:
if the url behind it fails, then you get a 500 back and an empty .txt file. Given that this is listed under the bug report for #1624 i decided to introduce a allowed geoserver validation (which is needed anyway) to fix this bug while I work on #1624. Most likely, #1624 would need a validation mod to cover 500ing URLs

anguss00:
Changes made @danfruehauf

grails-app/utils/au/org/emii/portal/HttpUtils.groovy
```diff
@@ -16,6 +16,7 @@ final class HttpUtils {
         static final HTTP_404_NOT_FOUND = 404
         static final HTTP_500_INTERNAL_SERVER_ERROR = 500
         static final HTTP_502_BAD_GATEWAY = 502
+        static final HTTP_403_NOT_ALLOWED = 403
```
danfruehauf: Call it HTTP_403_FORBIDDEN - it's more accurate.

danfruehauf: And also put it before the 404 code.

test/unit/au/org/emii/portal/DownloadControllerTests.groovy
```diff
@@ -108,6 +108,26 @@ class DownloadControllerTests extends ControllerUnitTestCase {
...
     }
 
+    void testValidateRequestInvalidHost() {
+
+        _setHostShouldBeValid(false)
+
+        controller.validateRequest()
+
+        assertEquals "Host for address 'http://www.example.com/' not allowed", mockResponse.contentAsString
+        assertEquals 403, controller.renderArgs.status
```
danfruehauf: Compare it to `HTTP_403_FORBIDDEN`.

test/unit/au/org/emii/portal/DownloadControllerTests.groovy
```diff
@@ -108,6 +108,26 @@ class DownloadControllerTests extends ControllerUnitTestCase {
...
+    }
+
+    void testValidateRequestValidHost() {
+
+        _setHostShouldBeValid(true)
+
+        controller.validateRequest()
+
+        assertEquals "http://www.example.com/", mockResponse.contentAsString
+        assertEquals 200, controller.renderArgs.status
```
danfruehauf: Compare it to `HTTP_200_OK`.

danfruehauf:
@jkburges @dnahodil  I'd like your opinion on that. I feel I'm not in a good spot to decide whether we want to introduce a pretty complicated fix for that. Back in the day just another window used to open, then I made it much more complicated by using some ugly techniques to avoid it. And @anguss00 was left with no choice but to add more logic to make sure it works in a very annoying corner case.

anguss00:
I've made the required changes here @danfruehauf

danfruehauf:
@anguss00 That's good. See what the other guys are saying...

dnahodil:
It would definitely be good if we can find a solution that covers all cases (disallowed hosts and 4XXs and 5XXs from the Portal or the proxied server). I reckon it'll take some solid research to find solutions that might work. I don't really know what that solution might be (I'd have to re-familiarise myself with the current way downloads work, as well).

I'd suggest we sit on this PR for now. We can see if we can come up with a general solution during this iteration. If we can't then this PR will still be ready to go. I don't see the downloader code changing so much so I don't think this PR will get out of date in that time. It's good that we'll have this ready if we can't find a solution for the general case.

jkburges:
Let me comment on this in a few minutes...

jkburges:
> I don't see the downloader code changing so much

... except for https://github.com/aodn/aodn-portal/pull/1684

web-app/js/portal/cart/Downloader.js
```diff
@@ -84,6 +88,25 @@ Portal.cart.Downloader = Ext.extend(Object, {
...
+        if ($downloaderLink && $downloaderLink.length > 0) {
+            $downloaderLink.attr('src', url);
+        } else {
+            $downloaderLink = $('<iframe>', { id:'downloaderLink', src: url }).hide().appendTo('body');
+        }
+    },
+
+    _onDownloadRequestFailure: function(resp) {
+        log.error("Synchronous download request has failed with the following error " + resp.responseText);
+        Ext.Msg.alert(
```
jkburges: I'd prefer to see some sort of error indicator/msg next to the row in the download page, rather than a popup.  This could tie in nicely with work to follow on from https://github.com/aodn/aodn-portal/pull/1684.

jkburges:
I'm with @dnahodil actually.  This PR, plus https://github.com/aodn/aodn-portal/pull/1684, plus any pending work for #1624 I think should all be considered together.

Let's talk about this tomorrow before getting too jumpy with the big green button.

dnahodil:
> ... except for #1684
Haha. Yeah, apart from that :S

danfruehauf:
So @anguss00, I'm leaving this as is, and we'll revisit that sooner than later. Alright?

anguss00:
No problemo

jkburges:
Ok, now that the ebola crisis is over, should we have a chat about this PR and https://github.com/aodn/aodn-portal/pull/1684?  Who wants to come, when?

dnahodil:
14:00 today?

anguss00:
Yep, definitely

jkburges:
Ok let's do it.  Hopefully it won't take too long.

danfruehauf:
Yep (if I'm invited).

jkburges:
> if I'm invited

Of course.

anguss00:
Closing PR following the results of meeting

jkburges:
I reckon we can use something like this: http://johnculviner.com/jquery-file-download-plugin-for-ajax-like-feature-rich-file-downloads/

But, I think it should be planned/discussed with @pblain first...

anguss00:
Nice @jkburges, that plugin looks great and seems to cover all of our issues. I was looking at this plugin https://github.com/eligrey/FileSaver.js but that one looks better. Do you want to have a chat to @pblain now?

anguss00:
Also, looks like he's heading in a similar direction to your pr for issue #1620, using a cookie to provide download status

Issue #1626, opened 2015-03-13
Start/End Date on step 3 are switched


kereid:
Instance name: IMOS
            App version: 3.39.1
            Build date: Tue Jan 20 17:08:04 EST 2015
            Build number: 321
            Git ref: 5a01b135de233f242c01e5a595a069639074a2e8

Steps to reproduce:

Select any collection to add to map.

Select a from date and a to date, ie From: 2015/02/17 To: 2015/02/18

Proceed to step 3

What should happen

Filters for collection read start date: 2015/Feb/17 end date: 2015/Feb/18

What does happen:

They read end date: 2015/Feb/17 start date: 2015/Feb/18

dnahodil:
Yeah, this has bugged me for ages. It's *technically* correct but it is confusing and needlessly complicated.

PR #1678, opened 2015-03-30
Big filters refactor

dnahodil:


anguss00:
I'm going in, cover me Porkins

dnahodil:
Stay on target.

... and let me know if you have any questions.

anguss00:
:+1: will do

dnahodil:
Sorry, just had to squeeze one last commit in there.

web-app/js/portal/filter/BooleanFilter.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+Ext.namespace('Portal.filter');
+
+Portal.filter.BooleanFilter = Ext.extend(Portal.filter.Filter, {
+
+    getSupportedGeoserverTypes: function() {
+
+        return ['boolean'];
```
anguss00: just wondering, should these types be configised somewhere, to avoid them being hardcoded into each filter type?

web-app/js/portal/filter/NumberFilter.js
```diff
@@ -0,0 +1,67 @@
...
+
+        return this._getCql(this.getLabel());
+    },
+
+    _getCql: function(alternateLabel) {
+
+        var cql = String.format(
+            '{0} {1} {2}',
+            alternateLabel || this.getName(),
+            this._getOperator(),
```
anguss00: should we attempt to convert the operator into a human readable form here? For instance, >= becomes 'greater than or equal to'

anguss00:
Just two comments @dnahodil

anguss00:
There also seems to be a bit of an issue with temporal filters on wms layers. When you try to apply a temporal filter where there is no data (seemingly), the disallowed dates are not greyed out and you get a whole stack of geoserver errors (and a red triangle of doom)

anguss00:
Seems to be limited to SOOP-SST near real time

anguss00:
Sorry, actually occurring on a few other layers too

web-app/js/portal/filter/BooleanFilter.js
dnahodil: Good question! As I see it, there are basically two options here.

Option (1): We could have a mapping so that there is one place to go to find out which GeoServer types map to which filter types. That is essentially what we had [here](https://github.com/aodn/aodn-portal/blob/d4044e3de30ee35770e4dbd8b88c2e8ff9ef1b91/web-app/js/portal/filter/BaseFilterPanel.js#L131-L149). One method that knows all the GeoServer types that each Filter type can handle.

Option (2) is what I've tried here. Essentially we have the ability to "ask" each Filter type what GeoServer types it supports. I liked this approach because it keeps this information within the class it is applicable for. Filter.js still needs to have a list of it's subclasses (see https://github.com/aodn/aodn-portal/blob/1455eb6141e42c60c90191c563b5968d4f4c17c6/web-app/js/portal/filter/Filter.js#L90-L116) but at least it doesn't need to also know which types each subtype supports.

I'll be interested to see how this holds-up as the Portal evolves.

web-app/js/portal/filter/NumberFilter.js
dnahodil: Yeah, absolutely. I'm actually tackling that problem in my work for #1626

web-app/js/portal/filter/BooleanFilter.js
anguss00: Yep, I think I like option 2 more, much better from an encapsulation point of view. At the end of the day, the configuration for supported geoserver types is still in one location per filter type here if changes need to be made.

dnahodil:
Thanks for the review @anguss00, I know there was a lot to go through.

web-app/js/portal/filter/NumberFilter.js
anguss00: Sweet!

anguss00:
No worries, it was good to get a grip on how things are working now :+1:

PR #1680, opened 2015-03-31
open links in download popup in new tab

danfruehauf:
Trivial fix.


danfruehauf:
Bleh, not working on ffox.

PR #1682, opened 2015-03-31
Always show confirmation dialog

jonescc:


src/test/javascript/portal/cart/DownloadConfirmationWindowSpec.js
```diff
@@ -9,63 +9,63 @@ describe("Portal.cart.DownloadConfirmationWindow", function() {
...
-    describe('showIfNeeded', function() {
+    describe('show', function() {
 
         beforeEach(function() {
 
-            confirmationWindow.showIfNeeded(downloadUrl, downloadFilename);
+            confirmationWindow.show(downloadUrl, downloadFilename);
         });
 
         it('shows window if not shown', function() {
```
danfruehauf: How about changing all of those tests to just one? Something alone the lines of "shows window". Because we have to show it EVERY time.

jonescc:
@danfruehauf , changes made

danfruehauf:
@jonescc Thanks. I'll just test it.

PR #1685, opened 2015-03-31
Revert "remove unused instance name"

danfruehauf:
Reverts aodn/aodn-portal#1668


danfruehauf:
@dnahodil Better safe than sorry.

I'll rework that commit to make it work and understand thoroughly the implications - which I clearly didn't.

dnahodil:
No worries, we'll revert for now.

I think that PortalInstance class is the last remnants of an old way of loading different content (using the instance name). I reckon you'll be able to get rid of that class with this work, too.

PR #1687, opened 2015-03-31
Remove development mode landing page switch

dnahodil:
I'm suggesting we remove this switch as there have now been two bugs (that I know of) which have made it to master which wouldn't have if this switch hadn't been in place. Sorry, @pmbohm, I know why we added this but I don't think it's worth the clicks it saves any more.


danfruehauf:
:+1: 

Esp. after I managed to break the landing page without even noticing.

danfruehauf:
Easy to revert if anyone REALLY objects to that...

PR #1689, opened 2015-04-01
Fixes #1670

### LINKED ISSUE #1670: Tracking non-user events to Google Analytics
Spotted on *master*

**Steps to reproduce**
In trackUsage.js add the following to trackFilterUsage(...) to see what is sent to Google
``` javascript
console.log(
    OpenLayers.i18n('filtersTrackingCategory'),
    OpenLayers.i18n(actionKey),
    label,
    collection
);
```
Open the Portal and select the Argo data collection

**What does happen**
You can see that a non-user event is sent to Google Analytics:
> Filters Combo Platform number=Loading ... IMOS - Argo Australia Profiles

**What should happen**
The 'Loading ...' message should not be logged to Google Analytics as it is just a placeholder value

jonescc:
Only track change events and only then when the value has actually changed

### DIFF
--- a/src/test/javascript/portal/filter/ComboFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/ComboFilterPanelSpec.js
@@ -79,14 +79,14 @@ describe("Portal.filter.ui.ComboFilterPanel", function() {
         });
     });
 
-    describe('onSelected', function() {
+    describe('onActualChange', function() {
         it('tracks usage using google analytics', function() {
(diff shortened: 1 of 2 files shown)


jonescc:
Needs to be rewritten for new filter model

danfruehauf:
@jonescc Can I delete the branch?

jonescc:
There will be another fix in theory shortly.  But if you want to delete the branch in the interim, thats OK

PR #1693, opened 2015-04-02
Fixes #1670

### LINKED ISSUE #1670: Tracking non-user events to Google Analytics

jonescc:
Only track events when the combo filter value has changed.

Switched to change/select events which only get generated when the component is enabled.

### DIFF
--- a/src/test/javascript/portal/filter/ui/ComboFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/ui/ComboFilterPanelSpec.js
@@ -55,7 +55,7 @@ describe("Portal.filter.ui.ComboFilterPanel", function() {
         });
     });
 
-    describe('onSelected', function() {
+    describe('onChange', function() {
 
         beforeEach(function() {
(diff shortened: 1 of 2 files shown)


jonescc:
@dnahodil, had a look at extracting tracking into a FilterChangeTracking class but looked like more work than was justified for this issue and not sure it was any cleaner in the end due to the differing processing requirements for date filters for example.

PR #1694, opened 2015-04-02
Grails upgrade 2.4.4

danfruehauf:
Fully good to be merged.

Everything works as far as I'm concerned.

Don't forget to run `grails clean` before you run it with 2.4.4.


danfruehauf:
So, except for a single test that fails (for a reason, the code needs to be fixed - not the test) and... the day light saving tests that fail, everything else seems to work.

danfruehauf:
That's good to go.

It also runs the jasmine tests.

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
```diff
@@ -84,7 +84,7 @@ class BulkDownloadService {
 
             zipStream.putNextEntry new ZipEntry(filenameToUse)
 
-            def bytesCopied = IOUtils.copy(streamFromUrl, zipStream)
```
jkburges: This library isn't available anymore?

scripts/CollatePortalJavaScriptSource.groovy
```diff
@@ -6,7 +6,7 @@
  *
  */
 
-includeTargets << grailsScript("Init")
+includeTargets << grailsScript("_GrailsInit")
```
jkburges: We should have a look at replacing this custom script with one of the standard grails plugins for minification/compression etc.  IIRC, the reason we were blocked on that in the past was because of some weird behaviour with CSS includes in IE (bit hazy - maybe someone else can elaborate?), which we no longer support anyway.

src/groovy/au/org/emii/portal/proxying/ExternalRequest.groovy
```diff
@@ -7,7 +7,7 @@
 
 package au.org.emii.portal.proxying
 
-import org.apache.commons.io.IOUtils
+import org.codehaus.groovy.grails.io.support.IOUtils
```
jkburges: Can this be used above?

test/unit/au/org/emii/portal/AsyncDownloadControllerTests.groovy
```diff
@@ -19,7 +19,7 @@ class AsyncDownloadControllerTests extends ControllerUnitTestCase {
         controller.downloadAuthService = downloadAuthService.createMock()
 
         gogoduckService = mockFor(GogoduckService)
-        gogoduckService.demand.static.registerJob { params -> return "gogoduck_rendered_text" }
```
jkburges: Did that work before?

jkburges:
In general, this would be a really good change to merge in, and I'd be happy to see that.

A couple of general things though:

* supporting infrastructure will need to support grails 2.4.4 - e.g. jenkins, travis.  Travis is probably just a config setting, jenkins we have to chef the version ourselves
* are the majority of developers happy to upgrade their environment?  Should be easy if using gvm.

jkburges:
Actually, a better way to make sure the grails 2.4.4 dependency is *easily* met on all systems (dev, CI envs) is to use the grails wrapper, see: http://grails.github.io/grails-doc/2.4.4/guide/commandLine.html#wrapper

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
danfruehauf: Wanna help me to see how we can pull it down? I had a bunch of conflicts because grails /sort of/ implements it. So we can pull apache.commons.io, but we're using just this little function. Can be changed.

scripts/CollatePortalJavaScriptSource.groovy
danfruehauf: Mmmmmm. Yep. Definitely.

Grails 3.x includes a minifier as well. Not sure about 2.4.4...

danfruehauf:
Regarding the build environments - we're still using maven (both for travis and jenkins) - so this is not a problem.

And for development, you can use:
```
$ vagrant ssh dev
$ portal_run
```
It takes care of using grails 2.4.4.

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
```diff
@@ -145,4 +145,14 @@ class BulkDownloadService {
         zipStream.write bytes, 0, bytes.length
         zipStream.closeEntry()
     }
+
+    public static void copy(final InputStream input, final OutputStream output, final int bufferSize) throws IOException
```
danfruehauf: @jkburges This is the only function we need from apache.commons.io. So I took it. But we can also pull apache.commons.io - either way works for me. the grails `IOUtils.copy` function closes both streams when it finishes, which is not good for us.

test/unit/au/org/emii/portal/AsyncDownloadControllerTests.groovy
danfruehauf: Yeah, did work, but I think it messed up the other tests. It's being called once, so I think enough to mock it once...

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
jkburges: > IOUtils.copy function closes both streams when it finishes, which is not good for us.

I don't understand - this was the previous behaviour (?) but you're saying that it doesn't work?

jkburges: If it's a hassle to pull in the library, don't worry too much about it, although I'm always wary of implementing fairly low-level stuff like this ourselves, as I think it's always a bug breeding ground.  p.s. dare I say it, the code conventions are not met in the method, i.e. opening braces are on newlines, no space after `while`).

jkburges:
> we're still using maven (... for  jenkins)

We're definitely *also* using grails in jenkins.

You *can* use the vagrant development env, but is everyone using this (or would be happy to)?  It's an assumption that needs to be checked I think....

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
danfruehauf: I'll try to make some sense out of that.

 * `org.apache.commons.io.IOUtils.copy` - Does not close streams when it's done - good for us, this was the original behaviour.
 * `org.codehaus.groovy.grails.io.support.IOUtils.copy` - Included with this grails version. Does close the streams when it's done - not good for us.

Options:
 * Implement stream copying (a few lines) like apache.commons.io
 * Pull `org.apache.commons.io` as a dependency. I've found it to be rather painful, but I'm not a maven/grails expert.

jkburges: Ok, I understand.  Don't worry about it too much, I know what a PITA it can be to get grails dependencies to play nicely together.

danfruehauf:
If not, you can easily use `gvm`. It's pretty simple just as well.

If people can juggle 2 grails versions (already) for go-go-duck and portal, I don't think there's much of a difference, but correct me if I'm wrong.

Regardless of that, it's a requirement by @pblain to upgrade to 2.4.4, not my crazy doing ;)

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
danfruehauf: Hold on, I'll fix it. That's a valid point...

jkburges:
> If not, you can easily use gvm. It's pretty simple just as well.

There are plenty of ways to do it, that's not the issue.  My original comment "are the majority of developers happy to upgrade their environment?" still stands I think.  People are going to come in one morning and all of a sudden find that they are in no position to develop on the portal.  It's not a huge issue, but people at least need to be aware.  Maybe just send out an email to devs point them to either the VM based env or gvm?

> it's a requirement by @pblain to upgrade to 2.4.4

I'm really interested to know why this is so...

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
danfruehauf: I actually wanted to open a bug on that. I don't understand why the behaviour is not compatible with `org.apache.commons.io`. It's silly.

danfruehauf:
Yet again, this is a requirement set by @pblain.

I will send an email around once this is in, and I've just also updated the README with a section on how to install grails 2.4.4 using gvm (was missing before).

jkburges:
Do you get the following "error" when running:

```
| Error log4j:WARN No appenders could be found for logger (grails.plugin.webxml.WebxmlGrailsPlugin).
| Error log4j:WARN Please initialize the log4j system properly.
| Error log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```
?

danfruehauf:
Yeah, me too. I didn't dive into that... But maybe I should?

jkburges:
Favicon looks broken for me.  This branch on the left, production portal on the right...

image

jkburges:
> Yeah, me too. I didn't dive into that... But maybe I should?

I think it'd be good to clean it up... broken windows and all.

danfruehauf:
favicon is broken - confirmed. Don't merge it yet.

jkburges:
Seems to be all working for me, apart from those couple of minor issues.  Let me know when fixed and I'll merge...

danfruehauf:
I think the favicon is something I broke on master when I stripped the instance name. I'll fix it on this PR.

It will not break in deployed instances BTW...

danfruehauf:
@jkburges Pretty painful with those logging errors. The last commit fixes it, but I don't think I understand what I did.

grails-app/views/_core_theme_includes.gsp
```diff
@@ -10,5 +10,5 @@
 <link rel="stylesheet" type="text/css" href="${resource(dir: 'js', file: 'ext-3.3.1/resources/css/xBaseTheme.css')}"/>
 
 <g:if test="${grailsApplication.config.portal.instance?.name}">
```
jkburges: I had to remove this guard to get favicon to work...

jkburges:
`grails test-app` is broken for me, see: https://gist.github.com/jkburges/26911ce37c2e2e1b02c2

grails-app/views/_core_theme_includes.gsp
danfruehauf: Yep, my bad. I'll remove it.

jkburges:
`mvn test` works for me, which travis is already telling us.  As does `mvn jasmine:bdd`.

danfruehauf:
Bahhh, I think I'll just omit that commit and we can live with those error logs in dev mode :/

jkburges:
Yeah I'd be happy with that for now.

danfruehauf:
Should finally pass in travis.

jkburges:
Travis is passing - I don't think that's been a problem (at least since today) - it used `mvn test`.  But `grails test-app` is still broken for me, as is `grails interactive`.  I use both of these commands frequently when developing.

So we can probably un-revert the SLF4J error change at least, since that seems to be unrelated (if you can, it looks as though it's been squashed).

jkburges:
Btw, it's possible something is not being cleaned properly in my env, I've tried `grails clean` but I haven't tried blowing away maven/ivy caches as yet - perhaps I should if the above commands do work for you?

danfruehauf:
Should work, it works for me. Lets see what's not working when you're here.

danfruehauf:
Just tried it on a clean dev box - works well.

danfruehauf:
Sorry, not here for a month, here for less than 2 weeks. I thought I had an older PR with grails 2.2.0 but apparently I haven't.

jkburges:
Ok, after performing the following steps, this now works for me:

* grails clean
* rm -rf ~/.m2
* rm -rf target
* rm -rf ~/.grails
* rm -rf ~/.ivy2

I think it was the deletion of .iv2 that actually had an effect.  Others *may* need to do this, if not using the dev VM.

I can confirm that it *does work* also for me in the dev VM though.

jkburges:
> can probably un-revert the SLF4J error change at least, since that seems to be unrelated (if you can, it looks as though it's been squashed).

I think the change made previously (now squashed away) is unrelated to me having `grails test-app` fail - can we please bring it back in so that we don't see an error message (as in https://github.com/aodn/aodn-portal/pull/1694#issuecomment-91383123)?

danfruehauf:
I was about to suggest deleting the `.ivy2` dir. But yep, I'll put it in the email I'll send around. So GTG?

jkburges:
Just making a note that I'm happy (keen even) for this to be merged, but @danfruehauf and I don't want merging this to interfere with @kereid's testing, so we're holding off until that is done.

danfruehauf:
I think finally.

I'll modify the jenkins jobs accordingly and send the email out.

PR #1702, opened 2015-04-08
reset dates in NcWmsPanel (fix #1673)

### LINKED ISSUE #1673: Clear subset does not clear acorn date filter
systest.aodn.org.au
Instance name: IMOS
App version: 4.0.0
Build date: Thu Mar 26 13:33:50 AEDT 2015
Build number: 1
Git ref: 22cd9f53c67f198551731bb0a8c0facff69a6906

To reproduce, 

* select an acorn layer and go to step 2
* select a from or to date from the  list and then,
* hit the clear subset button.

What I expect to see is the changes I've made cleared. 

Instead, the dates remain modified.

danfruehauf:
reset dates to real min and max and not the current subset that's
displayed. bug was probably carried over from my work on stateless
ncwms

### DIFF
--- a/src/test/javascript/portal/details/NcWmsPanelSpec.js
+++ b/src/test/javascript/portal/details/NcWmsPanelSpec.js
@@ -217,8 +217,8 @@ describe('Portal.details.NcWmsPanel', function() {
                 return this.temporalExtent;
             },
             setSubsetExtentView: function() {},
-            getSubsetExtentMin: function() { return extent.min(); },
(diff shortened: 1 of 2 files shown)


dnahodil:
I just checked this out and it seems there might be a problem. It seems that when I change one date picker it resets the other. So selecting a 'from' date resets the 'to' date and vice versa. This doesn't happen on edge. Can you please check this out, @danfruehauf.

danfruehauf:
@dnahodil Bugger. Yep. I knew I probably did something wrong.

PR #1703, opened 2015-04-08
Step 3 filter tidy (#1626)

dnahodil:
This is a tidy-up of filters display in step 3.  It stems from #1626 but goes a lot further.

The main changes are:
- On step 3 all filters are displayed as `label: value`. (There was inconsistency with this.)
- The labels for spatial and temporal extent are now consistent between step 2 and step 3. (Previously they weren't).
- Regarding displaying of filter values on step 3 there was a mix of word-type display and raw CQL. This has now been made more consistent.
- There was inconsistency with labelling between ncWMS spatial filters and WMS spatial filters
- There was inconsistency with labelling between ncWMS temporal filters and WMS temporal filters
- Moved some text to the i18n dictionary
- Added a tooltip to symbols in the number filter so you can see in words what the symbol means

### DIFF
--- a/web-app/js/portal/lang/en.js
+++ b/web-app/js/portal/lang/en.js
@@ -83,9 +83,6 @@ OpenLayers.Util.extend(OpenLayers.Lang.en, {
     mapTabTitle: 'Map',
 
     // Search form
-    boundingBoxDescription: 'Bounding Box',
-    boundingBoxDescriptionNcWms: 'Bounding Area',
-    maxExtentOfPolygon: "Max extent of polygon",
     northBL: 'N',
     eastBL: 'E',
     westBL: 'W',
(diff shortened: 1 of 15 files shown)


dnahodil:
This is ready for review. I discussed this with Pete and we came up with some further suggestions (see #1705) but they can be dealt with separately.

dnahodil:
I have another bug fix PR that's waiting on this one. It'd be nice to get them both merged before we release to systest so they can make it into the next release.

anguss00:
I'll look at this

web-app/js/portal/lang/en.js
```diff
@@ -193,6 +190,25 @@ OpenLayers.Util.extend(OpenLayers.Lang.en, {
...
+
+    dateFilterBeforeFormat: '{0}: before {2}',
+    dateFilterAfterFormat: '{0}: after {1}',
+    dateFilterBetweenFormat: '{0}: {1} to {2}',
+
+    numberFilterOptionsFields: [
+         'code', 'text',                     'cql',                 'symbol'
+    ],
+    numberFilterDropdownOptions: [
+        ['CLR',  'none'], // This option is in the dropdown to clear the number filter
```
anguss00: Nice @dnahodil, this is really sweet

PR #1704, opened 2015-04-08
reset button in NcWmsPanel works

danfruehauf:
* reset to original extent when clicked on reset

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -15,7 +15,6 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
     PENDING_EVENT_ATTR: 'PENDING_EVENT',
 
     constructor: function(cfg) {
-
         this.layer = cfg.layer;
 
         var config = Ext.apply({
@@ -36,10 +35,9 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
     },
 
     _initWithLayer: function() {
-
         this.geoNetworkRecord = this.layer.parentGeoNetworkRecord;
 
-        this._clearDateTimeFields();
+        this._disableDateTimeFields();
         this._attachTemporalEvents();
         this._attachSpatialEvents();
         this._removeLoadingInfo();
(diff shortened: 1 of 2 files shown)


danfruehauf:
@dnahodil This should work.

As a side note I'll just that this has to be rewritten to use the new filter architecture.

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -15,7 +15,6 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
     PENDING_EVENT_ATTR: 'PENDING_EVENT',
 
     constructor: function(cfg) {
-
```
dnahodil: @danfruehauf 'developer discretion' means that you can choose how you want to write your own code (or code that you update). It *doesn't* mean you get to reformat every file you touch. What a huge waste of time that would be for everyone.

I mentioned this specific issue when we spoke about blank lines in code:
> The thing with developer discretion is we don't want one developer adding blank lines, and another removing them, and then someone else adding them back in.

If you feel strongly about this *please* raise it again in internal discussions and we can come to an agreement (I think the discussion will go in favour of no blank lines so you don't need to worry). Until it is sorted please add these blank lines back in for the purposes of this PR.

dnahodil:
This change has fixed the issue we were seeing but now the original issue is back (the 'clear subset' button doesn't work).

web-app/js/portal/details/NcWmsPanel.js
dnahodil: As a note I wouldn't let anyone apply a 'developer discretion'  code style like this (to a whole file even though they only touched a small part of it) even if the new code style was my preference, too. The point is not to reformat back-and-forth and it's not to apply your style to other people's work. If we've agreed on 'developer discretion' it means we all need to be happy with either.

danfruehauf: So if I change a function name, can I also get rid of the new line below it?

danfruehauf: From now on BTW, my 'developer discretion' will be to have 10 empty lines at the end of every function implementation.

Issue #1675, opened 2015-03-27
Two time filters require different titles


kereid:
Found in sys-test but is also in production:

Steps to reproduce:

Add AATAMS Facility - Tracking of Short-Tailed shearwaters collection to map (Platform >biological to narrow down) 

What should happen:

Taken to step 2 and see 2 time filters one labelled as the "time stamp" and the other as release date ( @danfruehauf confirmed this is what they are)

What does happen:

See Temporal extent with 2 date selection filters underneath and cannot tell why they are there

danfruehauf:
Thanks @kereid ! Good spotting!

PR #1715, opened 2015-04-14
Issue #1675

dnahodil:
This addresses issue #1675

It introduces a new concept of a primary filter. This is based on some upcoming work where the geographic filter will be above and outside all layers (aka accordion) as will the temporal extent. However, we support more than one temporal extent so when that work happens we'll need a way to determine which is the primary temporal filter (which should therefore go outside the normal list of filters). For now a time filter is the primary if it is the only one for that layer. I'm happy to talk about this if my idea needs clarification.


anguss00:
I'll review

PR #1718, opened 2015-04-15
Remove unused Filter getType() function

dnahodil:
This is left over from the filters prototype. Things have changed in such a way that this function isn't used anymore.


anguss00:
Looks good @dnahodil

PR #1719, opened 2015-04-16
Remove usage of undefined variable

dnahodil:
This should have been part of 6328f509cfb2532a16c3481fd3f683ec507e1afb. This code is replaced with this line https://github.com/aodn/aodn-portal/blob/0b3e2aaf46e48a0e7e8471f5080fea3c69e22017/web-app/js/portal/filter/ui/NumberFilterPanel.js#L30


danfruehauf:
This must be hotfixed on systest. It breaks the following (NOAA Drifters):
https://portal-systest.aodn.org.au/imos123/home?uuid=e3cf7cfd-9034-4c03-82ca-5a40ce7f96d2

danfruehauf:
@kereid FYI - this happens on current systest.

dnahodil:
Thanks @danfruehauf, I'll update systest.

PR #1722, opened 2015-04-17
1620 notifications

jkburges:


web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -86,10 +86,33 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
 
+    // It's *actually* a button, but we're using it as a label here...
+    _createDownloadingLabel: function(collection) {
+        var elementId = this._getButtonId(collection, 'downloadButtonId');
+
+        Ext.fly(elementId).update("");
+
+        new Ext.Button({
+            text: OpenLayers.i18n('downloadStatusRequested'),
+            icon: 'images/spinner_orange_background.gif',
```
jkburges: @pmbohm or anyone else - there's probably a better way to do this (than introducing an entirely new images with a (fixed) orange background) - but I don't know of it. Please feel free to change this if so.

jkburges: Or let me know how to do it...

danfruehauf:
So looks good to me, I just don't quite manage to reproduce the problem.

jkburges:
Reproducing the original problem is a bit hit and miss, due to the database caching and what not.  You can *simulate* a slow-to-start (or failed) download by inserting the following snippet in to `DownloadController`:

```groovy
    def index = {
        // render text: 'urlFieldName was not provided', status: HTTP_400_BAD_REQUEST

        Thread.sleep(10000);
        _performProxyingIfAllowed()
    }
```

web-app/js/portal/cart/DownloadPanelItemTemplate.js
pmbohm: You could use Font Awsome. Thats the purely css way, but the FA spinner is a bit jumpy looking. The class would most likely be 'fa-spinner' from (http://fortawesome.github.io/Font-Awesome/cheatsheet/)

Examples of this is in Portal.search.FacetedSearchResultsDataView
```
    _buildLabel: function(fontAwesomeClass, text) {
        return "<span class=\"fa fa-fw " + fontAwesomeClass + "\"></span> " + text;
    },
```

danfruehauf:
@pmbohm You mind finishing to review it? I got a bit preoccupied with other urgent stuff...

web-app/js/portal/cart/DownloadPanelItemTemplate.js
jkburges: Ok thanks - let me try that.

pmbohm:
@danfruehauf yes ok. Were you happy with the code? Only testing it to do?

web-app/js/portal/cart/DownloadPanelItemTemplate.js
jkburges: @pmbohm the only font awesome spinner is this one: fa-spinner [&#xf110;] which looks a bit different.

I think I might leave this as is for now, as:

1. I don't want to change every spinner as part of this PR
2. the fa-spinner doesn't seem to actually spin - it's just static (at least where I'm using it).

pmbohm: @jkburges 
I gave you a bum steer on the spinner. It turns out that the ' fa-spin' class must be added. 
The disabled class can be overridden. Below works for me


```
        new Ext.Button({
            text: "<span class=\"fa fa-spin fa-spinner \"></span>" + OpenLayers.i18n('downloadStatusRequested'),
            cls: 'navigationButton navigationButtonActive',
            scope: this,
            disabled: true,
            disabledClass: '',
            renderTo: elementId
        });
```
CSS
```
.navigationButtonActive.x-btn {
    background-color: #89C8EE;
}
```

web-app/js/portal/utils/Logging.js
```diff
@@ -9,3 +9,5 @@ var log = log4javascript.getLogger();
 
 var ajaxAppender = new log4javascript.AjaxAppender('system/clientLog');
 log.addAppender(ajaxAppender);
+
+log.addAppender(new log4javascript.BrowserConsoleAppender);
```
pmbohm: What does this have to do with the commit?

web-app/js/portal/cart/DownloadPanelItemTemplate.js
jkburges: The above works - but it still looks different (i.e. the spinner style). I don't think we want that, do we?

jkburges:
@pmbohm updated as discussed.

jkburges:
GTG again, @pmbohm

PR #1724, opened 2015-04-20
Add multiple collections with ctrl+click

dnahodil:
This is a little something that I think will be handy. It lets you ctrl+click on step 1 and add multiple collections without being taken to step 2 for each one.

Current bugs:
* There is a problem with map binding to a few UI controls (to reproduce just ctrl+click before you've added any other data collections)

Nice to have:
* There are a few page elements that move about when you ctrl+click add a data collection (mainly the mini-map, I think). It would be nice if they didn't move when you added collections (and just greyed-out).

### DIFF
--- a/web-app/js/portal/details/SubsetPanel.js
+++ b/web-app/js/portal/details/SubsetPanel.js
@@ -24,6 +24,7 @@ Portal.details.SubsetPanel = Ext.extend(Ext.Container, {
         }
         else {
             var filterGroupPanel = new Portal.filter.ui.FilterGroupPanel({
+                map: cfg.map,
                 layer: cfg.layer
             });
 
--- a/web-app/js/portal/details/SubsetPanel.js
+++ b/web-app/js/portal/details/SubsetPanel.js
@@ -10,31 +10,16 @@ Portal.details.SubsetPanel = Ext.extend(Ext.Container, {
 
     constructor: function(cfg) {
 
-        this.layer = cfg.layer;
-
-        var items = [];
-
-        if (this.layer.isNcwms()) {
-            var ncwmsPanel = new Portal.details.NcWmsPanel({
-                map: cfg.map,
-                layer: cfg.layer
-            });
-
-            items.push(ncwmsPanel);
-        }
-        else {
-            var filterGroupPanel = new Portal.filter.ui.FilterGroupPanel({
-                map: cfg.map,
-                layer: cfg.layer
-            });
-
-            items.push(filterGroupPanel);
-        }
+        var panelType = cfg.layer.isNcwms() ? Portal.details.NcWmsPanel : Portal.filter.ui.FilterGroupPanel;
+        var newPanel = new panelType({
+            map: cfg.map,
+            layer: cfg.layer
+        });
 
         var config = Ext.apply({
             title: OpenLayers.i18n('subsetPanelTitle'),
             hideMode: 'offsets', // fixes #1278
-            items: items
+            items: [newPanel]
         }, cfg);
 
         Portal.details.SubsetPanel.superclass.constructor.call(this, config);
--- a/web-app/js/portal/search/FacetedSearchResultsDataView.js
+++ b/web-app/js/portal/search/FacetedSearchResultsDataView.js
@@ -20,9 +20,7 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
 
         this.rowId = 0;
 
-        this.setTplSizeVariables();
-
-        var tpl = new Ext.XTemplate(
+        this.tpl = new Ext.XTemplate(
             '<tpl for=".">',
             '<div class="resultsHeaderBackground">',
             '    <div class="x-panel-header">',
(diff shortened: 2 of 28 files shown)


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -316,14 +307,16 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
...
+        var multiSelect = clickEvent.ctrlKey;
         var superUuid = btn.container.id.replace("fsSearchAddBtn", '');
-        this.addRecordFromSuperUuid(superUuid);
+        this.addRecordFromSuperUuid(superUuid, multiSelect);
+
+        btn.addClass("x-btn-selected");
     },
 
-    addRecordFromSuperUuid: function(superUuid) {
+    addRecordFromSuperUuid: function(superUuid, multiSelect) {
```
jkburges: The naming `super...` is a bit weird here, IMO.  It's the UUID of the metadata record, right?

web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -333,7 +326,10 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
 
             Portal.data.ActiveGeoNetworkRecordStore.instance().add(record);
         }
-        Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_GEONETWORK_RECORD, record);
+
+        if (!multiSelect) {
+            Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_GEONETWORK_RECORD, record);
```
jkburges: I'm not sure that firing a *global* `VIEW_GEONETWORK_RECORD` event is appropriate here?  Am I right in saying that the only action we currently take on such an event is to switch to step 2? Seems like it might be better to localise this event?  Maybe it's a pain to change this though.

jkburges: On a related note, what do you think about making the following changes to renaming `ACTIVE_GEONETWORK_RECORD_ADDED` simply to `COLLECTION_ADDED` and so on? Bit shorter and more descriptive I think.

web-app/js/portal/search/FacetedSearchResultsPanel.js
```diff
@@ -46,20 +45,6 @@ Portal.search.FacetedSearchResultsPanel = Ext.extend(Ext.Panel, {
...
-    _viewButtonOnClick: function(btn) {
-
-        btn.addClass("x-btn-selected");
-        var uuid = btn.container.id.replace("fsSearchAddBtn",'');
-        var record = this._getRecordFromUuid(uuid);
-
-        if (!Portal.data.ActiveGeoNetworkRecordStore.instance().isRecordActive(record)) {
-            Portal.data.ActiveGeoNetworkRecordStore.instance().add(record);
-        }
-        Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_GEONETWORK_RECORD, record);
```
jkburges: Was this event being fired twice prior to this change?

jkburges:
Apart from the comments (where maybe there can be some refactoring), I'd be happy to merge this if:

* the bug of ctrl-clicking the first collection is fixed;
* @pblain signs off on the change and
* @kereid signs off on the change.

web-app/js/portal/search/FacetedSearchResultsDataView.js
dnahodil: It was exposed to a neutron blast when it was a young UUID and now it has great power (and responsibility) and is a *super*UUID.

Also, from what I can tell there were cases where the same metadata records were being returned in the search results. So although their UUIDs were unique for each one, they weren't unique on the page. So the superUUID combines the row number to make them unique on the page.

I don't know why there would be duplicates in the search results. Perhaps that is a bug which has since been fixed. I will check with @pmbohm who I think made those changes.

jkburges: A non-unique UUID, one might say.

web-app/js/portal/search/FacetedSearchResultsDataView.js
dnahodil: I think you're right about that event being not-quite-right. You'll also notice it's the only even that isn't saying *this thing happened* rather it's saying *do this thing*. I'll have a look at whether we can adjust the scope of that.

Yeah, I reckon we can do that rename. We still refer to geonetwork records (e.g. ActiveGeonetworkRecordStore) but we can start moving to using the data collection terminology.

web-app/js/portal/search/FacetedSearchResultsPanel.js
dnahodil: Nah, this function wasn't being called as far as I could tell.

dnahodil:
I have addressed your review comments, @jkburges.

@pblain is happy with this, and I think @kereid is, too, but maybe she was just being polite.

web-app/js/portal/details/SubsetPanel.js
```diff
@@ -10,30 +10,16 @@ Portal.details.SubsetPanel = Ext.extend(Ext.Container, {
...
+        var newPanel = new panelType({
+            map: cfg.map,
+            layer: cfg.layer
+        });
 
         var config = Ext.apply({
             title: OpenLayers.i18n('subsetPanelTitle'),
             hideMode: 'offsets', // fixes #1278
-            items: items
+            items: [newPanel]
```
jkburges: This looks like an unnecessary level of nesting? Perhaps it is ok in this case because it is hiding from the parent components the messiness of switching between ncwms/normal...

dnahodil: Yeah, I think the idea behind this component is to isolate the switching between two panels, and to allow them to be as similar as possible (e.g. this component has the title so we don't need to set it in two places in sub components).

jkburges:
I think 10% should be presented in the review, btw.  @pblain agree?

dnahodil:
Yep. @pblain said the same thing when I checked this with him.

jkburges:
He's a great mind, obviously.

PR #1727, opened 2015-04-20
Change client side log config

jkburges:
**DO NOT MERGE**

This demonstrates how hierarchical loggers can be configured (as with log4j), as well as setting different levels for different loggers.

I wanted to put this up as a topic for discussion.

Currently, we don't log very much on the client side, despite the code there making up the majority of the code base for the portal.

I think we all find logging during development pretty useful (whether it be using the `log4javascript` framework, or simply `console.log()`.  Not sure how useful logging in production is or has been - I don't look there very often personally.

As @pmbohm rightly pointed out in a previous PR of mine, it's probably not that useful to flood the browser console with log messages, just like when logging in any other app.  So, I think to do this we need finer grained control over loggers and their levels, as this PR demonstrates.

Or maybe it's all just overkill and we shouldn't bother with (permanent) logging at all - just add/remove logs when developing at will.

### DIFF
--- a/web-app/js/portal/ui/MainPanel.js
+++ b/web-app/js/portal/ui/MainPanel.js
@@ -15,6 +15,12 @@ Portal.ui.MainPanel = Ext.extend(Ext.Panel, {
 
     constructor: function(cfg) {
 
+        this.logger = log4javascript.getLogger('Portal.ui.MainPanel');
+        this.logger.setLevel(log4javascript.Level.INFO);
+
+        log.debug('root Hello world!');  // This will be logged...
+        this.logger.debug('MainPanel Hello world!');  // ... but this will not.
+
         Ext.apply(this, cfg);
 
         this.addEvents('tabchange');
--- a/web-app/js/portal/utils/Logging.js
+++ b/web-app/js/portal/utils/Logging.js
@@ -5,7 +5,9 @@
  *
  */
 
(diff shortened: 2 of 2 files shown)


dnahodil:
It's a good point you make about the size of our JS code compared to the server-side which had much more logging. However, I think JS is a different beast and so some of the ideas about logging need to be rethunk.

One good thing about the nature of JS is that you don't have to think too much about logging beforehand because you can always add logging statements locally at runtime (unlike for server code) and there are also really good debugger and troubleshooting tools so logging is actually less important to see what is going on.

My main thought is that we should not be writing to the JS console unless it is absolutely necessary. We've had this mindset for a while now and I'm not sure we've ever needed the few console.log messages that we have accepted into the codebase. So, whatever we do I think that a normal user should not get logging to the JS console apart from very rare cases. I think those rare cases would be times when the Portal fails in a way that the UI can't handle. So maybe the suggestion from this is logging to console in dev mode and logging to server-side in prod.

danfruehauf:
So what would be the actual downsides of merging this? It seems like as @jkburges, client side logging is more essential to us than server side.

I don't think it's an overkill. I think it's something all of us will use.

jkburges:
Ref: https://github.com/aodn/aodn-portal/releases/tag/client_side_logging

PR #1729, opened 2015-04-20
fixes #1696

### LINKED ISSUE #1696: Number filter causes broken tiles
**Steps to reproduce**
Add a layer to the map which has a number filter (e.g. NOAA Drifters)
Choose the 'between' option in the drop-down of the number filter
Enter a number in the first text field but leave the second text field empty
Click off the form controls to apply the filter

**What does happen**
The tiles on the map break (red exclamation is shown)

**What should happen**
The filter shouldn't be applied if it will generate invalid CQL

pmbohm:
Fixes the possibility of invalid CQL causing the portal to mark the layer invalid when using the 'between' filter and leaving either the first or second field empty.

### DIFF
--- a/web-app/js/portal/filter/ui/NumberFilterPanel.js
+++ b/web-app/js/portal/filter/ui/NumberFilterPanel.js
@@ -48,7 +48,7 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
             }
         });
 
-        this.firstField = new Ext.form.TextField({
+        this.firstField = new Ext.form.NumberField({
             name: 'from',
             width: 146,
             listeners: {
@@ -58,7 +58,7 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
             }
         });
 
-        this.secondField = new Ext.form.TextField({
+        this.secondField = new Ext.form.NumberField({
             name: 'to',
             width: 146,
             hidden: true,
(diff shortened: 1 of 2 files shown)


web-app/js/portal/filter/ui/NumberFilterPanel.js
```diff
@@ -101,6 +101,12 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
 
         if (this.firstField.validate() && this.secondField.validate()) {
 
+            // Fixes #1696
+            if (this._operatorIsBetween() &&
```
dnahodil: I think it'd be nice to extract this into a function. Something like `_shouldUpdate` then the code here can just be:
```javascript
if (!this._shouldUpdate() {
    return;
}
```

Doing that would make this code more readable (i.e. you don't need to read that boolean logic to work out what cases it is trying to catch).

dnahodil: I also notice that we have logic like this further down (https://github.com/aodn/aodn-portal/pull/1729/files#diff-0e56b1024e0670eba5465784a3b77ebdR150). I don't think there's any need to have it twice, and where you have the logic (within `_updateFilter`) seems to be the better place for it. So we can remove 4 variables and an if-condition from the further down block and just rely on the logic you've got in _updateFilter.

dnahodil:
Changes look good, @pmbohm. I've just made a couple of suggestions including some code we can remove because your change has made it redundant.

dnahodil:
Hey @pmbohm. I've made some more comments on this.

pmbohm:
@dnahodil Your changes are in. Nice spotting of the validate() calls. It led to discovering the validation always succeeded. The field are now NumberFields! and crap cant be entered anymore.

dnahodil:
Oh cool.

Changes look good, @pmbohm. Will merge when Trav is happy.

PR #1731, opened 2015-04-21
Cleanup of CSS

pmbohm:
As the portal is now stateless the Grails style sheets can be removed.
mobile.css was probably never ever used.

During previous work to allow externalised branding the IMOS theme CSS was renamed to default.css which caused confusion as there is also general.css.
General.css should be the only file required to get a functional portal, while imosTheme.css is the look we want for the IMOS or AODN portal. This pull request also contains the few shuffling of rules to make this work. 

Further work is required to remove imosTheme.css from the portal code so that when our code is forked the IMOS look is not part of the code base.


dnahodil:
Very nice work @pmbohm. It's great to get rid of these files we are no longer using! :+1:

PR #1734, opened 2015-04-23
track gogoduck downloads with GA

danfruehauf:
Request by @pblain.

The main difference is that we will have the aggregation parameters - and we can actually understand what people are doing with GGD.

### DIFF
--- a/web-app/js/portal/utils/trackUsage.js
+++ b/web-app/js/portal/utils/trackUsage.js
@@ -34,3 +34,11 @@ function trackFiltersUsage(actionKey, label, collection) {
     );
 }
 
+function trackDownloadUsage(action, collection, downloadParams) {
+    trackUsage(
+        OpenLayers.i18n('downloadTrackingCategory'),
+        action,
+        collection,
+        downloadParams
+    );
+}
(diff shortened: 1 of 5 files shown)


web-app/js/portal/utils/trackUsage.js
```diff
@@ -34,3 +34,11 @@ function trackFiltersUsage(actionKey, label, collection) {
     );
 }
 
+function trackDownloadUsage(actionKey, collection, downloadParams) {
+    trackUsage(
+        OpenLayers.i18n('downloadTrackingCategory'),
+        OpenLayers.i18n(actionKey),
```
dnahodil: This doesn't need to load from the dictionary as we've already done that. So you can rename the parameter just to 'action' (as it isn't a key) and remove this call to i18n().

dnahodil:
Changes look good @danfruehauf, just one comment

web-app/js/portal/utils/trackUsage.js
danfruehauf: You're 100% correct.

danfruehauf:
Pushed, let the travis do its magic...

danfruehauf:
@dnahodil Ready for you.

PR #1739, opened 2015-04-26
gogoduck link

danfruehauf:
* Server side parsing of GGD returned JSON + formatting HTML returned to client (together with https://github.com/aodn/go-go-duck/pull/100)
 * Client side displaying of the returned HTML in the alert box

This will enable users to have a link in the alert box after a job submission was successful.

### DIFF
--- a/grails-app/services/au/org/emii/portal/GogoduckService.groovy
+++ b/grails-app/services/au/org/emii/portal/GogoduckService.groovy
@@ -3,6 +3,7 @@ package au.org.emii.portal
 import grails.converters.JSON
 import groovyx.net.http.HTTPBuilder
 import groovyx.net.http.HttpResponseException
+import org.apache.http.util.EntityUtils
 
 class GogoduckService extends AsyncDownloadService {
 
@@ -18,15 +19,16 @@ class GogoduckService extends AsyncDownloadService {
             throw new Exception("No parameters passed to gogoduckService")
         }
 
-        _gogoduckConnection().post(
-            [
-                body: _roundUpEndTime(jobParameters),
-                requestContentType: groovyx.net.http.ContentType.JSON
-            ],
-            successHandler
-        )
+        def responseText
 
-        return "GogoDuck job registered"
+        _gogoduckConnection().post([
+            body: _roundUpEndTime(jobParameters),
+            requestContentType: groovyx.net.http.ContentType.JSON
+            ]) { response ->
+                responseText = EntityUtils.toString(response.getEntity());
+            }
+
+        return responseText
     }
 
     def _gogoduckConnection() {
@@ -34,10 +36,6 @@ class GogoduckService extends AsyncDownloadService {
         def registerJobUrl = "${grailsApplication.config.gogoduck.url}/job/"
 
         return new HTTPBuilder(registerJobUrl)
-    }
-
-    def successHandler = { response, reader ->
-        log.debug "GoGoDuck response: ${response.statusLine}"
     }
 
     // This is to compensate for a lack of precision in the timestamps that NcWMS publishes
--- a/web-app/js/portal/cart/Downloader.js
+++ b/web-app/js/portal/cart/Downloader.js
@@ -113,10 +113,18 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
         });
     },
 
-    _onAsyncDownloadRequestSuccess: function() {
+    _onAsyncDownloadRequestSuccess: function(response) {
+        var serviceMessage = "";
+        if (this.params.serviceResponseHandler) {
+            serviceMessage = this.params.serviceResponseHandler(response.responseText);
+        }
+
         Ext.Msg.alert(
             OpenLayers.i18n('asyncDownloadPanelTitle'),
-            OpenLayers.i18n('asyncDownloadSuccessMsg', { email: this.params.emailAddress })
+            OpenLayers.i18n('asyncDownloadSuccessMsg', {
+                email: this.params.emailAddress,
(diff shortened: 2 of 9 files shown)


grails-app/services/au/org/emii/portal/GogoduckService.groovy
```diff
@@ -56,4 +54,26 @@ class GogoduckService extends AsyncDownloadService {
...
+
+    private String parseResponse(response) {
+        try {
+            def responseBody = responseBody(response)
+            log.debug "GoGoDuck response: ${responseBody}"
+            def responseBodyJSON = JSON.parse(responseBody)
+
+            def statusLink = responseBodyJSON.statusLink
+            if (!statusLink) {
+                throw "Could not parse statusLink"
```
dnahodil: This look invalid to me. Probably need to make this an Exception?

grails-app/services/au/org/emii/portal/GogoduckService.groovy
```diff
@@ -56,4 +54,26 @@ class GogoduckService extends AsyncDownloadService {
...
+            }
+
+            return "<a class='external' target='_blank' href='${statusLink}'>Follow the progress of your job</a><br /><br />"
+        } catch (Exception e) {
```
dnahodil: Catch should be on a new line

web-app/js/portal/cart/Downloader.js
```diff
@@ -129,5 +131,14 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
...
         return str.replace(/:/g, "#").replace(/[/\\ ]/g, "_");
+    },
+
+    _getServiceMessage: function(response) {
+        if (response) {
+            return response;
+        }
+        else {
+            return "";
+        }
```
dnahodil: This can be shortened to just `response ?: ""`. Possibly making this function redundant?

dnahodil:
I haven't run this up yet but my main thought is that I'm worried that we're generating HTML on the server-side (see https://github.com/aodn/aodn-portal/pull/1739/files#diff-10581fd966ae8fee7d80f4e29b8787bdR73). Maybe it would be more tidy to return JSON from the Portal server to the client-sode and have the client-side build the HTML? Among other benefits that would allow us to put the link text into the i18n dictionary.

danfruehauf:
I was thinking that too, but then comes the question of whether the client side should have any GGD specific knowledge. I tend to say no. Imagine we plug into that more of those services, say... AODAAC :)

At first I implemented it with JS parsing the lot. But then something felt wrong. So I did it this way, and still something feels wrong - as you noted.

dnahodil:
Could part of the information we pass back from the portal to the client-side be a "service key"? Then that name could be combined with a known dictionary key to get the appropriate message (if there is one).

And the dictionary would have 
gogoduckStatusLink: 'Job created, to check your progress use thislink {0}'
jodaacStatusLink: 'Something JODAAC-related'

danfruehauf:
That's an option. I'll look into that.

danfruehauf:
@dnahodil So now the logic of parsing is in the client side, the server side just passes the response as is.

The parsing is done in `GogoduckDownloadHandler`, as it should. It's passing a callback to handle the returned response and embed another piece of HTML in the alert box.

web-app/js/portal/cart/Downloader.js
```diff
@@ -113,13 +113,25 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
         });
     },
 
-    _onAsyncDownloadRequestSuccess: function() {
+    _onAsyncDownloadRequestSuccess: function(response) {
+        var serviceMessage = "";
+        if (this.params.serviceResponseHandler) {
+            serviceMessage = this.params.serviceResponseHandler(response.responseText);
+        }
```
dnahodil: I think you might have intended to replace this with a call to _getServicemessage()?

dnahodil:
Just one more comment @danfruehauf 

And sorry again for deleting your PR :sob:

web-app/js/portal/cart/Downloader.js
danfruehauf: Na, this is fine now. I can break it to another function if you really want me to, however it'll entail even bit more changes here and there as the scope passed to this function is `params` rather than `this` as the Downloader.

web-app/js/portal/cart/Downloader.js
```diff
@@ -105,21 +105,27 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
...
             OpenLayers.i18n('asyncDownloadPanelTitle'),
-            OpenLayers.i18n('asyncDownloadSuccessMsg', { email: this.params.emailAddress })
+            OpenLayers.i18n('asyncDownloadSuccessMsg', {
+                email: options.params.emailAddress,
+                serviceMessage: this._getServiceMessage(options.params.serviceResponseHandler, response.responseText)
+            })
         );
     },
 
+    _getServiceMessage: function(serviceResponseHandler, response) {
```
danfruehauf: Eventually did factor it to a new function. I think it's clearer.

dnahodil: Yeah, this looks good

dnahodil:
Happy to merge. Do I need to wait for https://github.com/aodn/go-go-duck/pull/100 ?

danfruehauf:
Let me double check, generally speaking it should be fine. But I prefer to test it with GGD that doesn't return the url link.

dnahodil:
Ok

danfruehauf:
Yep, good to go.

PR #1740, opened 2015-04-28
do not post unnecessary parameters for GGD jobs

danfruehauf:
do not post unnecessary parameters for GGD jobs

It can probably be refactored to look a bit nicer, such as binding the success function with the correct parameters, however I'm not too fussed about it.

### DIFF
--- a/web-app/js/portal/cart/Downloader.js
+++ b/web-app/js/portal/cart/Downloader.js
@@ -103,21 +103,21 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
     _downloadAsynchronously: function(collection, downloadUrl, params) {
         log.debug('downloading asynchronously', downloadUrl);
 
+        this.params = params;
         Ext.Ajax.request({
             url: downloadUrl,
             scope: this,
-            params: params,
             success: this._onAsyncDownloadRequestSuccess,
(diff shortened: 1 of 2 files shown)


web-app/js/portal/cart/Downloader.js
```diff
@@ -103,21 +103,21 @@ Portal.cart.Downloader = Ext.extend(Ext.util.Observable, {
     _downloadAsynchronously: function(collection, downloadUrl, params) {
         log.debug('downloading asynchronously', downloadUrl);
 
+        this.params = params;
         Ext.Ajax.request({
             url: downloadUrl,
             scope: this,
-            params: params,
             success: this._onAsyncDownloadRequestSuccess,
```
danfruehauf: @dnahodil Do you know of a nicer way to bind parameters to this success function? Something along the lines of... `this._onAsyncDownloadRequestSuccess.apply(this, params)`?

dnahodil: You could make the callback function an anonymous function which calls the function we want. ie. `function() { this._onAsyncDownloadRequestSuccess(params) }`. That will work because `params` is in scope for the anonymous function even thought it won't get called until later.

danfruehauf: Might just do it, for the sake of cleanliness.

PR #1741, opened 2015-04-29
refactor Downlader call to GGD

danfruehauf:
not binding params to `this` anymore. it does make the testing a bit
uglier, but the code cleaner.


dnahodil:
It makes the tests a bit uglier but I think this is a good overall change.

Issue #1723, opened 2015-04-20
Adding forward slash to home page causes error


pmbohm:
### To Reproduce
Link to step 1 directly but add a forward slash after home
http://[ aportalinstance ]/aodn-portal/home/

### What happens
The page is broken

### What should happen
The request with a forward slash should be handled the same as without a forward slash

PR #1743, opened 2015-04-30
Issue 1723

pmbohm:


pmbohm:
fixes #1723 and future issues

grails-app/views/_js_includes.gsp
```diff
@@ -6,6 +6,12 @@
 
 --%>
 
+
+<base href="${grailsApplication.config.grails.serverURL}" >
```
dnahodil: Is this needed? Removing it didn't seem to break anything for me.

grails-app/views/header/_mainPortalHeader.gsp
```diff
@@ -9,8 +9,8 @@
 <div id="header">
 
     <div id="logoContainer">
-        <a href="landing"><img src="${portalBranding.logoImage}" alt="logo" id="logo" />
-        </a>
+        <g:link controller="landing"><g:img uri="${resource(base: portalBranding.brandingBase, file: portalBranding.logoImage)}" alt="logo" id="logo" />
```
danfruehauf: I'm not reviewing this PR, but it seems like this logic of link generation should be in `portalBranding` if at all s.t. it returns a fully qualified link from `portalBranding` if needed.

pom.xml
```diff
@@ -332,7 +334,7 @@
                         <include>portal/ui/VisualisePanel.js</include>
                         <include>portal/ui/MainPanel.js</include>
                         <include>portal/utils/moment.min.js</include>
-                        <include>portal/utils/Proxy.js</include>
+                        <!--<include>portal/utils/Proxy.js</include>-->
```
dnahodil: If this isn't needed then we should remove it, not just comment it out.

src/test/javascript/spec_helper.js
```diff
@@ -5,19 +5,19 @@
  *
  */
 
+var BASE_URL = "${grailsApplication.config.grails.serverURL}";;
```
dnahodil: Is this one still needed?

dnahodil:
I have made a few small comments, @pmbohm. Nothing major just a few things that could be tidied-up.

However, my big comment/question is this:
I believe these changes fix the problem (I did some testing) however I don't believe the problem is big enough to justify a change which affects so many files and functions of the Portal (every URL we use/make, essentially). If we had found a one-liner in UrlMappings.groovy or an Apache config change which could fix it without having to make changes throughout the Portal then I would be much happier. Would you agree that this is changing a lot of files to fix a relatively small bug, @pmbohm?

danfruehauf:
@dnahodil I'm with you on this one. I suspect that this change can create more problems down the track which we are not aware off. It does indeed fix a relatively small problem overall.

I think we should try and find a smaller one-line fix because it should exist.

danfruehauf:
In addition this change also breaks other things. For instance I have dev.aodn.org.au -> 172.X.X.X. This change makes all the requests go to 127.0.0.1 and breaks everything.

danfruehauf:
Furthermore, also portal 2 suffers from the same:
http://portal.aodn.org.au/aodn/home
http://portal.aodn.org.au/aodn/home/

dnahodil:
This issue has been fixed with #1746

PR #1746, opened 2015-05-03
issue 1723 proof of concept fix

danfruehauf:
**DO NOT MERGE**

Hardcoded and ugly, but a significantly simpler fix I think.

@pmbohm @dnahodil

### DIFF
--- a/grails-app/views/home/index.gsp
+++ b/grails-app/views/home/index.gsp
@@ -28,6 +28,11 @@ The AODN/IMOS Portal is distributed under the terms of the GNU General Public Li
         <g:render template="/public_theme_includes"></g:render>
 
         <script type="text/javascript">
+            // Fixes #1723, redirect 'home/' to 'home'
+            if (window.location.href.match(/\/home\/$/)) {
+                var homeWithoutTrailingSlash = window.location.href.substring(0, window.location.href.length - 1);
+                window.location.replace(homeWithoutTrailingSlash);
+            }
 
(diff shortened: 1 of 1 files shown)


dnahodil:
I'm happy to use this as the fix for #1723 @danfruehauf. Is there anything you'd want to tidy about it first?

Also, we should get @kereid to add a test case for this.

danfruehauf:
I thought we could tidy it up a bit in terms of not having `home` hardcoded etc. Not sure. If you think it's tidy enough then yeah, lets go for it...

dnahodil:
I don't think tidying it will make much difference.

I have just noticed where this new code is. I don't think this belongs in the JS-includes file as it is functional code, not just an include. Maybe in home/index.gsp would be better? It could go before the browser check code.

danfruehauf:
Either is good. There's also `common/helpers.js` where this kind of browser specific code used to be. I don't feel strongly about where this code sits TBH.

dnahodil:
I wouldn't suggest helpers.js. Unless you have better ideas let's put it in home/index.gsp

danfruehauf:
I'll move it then.

danfruehauf:
BTW, unlikely to be a problem, but it'll also match:
```
http://something.com/aodn-portal/home/home/
http://something.com/aodn-portal/anothercontroller/home/
```

dnahodil:
Yeah, shouldn't be a problem IMO.

Issue #1749, opened 2015-05-07
High detail satellite or terrain base layer


keaves:
Enhancement request: 
IMAS has a decent amount of coastal and estuarine data that would greatly benefit from a high detail satellite or terrain baselayer option.

![image](https://cloud.githubusercontent.com/assets/10750046/7509099/93ad5eba-f4cf-11e4-97ec-0a60c0d2aa27.png)
A third option here would be perfect.

danfruehauf:
@keaves It's funny you mention. I was just looking at an option for doing that and it requires virtually no work.

I'll push the changes in a second.

danfruehauf:
@keaves Have a look at #1750. This adds the blue marble layer from the osgeo server.

The PR itself might not be accepted as a whole because we don't necessarily like to have default configuration relying on external services. However also if it's not accepted, @markhepburn can use it as a reference to configure a different set of base layers on your servers.

keaves:
@danfruehauf That's awesome, thank you! I'll keep a watch on the progress of #1750 . Hopefully you can include the opengeo maps. Otherwise the work is already done for Mark. Much appreciated

PR #1750, opened 2015-05-07
Bluemarble baselayer

danfruehauf:
Addresses #1749. I played with this thing accidentally, regardless of the ticket opened by @keaves


dnahodil:
Let's hold off on including the Blue Marble layer for now. We should just check it's okay for us to use it (licenses, etc.)

danfruehauf:
Yep. I might strip the second commit though so we can still have base layers from multiple geoservers.

dnahodil:
Yeah, that feature will still be handy.

PR #1751, opened 2015-05-08
allow multiple base layer geoservers

danfruehauf:
@dnahodil Just this PR now.


danfruehauf:
@keaves and @markhepburn - this will make it to the next round of testing. So with some config changes it should be easy to include baselayers from other geoservers.

keaves:
@danfruehauf - appreciated, thanks.

Issue #1730, opened 2015-04-21
Map disappears


pblain:
Chrome: Version 41.0.2272.101

Steps to reproduce:
1. Step 1: Add a data collection
2. Step 2: Create a subset (anything - doesn't matter)
3. Step 3: Download the subset (or something else).  The important thing is that the download bar appears at the bottom of chrome (showing the file name)
4. Go back to Step 2
5. It may also be necessary to click the X at the far right of the download bar (to close the bar)

What should happen:
1. Not much. 

What does happen:
1. The map disappears

![image](https://cloud.githubusercontent.com/assets/2802359/7245653/3d3e2978-e832-11e4-96e4-bf7ce832157a.png)

anguss00:
Also present in version 42.0.2311.90, but need to close the download bar

danfruehauf:
Maybe we should update OpenLayers?

anguss00:
Could be a good idea. I've investigated this issue and it is a bug between OpenLayers and google chrome. See https://github.com/openlayers/openlayers/issues/1181 and https://github.com/openlayers/openlayers/issues/1207. The issue is caused (I think) by the window resizing that occurs when the download bar is removed in Chrome. The workaround suggested is adding the following to CSS: 

svg {
-webkit-transform: none !important;
}
.olTileImage {
-webkit-backface-visibility: visible !important;
-webkit-transform: none !important;
}

danfruehauf:
As much as this will probably fix it, I think that can also be a good reason to investigate an upgrade of OpenLayers. We're a major version behind after all.

anguss00:
Yep, I'd agree with that for sure. I'd prefer to update rather than introduce a hacky fix.

danfruehauf:
Lets just note that this may actually be "more than a bug" in our definition. Can get quite involved.

anguss00:
Yeah, agreed. If we want to fix this properly it's more than a bug

anguss00:
[...]

PR #1754, opened 2015-05-11
Openlayers 2.13.1

danfruehauf:
**Not to merge yet**

OpenLayers 2.13.1 upgrade. Includes:
 * OpenLayers 2.13.1 source code, replacing the 2.10 version (first commit)
 * Various fixes such as adjusting tests, fixing i18n
 * Remove unused class TemporalMap
 * Remove IE compatibility code
 * All OpenLayers CSS customizations are now externalized to the OpenLayers 2.13.1 directory
 * Shows error when drawing  a bbox over the ante meridian (because otherwise it'd be possible)


danfruehauf:
When reviewing that, do not review the first giant commit. That's the OpenLayers 2.10 -> 2.13.1 replacement.

And this also fixes #1730.

dnahodil:
We need to remember to update @kereid's test cases with the new open layers. Filtering over the anti-meridian, for instance, will need some tests devised.

danfruehauf:
Will do.

Although we first need to think /what/ we are going to do with this new feature. We might decide to prohibit users from drawing those boxes...

danfruehauf:
I think... lets have someone review this now because it's already a fair few commits. I think this is merge-able considering the only problem is when drawing a box over the anti-meridian. This can be addressed in another issue and still needs some thought put into it. So... Ready to be reviewed.

dnahodil:
I would say that we can't merge this at the moment. We either need to fix filtering and downloads with a  geogrpahic boundary over the antimeridian (for both WMS and ncWMS) *or* we need to do something to prevent those boxes being drawn. It might seem odd, buthaving downloads which are silently incorrect is a worse bug than the current situation (i.e. not being able to draw a box over the antimeridian - but at least that's obvious).

It can certainly start being reviewed.

danfruehauf:
What did I break now...

danfruehauf:
@pmbohm GTG.

pmbohm:
@danfruehauf merged

PR #1755, opened 2015-05-12
font-awesome icons where possible. Ext buttons look and behave in line with Ext tab headers. Mouse over highlighting of buttons.

pmbohm:


web-app/css/general.css
```diff
@@ -1118,3 +1117,8 @@ div.tool-plus-disabled {
     opacity: 0.8;
     font-family: FontAwesome! important;
 }
+
+.fa-lg.fa-angle-double-right,
```
danfruehauf: I think be consistent here and have them both on the same line.

web-app/js/portal/search/GeoSelectionPanel.js
```diff
@@ -33,30 +33,37 @@ Portal.search.GeoSelectionPanel = Ext.extend(Ext.Panel, {
...
+            height: 280,
+            width: 300
         });
 
         var config = Ext.apply({
             layout:'form',
             cls:'search-filter-panel filter-selection-panel',
+            defaults: {
+                style: {
+                    margin: '2px'
```
danfruehauf: Can this be moved to CSS potentially?

web-app/js/portal/search/GeoSelectionPanel.js
```diff
@@ -33,30 +33,37 @@ Portal.search.GeoSelectionPanel = Ext.extend(Ext.Panel, {
         this.facetMap = new Portal.search.FacetMapPanel({
             initialBbox: Portal.app.appConfig.portal.initialBbox,
             mainMap: cfg.mapPanel,
-            height: 250,
-            width: 250
+            height: 280,
+            width: 300
```
danfruehauf: Any reason this gets "fatter"?

web-app/js/portal/search/GeoSelectionPanel.js
pmbohm: This is the better way to go in this particular case I think. I prefer to have less CSS rules created as they sometimes have adverse outcomes. Very specific CSS rules also go stale as you saw not too long ago. Also using Id's are not the best way to go as we latter shuffle code

web-app/js/portal/search/GeoSelectionPanel.js
pmbohm: yes, it now fits the space better

web-app/css/general.css
pmbohm: What consistency am i following? This is readable to me

pmbohm:
@danfruehauf review comments addressed.

web-app/css/general.css
danfruehauf: All the rest of the times you have both in the same line. I don't think it's critical, it's just inconsistent.

PR #1756, opened 2015-05-12
Remove some code which is no longer being used

dnahodil:
FilterType (not required now that filters come from GeoServer)
FilterUnitTestCase (a superclass for tests of Grilas filters, which we no longer have)
StringCharArrayFixer (fixed a bug in Grails 1.3.7 which we're no longer using)


danfruehauf:
:+1:

Issue #1748, opened 2015-05-06
Metatdata information not displayed correctly


kereid:
Steps to reproduce:

Select satellite from platform facet list.
Add any SST L3S collection to the map.
On step 2 click the info tab 

What should happen:

See all text as it should read in particular the degrees symbol for lat/long

What does happen:

Degrees symbol is displaying as diamond with question mark inside
![degrees diamond](https://cloud.githubusercontent.com/assets/2906602/7505721/6435153c-f499-11e4-9c1d-1bf047c3b3c0.jpg)

keaves:
Seems to happen with the apostrophe as well
![image](https://cloud.githubusercontent.com/assets/10750046/7508632/dc9144e2-f4c6-11e4-8268-cb843b7cddee.png)

anguss00:
I'll have a look at this

danfruehauf:
This bug happened in systest and production, but NOT on edge and when running locally. Doesn't happen in vagrant.

After some thorough investigation by @anguss00 and myself, including the following:
 * Compare Jenkins builds
 * Compare OS versions of libraries
 * Compare apache configurations
 * Compare tomcat configurations
 * Compare environment variables tomcat is running with

The conclusion is that the culprit is having `LC_ALL=C` in systest and production in the tomcat environment. Now we need to find where this is coming from...

danfruehauf:
`LC_ALL=C` is being added ONLY when chef is restarting the tomcat container on this node. Restarting the tomcat container via regular SSH login will result in no `LC_ALL=C`.

danfruehauf:
Bug opened in Chef:
https://github.com/chef/chef/issues/3367

danfruehauf:
This is a rather subtle bug I think. I had a lengthy discussion about it in the chef IRC channel. We're setting just `LANG=en_AU.utf8` but not `LC_ALL`. Chef then overrides with `LC_ALL=C`.

In chef 12.x there is a method of probing `LC_ALL` as long as you are running on `LANG=en_*`. Chef 11 does not support it. The reason edge has not suffered from that issue is because it was reinstalled with Chef 12.x.

Given that, I think this is a good reason to bulk upgrade to Chef 12.x.

Funny how this subtle bug just propagated all the way through.

anguss00:
[...]

PR #1761, opened 2015-05-14
Determine character encoding of metadata xml stream before escaping

anguss00:
Fixes issue #1748 in conjunction with chef upgrade

### DIFF
--- a/grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
+++ b/grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2015 IMOS
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ *
+ */
+
+package au.org.emii.portal
+
+final class HtmlCharacterEntityReferences {
+
+    /**
+     * Taken from Spring source code
+     * https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/util/HtmlCharacterEntityReferences.java#L132
+     */
+
+    /**
+     * Return the reference mapped to the given character or {@code null}.
+     * @since 4.1.2
+     */
+    public String convertToReference(char character, String encoding) {
+        if (encoding.startsWith("UTF")){
+            switch (character){
+                case '<':
+                    return "&lt;";
+                case '>':
+                    return "&gt;";
+                case '"':
+                    return "&quot;";
+                case '&':
+                    return "&amp;";
+                case '\'':
+                    return "&#39;";
+            }
+        }
+        else if (character < 1000 || (character >= 8000 && character < 10000)) {
+            int index = (character < 1000 ? character : character - 7000);
+            String entityReference = this.characterToEntityReferenceMap[index];
+            if (entityReference != null) {
+                return entityReference;
+            }
+        }
+        return null;
+    }
+}
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -9,8 +9,6 @@ package au.org.emii.portal
 
 import au.org.emii.portal.wms.NcwmsServer
 import au.org.emii.portal.wms.GeoserverServer
-import org.springframework.web.util.HtmlUtils
-import org.xml.sax.SAXException
 
 import grails.converters.JSON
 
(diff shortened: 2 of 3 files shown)


dnahodil:
Looks good @anguss00

anguss00:
:+1:

grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2015 IMOS
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
```
danfruehauf: That's probably wrong. It's only OK if also spring is GPL as well. Otherwise - you cannot do that.

dnahodil: Erg. Yeah, I didn't think of that. We'd better check that on Monday

danfruehauf:
I had a couple of comments about it. Given it's going to be revisited:
 * Is `grails-app/utils` the best place to have those things? I was thinking `src/groovy` or `src/java`?
 * Why not have those 2 functions in the same file? Potentially they can also be slightly simplified.

grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
anguss00: After researching this, I found that Spring Framework is distributed according to the Apache License 2.0, with the following re-distribution statement: 

**Redistribution.** You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without modifications, and in Source or Object form, provided that You meet the following conditions:

You must give any other recipients of the Work or Derivative Works a copy of this License; and
You must cause any modified files to carry prominent notices stating that You changed the files; and
You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and
If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License. 

You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.

So while it isn't GPL, would it be true to say that we could modify our distribution statement and be OK?

julian1: I found this, discussing apache and GPL3 http://www.apache.org/licenses/GPL-compatibility.html . I don't see there's a problem.

danfruehauf: My recommendation would be to:
 * Stuff those 2 functions in the same file
 * Have an apache 2.0 license header (optional)
 * State that this was deliberately taken from spring

anguss00: Furthermore, according to this document http://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses the Apache License 2.0 and the GNU GPL are compatible given the clause mentioned above

danfruehauf: @julian1 There's definitely not a problem to include it, but you cannot re-license it, that's all.

anguss00: Assuming of course that we use GPL version 3.0

julian1: We're already GPL3,

https://github.com/aodn/aodn-portal/blob/master/COPYING

anguss00: No worries @danfruehauf, I'll make those changes as required

Issue #1744, opened 2015-04-30
Step 2 number filter gives javascript error


dnahodil:
> App version: 4.2.1
> Build date: Fri Apr 17 16:24:33 AEST 2015
> Build number: 4
> Git ref: fee6a717030ec18a707839d7a160d5a9798e1315

**Steps to reproduce**
- Add a collection with a number filter to the map (e.g. [Platform: float], NOAA Drifters)
- On step 2 enter a number into the text field without having first chosen an operator

**What does happen**
- Javascript error in console

> TypeError: record is undefined
> http://portal-systest.aodn.org.au/imos123/js/portal-all.js?v=4.3.4
> Line 6401

**What should happen**
- No error in console
- Filter should not be applied (as it is invalid without an operator)

smancini:
Testing

PR #1762, opened 2015-05-15
Validate operator before attempting to update number filter

anguss00:
Fixes issue #1744

### DIFF
--- a/web-app/js/portal/filter/ui/NumberFilterPanel.js
+++ b/web-app/js/portal/filter/ui/NumberFilterPanel.js
@@ -11,6 +11,7 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
 
     OPERATOR_CLEAR: 'CLR',
     OPERATOR_BETWEEN: 'BTWN',
+    OPERATOR_EMPTY: '',
 
     constructor: function(cfg) {
 
@@ -33,6 +34,7 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
             triggerAction: 'all',
             mode: 'local',
             emptyText : OpenLayers.i18n("pleasePickCondensed"),
+            blankText: OpenLayers.i18n("pleasePickNumberOperator"),
             width: 165,
             editable: false,
             fieldLabel: "Value",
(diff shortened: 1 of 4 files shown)


dnahodil:
I'm just trying this out and it looks good. However, I'm just thinking it might be good to set the drop-down menu to be invalid to show the user that they should change that so their filter is applied. What do you think about that? I know this wasn't listed in 'what should happen'.

anguss00:
Yes, that's definitely a good point. I was thinking about doing that actually before I pushed this up... Can definitely implement if you think it's worthwhile @dnahodil

anguss00:
So have a red line around the numberfield while the operator dropdown is not set?

dnahodil:
Yeah, let's do it. I would make the dropdown invalid, actually. The number is fine in this case, the problem is that the drop-down is not set.

dnahodil:
Hmm. We don't actually do the opposite. If you select and operator but haven't yet enterred a number we don't highlight the number field. We should probably do that, too. :)

anguss00:
Good point, I'm on it

pmbohm:
@anguss00 Ive added what I think is good validation to the NumberFilterPanel, and now i read your discussion with @dnahodil. I wont look too close at what was said in case it conflicts with my changes :)

anguss00:
Haha thanks @pmbohm I'll have a look now

pmbohm:
@anguss00 the updating of the map when the user selects none is now done. Let me know if this is good and then i will squash yours an my work to one. The differences are large and confusing to leave two commits i think?

web-app/js/portal/filter/ui/NumberFilterPanel.js
```diff
@@ -115,10 +154,8 @@ Portal.filter.ui.NumberFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel
         });
 
         this._fireAddEvent();
+        trackFiltersUsage('filtersTrackingNumberAction', this._getTrackUsageLabel(), this.layer.name);
```
anguss00: Only issue is here; if I make an operator selection and then add a value in the first field, trackUsage is called twice (once for the operator and one for the numberField selection).

anguss00:
Looks good aside from one comment @pmbohm

pmbohm:
@anguss00 trackchanges hole now plugged. Commits squashed

anguss00:
Me gusta

PR #1765, opened 2015-05-18
Fix licensing for Spring framework overrides

anguss00:
Remove the need for spock assert class importing

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -32,14 +32,13 @@ class LayerController {
 
             try {
                 def con = new URL(_getMetadataUrl(params.uuid)).openConnection()
-                def reader = new InputStreamReader(con.content)
                 def metadataText = con.content.text
 
                 if (_isXmlContent(con.contentType)) {
 
                     def xml = new XmlSlurper().parseText(metadataText)
                     //TODO: Validate schema before proceeding
-                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), reader.getEncoding())
+                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), _contentTypeCharset(con.contentType))
 
                     response = abstractText
                     status = HTTP_200_OK
@@ -116,5 +115,10 @@ class LayerController {
 
     def _isXmlContent(contentType) {
         return contentType.find(/(text|application)\/xml/)
+    }
+
+    def _contentTypeCharset(contentType) {
+        def stringComponents = contentType.split("=")
+        return stringComponents[1]
     }
 }
--- a/grails-app/controllers/au/org/emii/portal/LayerController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/LayerController.groovy
@@ -38,7 +38,7 @@ class LayerController {
 
                     def xml = new XmlSlurper().parseText(metadataText)
                     //TODO: Validate schema before proceeding
-                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), _contentTypeCharset(con.contentType))
+                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), _extractCharsetType(con.contentType))
 
                     response = abstractText
                     status = HTTP_200_OK
@@ -117,8 +117,13 @@ class LayerController {
         return contentType.find(/(text|application)\/xml/)
     }
 
-    def _contentTypeCharset(contentType) {
-        def stringComponents = contentType.split("=")
-        return stringComponents[1]
+    def _extractCharsetType(contentType) {
+        def returnedCharset
+        contentType.split(';').each { token ->
+            if (token.contains('charset=')) {
+                returnedCharset = token.split('=')[1]
+            }
+        }
+        return returnedCharset
     }
 }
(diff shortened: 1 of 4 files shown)


grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
```diff
@@ -1,7 +1,10 @@
 /*
  * Copyright 2015 IMOS
  *
- * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ * The AODN/IMOS Portal is licensed under the Apache License, Version 2.0 (the "License")
```
danfruehauf: The portal is licensed under GPLv3, only this file (and the other) are under the Apache license.

anguss00: Ok, I'll fix

anguss00: So we still need to say that the Portal is distributed under GPL here? And then just say that this file is specifically licensed under apache?

danfruehauf: I think say the portal is under GPLv3, but this file is under the Apache license + add header to Apache license.

danfruehauf: By all means, I'm not a lawyer. It's just that if it was my code here in the file, this is what I would have expected.

anguss00:
changes made @danfruehauf

danfruehauf:
That's good.

anguss00:
:+1: thanks!

danfruehauf:
I gonna test it first in the VM...

anguss00:
@danfruehauf I have a solution, but I'm not sure how good it is...

grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
```diff
@@ -19,7 +29,7 @@ final class HtmlCharacterEntityReferences {
      * @since 4.1.2
      */
     public String convertToReference(char character, String encoding) {
-        if (encoding.startsWith("UTF")){
+        if (encoding.startsWith("UTF-")){
```
danfruehauf: Why this change?

grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -117,4 +116,9 @@ class LayerController {
     def _isXmlContent(contentType) {
         return contentType.find(/(text|application)\/xml/)
     }
+
+    def _contentTypeCharset(contentType) {
+        def stringComponents = contentType.split("=")
```
danfruehauf: This is a tad bit too naive.

grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -117,4 +116,9 @@ class LayerController {
...
+        return stringComponents[1]
```
danfruehauf: And this will return you a bunch of stuff you might not want. So generally speaking you should parse it properly or not parse it at all.

In the case of `Content-Type: application/xml;charset=UTF-8`, this will actually work fine, however think about
`Content-Type: application/xml;charset=UTF-8;value=booga`. Then in your case you gonna return `UTF-8;value` which is undesirable.

grails-app/utils/au/org/emii/portal/HtmlCharacterEntityReferences.groovy
anguss00: The return type from getEncoding() from an input stream was UTF8, while the value contained in the header is "UTF-8" (which is a better match for the internal groovy definition of the charset). I've taken the opportunity to tighten this conditional

anguss00:
Changes made @danfruehauf... not sure why the commit name is still fixup after I used ``git rebase -i HEAD~2 `` and modified the commit label....

anguss00:
Got rid of it, read the instructions @anguss :-p

grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -117,4 +116,14 @@ class LayerController {
     def _isXmlContent(contentType) {
         return contentType.find(/(text|application)\/xml/)
     }
+
+    def _extractCharsetType(contentType) {
+        def returnedCharset
+        contentType.split(';').each { token ->
+            if (token.contains('charset')) {
```
danfruehauf: Go with:
```
if (token.contains('charset=')) {
```

Be as strict as possible here, especially when we're doing "non-standard" parsing.

danfruehauf:
It generally looks good. We'll just have to test it.

anguss00:
Made that change @danfruehauf

danfruehauf:
Let me test it quickly. I should have everything in place...

danfruehauf:
Still doesn't work :(

grails-app/controllers/au/org/emii/portal/LayerController.groovy
```diff
@@ -32,14 +32,13 @@ class LayerController {
...
                 def con = new URL(_getMetadataUrl(params.uuid)).openConnection()
-                def reader = new InputStreamReader(con.content)
                 def metadataText = con.content.text
 
                 if (_isXmlContent(con.contentType)) {
 
                     def xml = new XmlSlurper().parseText(metadataText)
                     //TODO: Validate schema before proceeding
-                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), reader.getEncoding())
+                    def abstractText = HtmlUtils.htmlEscape(xml.identificationInfo.MD_DataIdentification.abstract.CharacterString.text(), _extractCharsetType(con.contentType))
```
danfruehauf: If I hard code the second parameter here to `UTF-8` and have `LC_ALL=C` - still breaks. Not sure what's the deal...

anguss00: Wow... that's pretty depressing :-( not sure how we progress here...

danfruehauf: I'm not sure what's the deal. We'll have a look tomorrow perhaps?

anguss00: Sounds like a good plan, I'll mull it over in the meantime

danfruehauf:
See #1767

PR #1767, opened 2015-05-20
Issue 1748 no encoding fix

### LINKED ISSUE #1748: Metatdata information not displayed correctly

danfruehauf:
Finally fixes #1748 

Sometimes less is more. Using this will ensure the correct encoding when using XmlSlurper:
```
def xml = new XmlSlurper().parse(_getMetadataUrl(params.uuid))
```

Tested with `LC_ALL=C` and works.


danfruehauf:
Now it's sorted, GTG.

PR #1768, opened 2015-05-20
Fixes #1757

### LINKED ISSUE #1757: Removing data collections quickly that have been added via ctrl click results in double select button
App version: 4.4.1
            Build date: Fri May 08 17:06:37 AEST 2015
            Build number: 12
            Git ref: eb55aa288ba068b736d1c7a59d2fe61fa71f8c0e

This is also in production

Steps to repeat:

Add 5-6 random collections to the map using ctrl click function (hold control key down while clicking to stay on step 1)

Proceed to step 2 here align mouse pointer over the "remove" cross on the collection at the top of the list.

Quickly and continuously click until all collections have been removed, 

Go back to step 1

What should happen:

All results only have 1 select button

What does happen:

Collections have 2 select buttons sometimes even 3

pmbohm:


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -205,6 +205,9 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
 
         var buttonElementId = this.buttonElementId(uuid);
 
+        // remove any existing content fixes #1757
```
danfruehauf: Capitalize comment ;)

web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -248,6 +251,9 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
 
         values.mapContainerId = this.mapElementId(values.uuid);
 
+        // remove any existing content fixes #1757
```
danfruehauf: Here too.

web-app/js/portal/common/helpers.js
```diff
@@ -316,3 +316,8 @@ function getPortalBase(pathname) {
 function normaliseLongitude(longitude) {
     return (longitude + 540) % 360 - 180;
 };
+
+// Remove all child nodes of HTML elements from the DOM
+function clearContents(jQuerySelector) {
```
danfruehauf: Since this is used only in `FacetedSearchResultsDataView.js` and is not necessarily very reusable - I'd stick it in `FacetedSearchResultsDataView.js`.

danfruehauf:
Fix is good, works as expected.

web-app/js/portal/common/helpers.js
pmbohm: It is so usable, I decided to stick it in helpers.js. There are so many convoluted ways to do this in Ext land, that are (or used to be) in the code, but because they are convoluted, they are hard to find. I think this is a great method to have.

web-app/js/portal/search/FacetedSearchResultsDataView.js
pmbohm: now thats just silly @danfruehauf

danfruehauf: It's breaking a bit the coding convention but OK. I can live with it.

PR #1778, opened 2015-06-02
Chevrons removed in favour of true double arrows

pmbohm:


anguss00:
We have merge conflicts here @pmbohm after merging the big Kahuna. But I think we can easily fix them

PR #1783, opened 2015-06-04
Sort filters properly before adding them to panels on step 2

anguss00:


web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -147,34 +148,32 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Ext.Container, {
...
-        var typeOrder = function (panel) {
-            return panelOrder.indexOf(panel.constructor) * -1;
+        var typeOrder = function (filter) {
+            return panelOrder.indexOf(filter.constructor) * -1;
         };
 
         var _this = this;
-        panels.sort(function(firstPanel, secondPanel) {
-            var comparisonResult = _this._compareElements(typeOrder(firstPanel), typeOrder(secondPanel));
+        filters.sort(function(firstFilter, secondFilter) {
```
danfruehauf: I'd like the logic of this function to become something simpler, perhaps also not in a closure.

It should be something along the lines of:
```
compareFilterOrder: function(lhs, rhs) {
    return compareType(lhs, rhs) || compareLabel(lhs, rhs);
},

compareType: function(lhs, rhs) {
    if (lhs.type == rhs.type) {
        return false;
    }
    else {
        return lhs > rhs ? lhs : rhs;
    }
},

compareLabel: function(lhs, rhs) {
    return lhs > rhs ? lhs : rhs;
}
```

I'm not a big fan of the `-1` multiplier. It's very vague.

anguss00:
Change made @danfruehauf

PR #1786, opened 2015-06-05
Accordion collection ordering

pmbohm:


jonescc:
Looking at this

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -39,10 +39,16 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
 
         this._disableDateTimeFields();
         this._attachTemporalEvents();
+
+        if (this.layer.temporalextentloaded) {
```
jonescc: Would it be better/possible to store this ui state locally rather than on the layer

jonescc: Or, if it is information about the layer and loading - perhaps this code should be moved to the layer or to wherever the loading occurs.  It just seems odd updating one from the other here.

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -323,7 +329,8 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
         if (this.geoNetworkRecord) {
             this._addDateTimeFilterToLayer();
             this.geoNetworkRecord.updateNcwmsParams(dateRangeStart, dateRangeEnd, geometry);
-            Ext.MsgBus.publish(PORTAL_EVENTS.DATA_COLLECTION_MODIFIED);
+            // todo cant have this for accordion
+            //Ext.MsgBus.publish(PORTAL_EVENTS.DATA_COLLECTION_MODIFIED);
```
jonescc: Perhaps removing this comment would be better.  It won't mean much to anyone after this change is merged.

jonescc:
@pmbohm, some comments above.  But more importantly after using the move button, the ncwms layers don't work for me any more

PR #1795, opened 2015-06-15
Re-ordering accordion. Done without recreating items

pmbohm:


dnahodil:
I'll review this. However, it looks like it requires a rebase, @pmbohm.

web-app/css/imosTheme.css
```diff
@@ -337,5 +337,14 @@ ul.x-tab-strip-top {
...
 
+.subsetPanelAccordionItem .x-tool:not(:first-of-type) {
+    margin: 3px 0;
+}
+
+.subsetPanelAccordionItem .x-tool:first-of-type {
+    margin-left: 6px;
+    margin-right: 0;
+}
+
```
dnahodil: Are these changes really IMOS-theme specific? or should they go in the general.css file?

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -323,7 +323,6 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
         if (this.geoNetworkRecord) {
             this._addDateTimeFilterToLayer();
             this.geoNetworkRecord.updateNcwmsParams(dateRangeStart, dateRangeEnd, geometry);
-            Ext.MsgBus.publish(PORTAL_EVENTS.DATA_COLLECTION_MODIFIED);
```
dnahodil: Why don't we send the modded event when filters are changes anymore?

web-app/js/portal/details/SubsetItemsWrapperPanel.js
```diff
@@ -0,0 +1,87 @@
...
+
+        var tabPanelForLayer = new Portal.details.SubsetItemsTabPanel({
+            map: cfg.map,
+            layer: cfg.layer,
+            listeners: {
+                beforeTabChange: this._doTracking
+            }
+        });
+
+        this.layer = cfg.layer;
```
dnahodil: This line shouldn't be needed. All values from config should be automatically applied to 'this'.

dnahodil:
It also seems to introduce 2 bugs:
1) If you ctrl-click add a few layers quickly then their subsetting options can fail to load
2) Rearranging the layers in the accordion does not change the order they appear on the map

web-app/css/imosTheme.css
pmbohm: mmm... My thoughts are that themeing is useless as everyone starts with the imosTheme anyway.
This change affect prettiness only and as such why not keep it for IMOS

web-app/js/portal/details/SubsetItemsWrapperPanel.js
pmbohm: Absolutely!

web-app/css/imosTheme.css
dnahodil: Okay, cool. I didn't look into how these changes affect things, but if they're not required for the thing to work then they're fine in here as you say.

web-app/js/portal/details/NcWmsPanel.js
pmbohm: This is now a while ago. It was the cause of an infinite loop in my changes. As to why it was needed in the first place ... no idea.
Basically now the ActiveGeoNetworkRecordStore triggers this event
the DownloadPanel and SubbsettingPanel listen
So the event is more like AN_ITEM_IN_THE_DATA_COLLECTION_STORE_IS_MODIFIED_OR_JUST_THE_ORDER

pmbohm:
@dnahodil bugs now fixed

dnahodil:
Changes look good @pmbohm. Nice one.

PR #1797, opened 2015-06-15
Issue #1770

### LINKED ISSUE #1770: Single click loses manually typed bounding box co-ordinates
App version: 4.5.0
            Build date: Fri May 22 14:04:54 AEST 2015
            Build number: 13
            Git ref: 72b9e6b8cbedd3738bc5e2bc21d4df04c529e2e0
        -->

Think this may be something to do with new openlayers.

Steps to reproduce:

Add any collection to be taken to step 2, manually type bounding box co-ordinates into the fields eg N -13, W 137, E 155, S -27 hover anywhere on map and click a single point

What does happen:

Click point shows on map as red (invalid option) then disappears and previously drawn bounding box and co-ordinates are reset.

What should happen:

Functionality added to the portal prevented a single click point from occurring as users complained about not being able to see it and then not understanding why they got no data to download so the click was invalid and affected nothing.

dnahodil:
Fixes #1770

### DIFF
--- a/src/test/javascript/portal/ui/openlayers/control/SpatialConstraintSpec.js
+++ b/src/test/javascript/portal/ui/openlayers/control/SpatialConstraintSpec.js
@@ -362,6 +362,7 @@ describe('Portal.ui.openlayers.control.SpatialConstraint', function() {
     describe('_showSpatialExtentError', function() {
 
         var testLayer = {};
+        var testGeometry;
 
         beforeEach(function() {
             spyOn(spatialConstraint, 'addAntimeridian');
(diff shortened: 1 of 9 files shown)


src/test/javascript/portal/ui/openlayers/control/SpatialConstraintSpec.js
```diff
@@ -309,22 +308,18 @@ describe('Portal.ui.openlayers.control.SpatialConstraint', function() {
             expect(spatialConstraint._checkSketch(feature)).toEqual(true);
         });
 
-        it('returns false if geometry is too small, shows error', function() {
```
jkburges: image

web-app/js/portal/ui/openlayers/control/SpatialConstraint.js
```diff
@@ -13,7 +13,7 @@ Portal.ui.openlayers.SpatialConstraintType = {
```
jkburges: This class is quite big, and complex, IMO.  What are its main responsibilities currently (I could answer that but since you've just looked at the code, perhaps you're in a better position to do so)?

src/test/javascript/portal/ui/openlayers/control/SpatialConstraintSpec.js
dnahodil: The _checkSketch() method previously checked the validity of the geometry *and* called the _showAnteMeridianError(). Now it's only job is to return the validity.

But I now realise I didn't write tests to check my new behaviours so I will do that.

web-app/js/portal/ui/openlayers/control/SpatialConstraint.js
dnahodil: Good question.

It represents the control (a non-visible object) for the geometry you can draw on the map.
There is code to:
* Attach / remove it from the map
* Initialise its events
* Get the constraint / geometry that has been drawn
* Arrange the layer showing the geometry on the map
* Validating the geometry that is drawn
* Visualising an error if it occurs

So it does do a lot of things. I will have a think which of those responsibilities could be nicely extracted.

dnahodil:
@jkburges review changes have been made. I've added some more tests for the functionality I added and I extracted the logic for validating the drawn geometry. There are probably other candidates for new classes to be extracted from the SpatialConstraint but I think we should leave them for another day.

jkburges:
:+1: it's a step in the right direction.

PR #1813, opened 2015-06-26
Add spatial constraint validator to geometry facet

### LINKED ISSUE #1811: Step 1 geographic facet is broken
Using the geographic facet on step 1 gives a JavaScript error

> TypeError: this.validator is undefined
	return this.validator.isValid(geometry);

dnahodil:
Fixes #1811

### DIFF
--- a/web-app/js/portal/search/FacetMapPanel.js
+++ b/web-app/js/portal/search/FacetMapPanel.js
@@ -23,6 +23,7 @@ Portal.search.FacetMapPanel = Ext.extend(Portal.common.MapPanel, {
 
         this._initGeoFacetMapToolbar();
         this._initMap();
+        this._initSpatialConstraintValidator();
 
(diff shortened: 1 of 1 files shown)


web-app/js/portal/search/FacetMapPanel.js
```diff
@@ -57,6 +58,12 @@ Portal.search.FacetMapPanel = Ext.extend(Portal.common.MapPanel, {
         });
     },
 
+    _initSpatialConstraintValidator: function() {
+        this.geoFacetMapToolbar.spatialConstraintControl.validator = new Portal.filter.validation.SpatialConstraintValidator({
```
dnahodil: It is not nice that we have to go into the object and directly set the validator. For the main map we set the validator with dependency injection, when the spatial constraint is created, which is much better. However, there is a confusion of maps here (2 different maps are referenced by `this.map` during the constructor) and I couldn't find how to untangle that.

PR #1814, opened 2015-06-29
I, for one, welcome our new web crawling overlords

jkburges:


danfruehauf:
No `robots.txt`? Means it'll return a 404 for that. Is that desirable?

jkburges:
> No robots.txt? Means it'll return a 404 for that. Is that desirable?

Without one, crawlers will be free to crawl the entire website - which is ok for us (now).  The only reason we might want one (that I can think of) is to point crawlers to an xml `sitemap` - but do we have one?

danfruehauf:
We actually do!

It's in sitemap.xml (https://imos.aodn.org.au/sitemap.xml)

jkburges:
Ok let me change this then!

jkburges:
Come to think of it, I think I helped work on that :-)

jkburges:
Closing in favour of https://github.com/aodn/aodn-portal/pull/1815

PR #1815, opened 2015-06-29
I, for one, welcome our new web crawling overlords

jkburges:


grails-app/views/robots.gsp
```diff
@@ -7,5 +6,7 @@
 
 --%>
 
-<%@ page contentType="text/plain;charset=UTF-8" %>User-agent: *
-Disallow: /
+<%@ page contentType="text/plain;charset=UTF-8" %>
+User-agent: *
+Allow: /
+Sitemap: https://imos.aodn.org.au/sitemap.xml
```
danfruehauf: Isn't that supposed to be a relative URL?

jkburges: Not according to google: https://developers.google.com/webmasters/control-crawl-index/docs/robots_txt#sitemap

Interestingly, it's possible to have multiple sitemap entries, so perhaps we should consider cross-linking to geonetwork (outside scope of this PR, though)?

jkburges: > perhaps we should consider cross-linking to geonetwork

Not that there's one there, however.

PR #1822, opened 2015-07-02
fixes #1821 - layout manager internally properly updated when collections are deleted

### LINKED ISSUE #1821: Step 2 accordion - Sometimes newly-added collection is collapsed
Spotted today on edge
```
App version: 4.7.0
Build number: 31
```

**Steps to reproduce**
* Return to step 1
* Return to step 1
* Add another collection

**What does happen**

**What should happen**

pmbohm:
Surplus expands and circular listeners removed

### DIFF
--- a/web-app/js/portal/details/SubsettingPanel.js
+++ b/web-app/js/portal/details/SubsettingPanel.js
@@ -67,29 +67,17 @@ Portal.details.SubsettingPanel = Ext.extend(Ext.Panel, {
         var layerContainer = new Portal.details.SubsetItemsWrapperPanel({
             map: this.map,
             layer: layer,
-            layerItemId: this._getItemIdForLayer(layer),
-            listeners: {
-                expand: this._fireSelectedLayerChangedEvent(layer),
-                scope: this
-            }
+            layerItemId: this._getItemIdForLayer(layer)
         });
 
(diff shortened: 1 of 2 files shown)


pmbohm:
This also fixes the case where an accordion panel was deleted that was the currently expanded item and the no accordion panel is opened to replace it at the currently selected item

danfruehauf:
Are there tests for this code? I think a test like "always expand last added collection" can be a good one.

It fixed the bug BTW, I verified it.

pmbohm:
I dont see how i can test this. @danfruehauf

danfruehauf:
You can test that `this.currentlyExpandedPanel` will be set to null when `onDestroy` is called, and verify the opposite for the other case. What do you think?

pmbohm:
Test Ext.ux.NonCollapsingAccordionLayout directly?
That involves rendering items?

danfruehauf:
You can mock whatever is going to emit errors because of rendering I believe.

pmbohm:
I would need to actually render, as the logic in Ext.ux.NonCollapsingAccordionLayout kicks off when an item is added via 'renderItem'.

Anyway "What do I think?". No I dont see myself writing a jasmine test for this

danfruehauf:
Alright.

PR #1824, opened 2015-07-03
Revert "fixes #1821 - layout manager internally properly updated when…

### LINKED ISSUE #1821: Step 2 accordion - Sometimes newly-added collection is collapsed

dnahodil:
… collections are deleted"

Had to revert this as it creates a bug which hides some behaviour important for the review on Monday.

This reverts commit 48bebb4980339666e7d2bffcaa5b5c777522727d.


danfruehauf:
I'll merge in a sec.

PR #1825, opened 2015-07-06
fixes #1821 - without creating new bugs!

### LINKED ISSUE #1821: Step 2 accordion - Sometimes newly-added collection is collapsed

pmbohm:
Now publishes 'PORTAL_EVENTS.SELECTED_LAYER_CHANGED' when an accordion item is expanded, therefore the map is notified


dnahodil:
That fixes the bug. Thanks @pmbohm

PR #1826, opened 2015-07-06
Step1 button enhancements

pmbohm:


dnahodil:
Is there any way we can extract that font-awesome html fragment so that we only need to define it once or twice instead of having it spread in a dozen different places?

pmbohm:
Yes ok where possible (non gsp places) I'll do it. Maybe a super css class of our own that does what 'fa fa-fw fa-external is doing is cleaner simpler? @dnahodil

pmbohm:
Using the 'external' css style now will get the same Bootstrap Fontawesome effect as appending a span with "fa fa-fw fa-external-link' classes. The css class method is more simple to apply using gsp's or javascript

dnahodil:
This appears to add unwanted spacing to the right hand side of links which are within a paragraph of text (see the landing page).

pmbohm:
@dnahodil Disgusting unwanted spacing removed :)
I did look earlier at the landing page, just not when it really mattered.

PR #1827, opened 2015-07-07
Mark links as external

dnahodil:
@pmbohm I noticed that these links (landing page footer and general footer links) were not marked as external. Is it worth bringing them into line?


pmbohm:
Its uglier? 
But its more correct :(

PR #1840, opened 2015-07-15
Ext js 4 sandbox

jkburges:
GLHF


danfruehauf:
And the objective now is to get the following command to output `0`:
```
$ grep -r "Ext\." src/test/javascript/* web-app/js/portal/* | wc -l
```

jkburges:
> And the objective now is to get the following command to output 0

Correct, at which point we *should* be able to do:
* a global search and replace of `Ext4` to `Ext`
* `Ext.createWrapped` -> `Ext.create`
* include 'proper' Ext JS 4 files, instead of sandbox ones.

danfruehauf:
This is GTG once we get some smoke testing on it. It looks fine to me...

PR #1841, opened 2015-07-15
fixes #1823

### LINKED ISSUE #1823: Accordion - multiple items open at same time
**Steps to reproduce**
Quickly multi select 3 or 4 layers on step one - going to step 2 and then going back to step 1 multi-selecting a few more layers and going to step 2 again.  Sometimes two or more accordion panels will then be expanded.

**What does happen**
You can end up with multiple items open in the accordion

**What should happen**
Only the last-added item should be open

pmbohm:


web-app/js/portal/prototypes/NonCollapsingAccordionLayout.js
```diff
@@ -30,7 +29,16 @@ Ext.ux.NonCollapsingAccordionLayout = Ext.extend( Ext.layout.Accordion, {
...
-        panelToCollapse.collapse();                         // Collapse the previously selected panel
+
+        if (panelToCollapse.collapsed) {
+            // fixes #1823
+            setTimeout(function() {
+                panelToCollapse.collapse();
+            }, 400 );
+        }
+        else {
+            panelToCollapse.collapse();                     // Collapse the previously selected panel
```
dnahodil: I would say this comment is pretty redundant and that it can be removed. If you disagree (which you might) I think we should at least get rid of the spaces and also make a similar comment on line 36 (as it is essentially the same code, uncommented).

pmbohm: @dnahodil Those comments are a hangover from the original author. I wont mind if all of the comments disappear, or are in line with our code base, as we now have made big changes to that original code. maybe the "fixes #1823" is still useful?

dnahodil: No problems, @pmbohm. You're right about this code having changed a lot from the original, now. I might take this opportunity to have it match our code styles, too.

I'll get one of the other devs to finish this PR with me.

PR #1844, opened 2015-07-16
Removed some vars from window scope

pmbohm:


danfruehauf:
Did you verify it works as expected?

pmbohm:
@danfruehauf 
the home/ redirect still works.
the 'drilldownFilterIndex' only occurs in the 'hasDrilldown' function in the whole project.

grails-app/views/home/index.gsp
```diff
@@ -28,11 +28,14 @@ The AODN/IMOS Portal is distributed under the terms of the GNU General Public Li
...
-                window.location.replace(homeWithoutTrailingSlash);
-            }
+
+            (function () {
+                // Fixes #1723, redirect 'home/' to 'home'
+                if (window.location.href.match(/\/home\/$/)) {
+                    var homeWithoutTrailingSlash = window.location.href.substring(0, window.location.href.length - 1);
+                    window.location.replace(homeWithoutTrailingSlash);
+                }
+            }());
```
dnahodil: @pmbohm Why this change?

pmbohm:
oh but chrome is not as happy with the home/ redirect!!!

grails-app/views/home/index.gsp
pmbohm: I was looking at fire bug and i saw a bunch of stuff that I thought was unnecessary to be in the global scope. 
'homeWithoutTrailingSlash' may be necessary!

dnahodil: Okay, so it was to change the scope of the variable, @pmbohm?

PR #1850, opened 2015-07-27
Revert "Ext js 4 sandbox"

danfruehauf:
Reverts aodn/aodn-portal#1840

We're not doing ext4, lets get rid of that then.


dnahodil:
Yeah, rollback

Issue #1839, opened 2015-07-15
Flash player warnings show when loading Portal


dnahodil:
Firefox have recently disabled Flash player by default and since then I have been seeing messages in the Portal showing that Flash player has been blocked. @lbesnard noticed this a little while ago, but we all decided he was probably just crazy.

It would be good to know why these flash messages are appearing (I can't think of where we would be using Flash) and if there's an alternative we can consider now that one of the major browsers is disabling Flash by default [citation needed].

**Steps to reproduce**
Load the Portal in FireFox with Flash player disabled

**What does happen**
Warnings about flash player being disabled

**What should happen**
No warnings
Warm fuzzy feelings inside as you easily find data relevant to your interests and download it in a format convenient to your needs.

danfruehauf:
https://addons.mozilla.org/en-US/firefox/blocked/p946

dnahodil:
See https://github.com/aodn/aodn-portal/blob/182b0ac883c8e8b926cee97a6be95f88ac0c1338/web-app/js/ext-3.3.1/ext-all-debug.js line 36573

lbesnard:
thanks @dnahodil , you're proving I'm not totally insane

PR #1852, opened 2015-07-28
disable flash checking in ext

danfruehauf:
fix for #1839. firefox has disabled flash, avoid the ugly popup
by not probing the flash version in the first place

sadly, there's no proper way to override this function as it is
called when the dom is loading


danfruehauf:
I understand if we are going to be hesitant before merging that, however:
 * I don't think it's worth spending heaps of time on that
 * Probably doesn't happen on ext4/5/6
 * I doubt there's a better way to fix it, but please surprise me if there is

dnahodil:
I don't know of a better way to fix this, either.

I assume the ext-all.js change is the same?

danfruehauf:
Yeah, `ext-all.js` fixed as well manually. I just took the parenthesis away so the function is not called (`swfobject`).

Please do check me though. Please verify this fix works also for you (I tested only on newest Firefox and made sure I don't get the flash artifact showing).

dnahodil:
Yep. I'll check it out.

dnahodil:
Looks good, @danfruehauf

PR #1853, opened 2015-07-28
Issue 1835

jkburges:
I have to admit, I'm not overly happy with 642a4ae, but it does seems to work ok for me.

### DIFF
--- a/web-app/js/portal/PortalEvents.js
+++ b/web-app/js/portal/PortalEvents.js
@@ -3,8 +3,6 @@ var PORTAL_EVENTS = {
     SELECTED_LAYER_CHANGED:        "selectedLayerChanged",
     BASE_LAYER_CHANGED:            "baseLayerChanged",
     LAYER_REMOVED:                 "layerRemoved",
-    LAYER_LOADING_START:           "layerLoadingStart",
-    LAYER_LOADING_END:             "layerLoadingEnd",
     BASE_LAYER_LOADED_FROM_SERVER: "baseLayersLoadedFromServer",
     DATA_COLLECTION_ADDED:         "dataCollectionAdded",
     DATA_COLLECTION_MODIFIED:      "dataCollectionModified",
--- a/web-app/js/portal/details/SubsetItemsWrapperPanel.js
+++ b/web-app/js/portal/details/SubsetItemsWrapperPanel.js
@@ -11,15 +11,21 @@ Portal.details.SubsetItemsWrapperPanel = Ext.extend(Ext.Panel, {
 
     constructor: function(cfg) {
 
-        var tabPanelForLayer = new Portal.details.SubsetItemsTabPanel({
-            map: cfg.map,
-            layer: cfg.layer,
-            listeners: {
-                beforeTabChange: this._doTracking
-            }
+        var tabPanelForLayer = this._initSubsetItemsTabPanel(cfg);
+
+        this.createTools(cfg.layer);
+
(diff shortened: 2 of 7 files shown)


danfruehauf:
On that branch, in chrome I get the following when trying to add a collection:
```
Uncaught TypeError: Cannot read property 'added' of undefined
```

web-app/js/portal/PortalEvents.js
```diff
@@ -3,8 +3,6 @@ var PORTAL_EVENTS = {
     SELECTED_LAYER_CHANGED:        "selectedLayerChanged",
     BASE_LAYER_CHANGED:            "baseLayerChanged",
     LAYER_REMOVED:                 "layerRemoved",
-    LAYER_LOADING_START:           "layerLoadingStart",
```
danfruehauf: While you're at it, also those can be removed:
 * `BEFORE_SELECTED_LAYER_CHANGED`
 * `LOAD_SNAPSHOT`

web-app/js/portal/ui/openlayers/MapOptions.js
```diff
@@ -50,6 +50,7 @@ Portal.ui.openlayers.MapOptions = Ext.extend(Object, {
             this.clickControl
         ];
 
+        this.tileManager = null;
```
danfruehauf: Why is this needed here?

web-app/js/portal/details/SubsetItemsWrapperPanel.js
```diff
@@ -74,6 +89,7 @@ Portal.details.SubsetItemsWrapperPanel = Ext.extend(Ext.Panel, {
         this.spinnerToolItem = {
             id: 'spinnerToolItem',
             styles: 'fa-spin fa-spinner',
+            hidden: !layer.loading,
```
danfruehauf: Would it be safer to have `layer.loading !== false`? Just not sure what behaviour you want when `layer.loading` is either `undefined` or `null`.

web-app/js/portal/PortalEvents.js
jkburges: I'll do this on another PR.

danfruehauf: Fair.

web-app/js/portal/ui/openlayers/MapOptions.js
jkburges: It "fixes" the problem that you mention here: https://github.com/aodn/aodn-portal/commit/63472c212a1c2cdfb4cffb6459fd7e4919cb91e2

Along with this, and having our override removed, panning + layer events works.  Like I say above, I'm not very happy with this, hopefully I haven't broken anything.

web-app/js/portal/details/SubsetItemsWrapperPanel.js
jkburges: If `undefined` or `null`, then the spinner will be hidden, which I think is correct.  I prefer this style (of "truthiness") but it's just a preference.

jkburges:
The `Cannot read property 'added' of undefined` error has now been fixed.  It was a rebase gone wrong, and me not testing properly afterwards :crying_cat_face:

web-app/js/portal/PortalEvents.js
jkburges: See: https://github.com/aodn/aodn-portal/pull/1856

PR #1854, opened 2015-07-28
Testing helper function

dnahodil:
I get really sick of writing anonymous functions in the Jasmine tests. Would people be happy using this helper? **I don't *think* this would be controversial but maybe we get the okay from 2 devs before we merge this?**

Essentially instead of writing: (Which happens heaps in test specs)
``` javascript
var testLayer = {
    isNcwms: function() { return true }
}
```
You can just write:
``` javascript
var testLayer = {
    isNcwms: returns(true)
}
```

Note, it can't be used if the variable being returned will change after the function is created. In that case you'd need to write the anonymous function the old-fashioned way. I think there were about 3 cases of this is the current test suites.

### DIFF
--- a/src/test/javascript/SpecHelperSpec.js
+++ b/src/test/javascript/SpecHelperSpec.js
@@ -67,4 +67,21 @@ describe('Custom matchers', function () {
             expect([]).not.toBeNonEmptyString();
         });
     });
+
+    describe('returns()', function() {
+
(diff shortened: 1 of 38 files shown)


danfruehauf:
Why close?

OK from me.

dnahodil:
Recreated as #1855 

Just a mistake, really

PR #1855, opened 2015-07-28
New helper function for testing

dnahodil:
I get really sick of writing anonymous functions in the Jasmine tests. Would people be happy using this helper? **I don't *think* this would be controversial but maybe we get the okay from 2 devs before we merge this?**

Essentially instead of writing: (Which happens heaps in test specs)
``` javascript
var testLayer = {
    isNcwms: function() { return true }
}
```
You can just write:
``` javascript
var testLayer = {
    isNcwms: returns(true)
}
```

Note, it can't be used if the variable being returned will change after the function is created. In that case you'd need to write the anonymous function the old-fashioned way. I think there were about 3 cases of this is the current test suites.

### DIFF
--- a/src/test/javascript/SpecHelperSpec.js
+++ b/src/test/javascript/SpecHelperSpec.js
@@ -67,4 +67,21 @@ describe('Custom matchers', function () {
             expect([]).not.toBeNonEmptyString();
         });
     });
+
+    describe('returns()', function() {
+
+        it('returns the value it was created with', function() {
+
+            var testObj = {};
+            var returnNumber = returns(64);
+            var returnBool = returns(false);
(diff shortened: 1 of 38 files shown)


danfruehauf:
As said before - OK from me.

dnahodil:
@jkburges any thoughts?

jkburges:
I think it's good :+1:

dnahodil:
Thanks guys. If someone could merge, then, please?

jkburges:
image

jkburges:
Sorry, I'm feeling a bit emotional w.r.t. to git after a bad rebase experience.

dnahodil:
:-1: we should stuck with SVN...

PR #1857, opened 2015-07-28
Stop the gloats

jkburges:
... from travis.


jkburges:
This doesn't actually *stop* the gloats, it just stops myself and @dnahodil getting every travis notification and no-one else getting any.

Now, the default travis notification behaviour will apply: http://docs.travis-ci.com/user/notifications/#Email-notifications

PR #1858, opened 2015-07-28
Disable flash detection in ext-deb-all-w-comments.js

dnahodil:
@danfruehauf apparently Jasmine uses this file for Ext so I've disabled the flash detection here as well.


danfruehauf:
Legit, I'll merge.

dnahodil:
I didn't think to check Jasmine on the other PR.

PR #1865, opened 2015-08-04
Style panel refactor

jkburges:
The original `StylePanel` had a couple of distinct categories of components.  This refactor makes that explicit.

It'll make some of the multi WMS stuff easier to implement.


web-app/js/portal/details/StylePanel.js
```diff
@@ -8,15 +8,10 @@
...
-
     constructor: function(cfg) {
         this.layer = cfg.layer;
 
-        var config = Ext.apply({
-            title: OpenLayers.i18n('stylePanelTitle')
-        }, cfg);
-
-        Portal.details.StylePanel.superclass.constructor.call(this, config);
+        Portal.details.LayerDetailsPanel.superclass.constructor.call(this, cfg);
```
danfruehauf: I think this is not calling the right constructor, you probably want:
```
Portal.details.StylePanel.superclass.constructor.call(this, cfg);
```

Or am I missing something?

src/test/javascript/portal/details/LayerDetailsPanelSpec.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2012 IMOS
```
danfruehauf: 2015 :)

src/test/javascript/portal/details/LayerDetailsPanelSpec.js
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ *
+ */
+
+describe("Portal.details.LayerDetailsPanel", function() {
```
danfruehauf: Are there actually any tests in that file?

web-app/js/portal/details/LayerControlPanel.js
```diff
@@ -0,0 +1,77 @@
...
+            }
+        });
+
+        this.items = [
+            this.opacitySliderContainer,
+            this.layerVisibilityCheckbox,
+            this.zoomToLayer,
+        ];
+
+        Portal.details.LayerDetailsPanel.superclass.initComponent.call(this);
```
danfruehauf: I think here you should call:
```
Portal.details.LayerControlPanel.superclass.initComponent.call(this);
```

src/test/javascript/portal/details/LayerDetailsPanelSpec.js
jkburges: Whatevs, you know how I feel about these notices :-)

danfruehauf: I know, I know...

web-app/js/portal/details/StylePanel.js
jkburges: Yeah that's a mistake.  It worked (with this errors) because the superclass is Ext.Panel in both cases.

web-app/js/portal/details/LayerControlPanel.js
```diff
@@ -0,0 +1,77 @@
...
+            value: true,
+            boxLabel: OpenLayers.i18n('showMapLayer'),
+            checked: true,
+            listeners: {
+                scope: this,
+                check: this._visibilityButtonChecked
+            }
+        });
+
+        this.zoomToLayer = new Ext.ux.Hyperlink({
```
danfruehauf: Can we make this a simple button?

I'm saying that because we should stop using that `Ext.ux.Hyperlink` thing wherever we can.

src/test/javascript/portal/details/LayerDetailsPanelSpec.js
jkburges: No, I left it just in case but on second thoughts I'll get rid of it.

danfruehauf: I think it's OK to have the file + the `describe` thing, but no need then for the `beforeEach` block.

web-app/js/portal/details/LayerControlPanel.js
jkburges: Um possibly but I don't want to make that change in this PR.

danfruehauf: Fair enough. It is code that was moved. Don't bother, I might just bulk change them all at some point.

jkburges: I tried to keep code changes to a minimum, apart from moving things around of course.  Although I was able to get ride of a weird bit of code that didn't seem necessary anymore: https://github.com/aodn/aodn-portal/pull/1865/files#diff-8da76661a875615270e9fb526f5efac0L149

jkburges:
@danfruehauf GTG again I think.  I haven't changed the copyright header - bulk commit coming soon on that I think.

web-app/js/portal/details/LayerControlPanel.js
dnahodil: > I'm saying that because we should stop using that Ext.ux.Hyperlink thing wherever we can.

Why do you say that @danfruehauf ?

danfruehauf: While trying to upgrade to ext4, I noticed `Ext.ux.Hyperlink` can be factored into a simple button. I'm not entirely sure what that class gives us on top of what a simple button does. I might be wrong though as I haven't looked /that/ deep into that.

dnahodil: Ok. I think the reason we included it was that we wanted UI elements which were hyperlinks. If with simple styling we can create buttons which look and work like hyperlinks then I'd say we could get rid of this.

danfruehauf: Then I'm not sure about it. However also @pmbohm said something about it which I didn't quite understand at the time, but said we probably should get rid of those too.

We're not moving to Ext4 any time soon, so I guess it's fine anyway. However it'd be good to start removing out dependence on `Ext.ux` because it locks us down a bit tighter to specific Ext versions, making the upgrade more difficult if it ever takes place.

PR #1866, opened 2015-08-04
Remove copyright

jkburges:
I got tired of adding/editing this totally redundant information in every source file.  I *believe* it is there for historical reasons only, but it is not required anymore (??).

We have https://github.com/aodn/aodn-portal/blob/master/COPYING in the root of the project, which clearly states what license this software is release under.  With git, as opposed to subversion, one has to go out of one's way to clone *part* of a project, so it's pretty certain that people are going to end up with the above file in their copy of the source code.

Some more concrete arguments aside from "I don't like it":

* I use lines added/removed as an *indicator* metric when reviewing code changes, e.g. does this change add and/or remove a lot of code.  When many small, targerted classes are added, then the copyright header distorts this metric.
* it's a barrier for people wanting to refactor code in to "single responsibility" classes

I think @pblain and @jachope should give this the ok before merging, in case there is *still* some reason we need to have these headers.


dnahodil:
Good one, it'll be nice to see these go

jkburges:
@pblain @jachope bump.

pblain:
I'm fine with it.  See: https://help.github.com/articles/open-source-licensing/

The important section reads as follows ...

Most people place their license text in a file named LICENSE.txt (or LICENSE.md) in the root of the repository; 
...
As long as your intention is clear, you can place your license information wherever you like. Some projects only include a descriptive line in the project's README (e.g., "This project is licensed under the terms of the MIT license."). However, to avoid any potential ambiguity, it's best to include the license file with your project as well.

@jachope, what do you think?

danfruehauf:
Bump!

Need rebase too.

jachope:
I am happy with what Pete suggests - my only concern was regarding if it was possible to get code and use it without seeing the license text and therefore not adding the same license to the code once they take it.  For those of you who were about this happy to some of our code in 2009...

dnahodil:
@jkburges see https://github.com/aodn/aodn-portal/pull/2015#issuecomment-145695959

jkburges:
> my only concern was regarding if it was possible to get code and use it without seeing the license text

With git, unlike some other version control systems, you have to go somewhat out of your way to get only *part* of a repository.  So under normal git usage, you would get the LICENSE file.  Whether or not someone sees it, well...  they may also not see the header in a source file :smile:

danfruehauf:
:+1:

I'll do the same for chef

Issue #1796, opened 2015-06-15
Multiple maps created when using share uuid's


pmbohm:
## To recreate
Add:
https://imos.aodn.org.au/imos123/home?uuid=4402cb50-e20a-44ee-93e6-4728259250d2&uuid=52f09a23-63a2-4c14-8b3b-1fc7c8167281&uuid=8964658c-6ee1-4015-9bae-2937dfcc6ab9&uuid=b4e742f9-58e5-46b8-b365-15430ea45599
## What happens
On step one the results will all have scroll bars, as there is actually multiple maps added
## What should happen
There should be no vertical scroll bars as there will only be one map per collection

PR #1868, opened 2015-08-06
New DataCollection model for Portal client-side

dnahodil:
**Not ready for merging**

This data model change is to support multi-WMS, and later multi-WFS, data collections.

The big change is that previously the Portal used either GeoNetworkRecords or Layers and it's main data model. Now the main data model is a DataCollection. Most UI components will work directly with a DataCollection. DataCollections will link to a MetadataRecord (formerly a GeoNetworkRecord) and one or more Layers.


dnahodil:
@markhepburn FYI these are the data model changes I was telling you about. This branch is still in development and will still change.

dnahodil:
@jkburges this will still need some git rebase lovin'
We can talk about that tomorrow.

dnahodil:
Bug spotted:
If you add a collection, then draw a bbox on step 2, and then add a second collection then bbox will disappear from the map. The constraint still appears to be applied correctly.

dnahodil:
It seems we've reintroduced #1796 as well. The fix for that was to only load one data collection from a URL. I think loading many layers for one data collection (which will now be allowed) has the same effect.

dnahodil:
:bug: I think I've broken ncWMS filtering.

dnahodil:
:bug: Style panel doesn't update as you change selected layers

jkburges:
Selecting an accordion folder should bring the associated layer to the top of the map, but it doesn't.

jkburges:
Selecting an accordion folder should bring the associated layer to the top of the map, but it doesn't.

jkburges:
> Selecting an accordion folder should bring the associated layer to the top of the map, but it doesn't.

Imma work on this one now.

jkburges:
The time picker only affects the initially selected layer (rather than the currently selected layer).  Imma fix that.

dnahodil:
Known to need tests:
DataCollection
DataCollectionStore
DataCollectionLayers
...

dnahodil:
:bug: Clear and reset link on step 3 doesn't work

web-app/js/portal/data/DataCollectionStore.js
```diff
@@ -0,0 +1,107 @@
...
+        });
+    },
+
+    _recordLoaded: function(dataCollection) {
+        dataCollection.loaded = true;
+        Ext.MsgBus.publish(PORTAL_EVENTS.DATA_COLLECTION_ADDED, dataCollection);
+    },
+
+    _removeFromLayerStore: function(dataCollection) {
+        this.layerStore.removeUsingOpenLayer(dataCollection.getSelectedLayer());
```
dnahodil: Check this. It'll need to remove all layers from the layer store, I think. Not just the current one. Unless we actually add/remove them as we change them. (Possibly we do in DataCollectionLayers or whatever it is called now)

dnahodil:
We need to confirm GoogleAnalytics tracking.
* For layer switcher
* Change to use DataCollection name rather than layer name
* Check that tracking is happening for changing scale range and style choice

PR #1879, opened 2015-08-17
Data Collection Layers refactor

jkburges:
* remove `DataCollection.getSelectedLayer()` - this must now be accessed via the `DataCollectionLayers` object.

I am now looking to:
1. remove unnecessary access to the `DataCollectionLayers` object
1. rename the class to something better (`LayerList` is the leading candidate currently)

1) above will hopefully allow me to delete much of the mocking present in this PR.


dnahodil:
I'll check this out

PR #1882, opened 2015-08-17
Filters to data collection

dnahodil:
This moves this logic from LayerDescriptor and now the logic of choosing a layer
name for some operations (size estimate, CSV download, etc.) is internal to
the DataCollection class.


jkburges:
I'll check this.

web-app/js/portal/data/DataCollection.js
```diff
@@ -67,6 +67,16 @@ Portal.data.DataCollection = function() {
         return applicableDownloadOptions;
     };
 
+    constructor.prototype.getFilterParams = function() {
+        var layer = this.getSelectedLayer();
```
jkburges: I'm surprised this works? `getSelectedLayer()` shouldn't be in `DataCollection` any more...

web-app/js/portal/data/DataCollection.js
```diff
@@ -67,6 +67,16 @@ Portal.data.DataCollection = function() {
...
+        var layerName = this.getDownloadLayerName();
+
+        return {
+            server: layer.server.uri,
+            layer: layerName
+        }
```
jkburges: ;

web-app/js/portal/data/DataCollection.js
```diff
@@ -131,6 +141,30 @@ Portal.data.DataCollection = function() {
...
+            if (this._workspaceFromName(linkName)) {
+                return linkName;
+            }
+            else {
+                return this._workspaceFromName(firstWmsLink.data.name) + ':' + linkName;
+            }
+        }
+    };
+
+    constructor.prototype._workspaceFromName = function(layerName) {
```
jkburges: This seems like something that should be a method of a `Layer` (of some sort).

web-app/js/portal/data/DataCollection.js
```diff
@@ -67,6 +67,16 @@ Portal.data.DataCollection = function() {
         return applicableDownloadOptions;
     };
 
+    constructor.prototype.getFilterParams = function() {
```
jkburges: What's the difference between `getFilter` and `getFilterParams`?  

Secondly, let's sit down together and review this class together because I think we should refactor it ASAP to be more like what we'd originally envisaged.

web-app/js/portal/data/DataCollection.js
dnahodil: Just trying to remember if I tested after the rebase. It might not work :S ...

web-app/js/portal/data/DataCollection.js
dnahodil: > let's ... review this class together  
Yeah okay

web-app/js/portal/data/DataCollection.js
dnahodil: It could be a util thing. The 'knowledge' of workspaces and layer names is from the OGC WMS spec.

dnahodil:
Good to go, again

PR #1884, opened 2015-08-18
google analytics for free text search facet

danfruehauf:


dnahodil:
The change is fine. It should probably have a unit test, though.

danfruehauf:
What do you want to test?

dnahodil:
That trackFacetUsage is called with the expected values. Other GA trackings are unit tested.

danfruehauf:
Sure.

dnahodil:
It's not required but this might be a good time to enable free-text search in the development environment. If it'll be enabled in prod soon we might as well start seeing/using it. What do you reckon?

danfruehauf:
> It's not required but this might be a good time to enable free-text search in the development environment. If it'll be enabled in prod soon we might as well start seeing/using it. What do you reckon?

Can do that. I'll amend the test and also the config part.

danfruehauf:
@dnahodil GTG

PR #1887, opened 2015-08-18
show error when loading invalid data collection

### LINKED ISSUE #1874: Unknown UUID in direct link breaks the Portal
**Steps to reproduce**
Visit https://imos.aodn.org.au/imos123/home?uuid=not-a-real-uuid

**What does happen**
* Arrive on step 2
* No collection added, but the 'No collections' message is not shown
* JS errors in console

**What should happen**
* No JS errors
* Probably should be on step 1, but if you are taken to step 2 the 'No collections' message should definitely be seen
* No collection added
* Maybe a message saying we couldn't load the collection you were after (check with @pblain)

danfruehauf:
* fix #1874, show message box when invalid data collection uuid is
   requested
 * land user on step 1 if data collection from url could not be
   loaded
 * if at least one collection was loaded, land in step 2

### DIFF
--- a/web-app/js/portal/data/GeoNetworkRecordFetcher.js
+++ b/web-app/js/portal/data/GeoNetworkRecordFetcher.js
@@ -38,12 +38,12 @@ Portal.data.GeoNetworkRecordFetcher = Ext.extend(Ext.util.Observable, {
             var metadataRecord = store.getAt(0);
             var dataCollection = Portal.data.DataCollection.fromMetadataRecord(metadataRecord);
 
-            if (!metadataRecord) {
-                _this._errorLoadingDataCollection(uuid);
-            }
-            else {
+            if (metadataRecord) {
                 _this.dataCollectionStore.add(dataCollection);
                 Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_DATA_COLLECTION, dataCollection);
+            }
+            else {
+                _this._errorLoadingDataCollection(uuid);
             }
         });
     },
@@ -57,10 +57,6 @@ Portal.data.GeoNetworkRecordFetcher = Ext.extend(Ext.util.Observable, {
         var params = Ext.urlDecode(getParams[1]);
 
         return params.uuid || [];
-    },
-
-    hasUuidsInUrl: function() {
-        return this.getUuidsFromUrl().length > 0;
     },
 
     _loadCollectionsFromUrl: function() {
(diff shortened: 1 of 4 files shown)


danfruehauf:
If that's our intention in this bug fix, I'll add tests. The behaviour is as described in the commit message.

web-app/js/portal/lang/en.js
```diff
@@ -149,6 +149,8 @@ OpenLayers.Lang.en = OpenLayers.Util.extend(OpenLayers.Lang.en, {
     btnSave: 'Save',
     btnCancel: 'Cancel',
 
+    errorLoadingCollection: 'Error loading collection with UUID \'{0}\'',
```
danfruehauf: @kereid @pblain If any of you could assist me with the wording in the message box, it'd be great. Current message is succinct and might not be very user friendly.

danfruehauf:
@jkburges was happy with that fix, so perhaps just look at the wording as the last thing here.

web-app/js/portal/data/GeoNetworkRecordFetcher.js
```diff
@@ -38,11 +38,20 @@ Portal.data.GeoNetworkRecordFetcher = Ext.extend(Ext.util.Observable, {
...
 
-            _this.dataCollectionStore.add(dataCollection);
-            Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_DATA_COLLECTION, dataCollection);
+            if (!metadataRecord) {
+                _this._errorLoadingDataCollection(uuid);
+            }
+            else {
+                _this.dataCollectionStore.add(dataCollection);
+                Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_DATA_COLLECTION, dataCollection);
+            }
```
dnahodil: Can I suggest we swap this condition around? I think it'd be nicer because then the condition won't be negated and the happy-path behaviour comes first.

jkburges:
I'm down with this change :+1:

web-app/js/portal/ui/MainPanel.js
```diff
@@ -24,7 +24,7 @@ Portal.ui.MainPanel = Ext.extend(Ext.Panel, {
         });
 
         var config = Ext.apply({
-            activeItem: fetcher.hasUuidsInUrl() ? TAB_INDEX_VISUALISE : TAB_INDEX_SEARCH,
+            activeItem: TAB_INDEX_SEARCH,
```
dnahodil: It's nice that we got rid of this switch!

dnahodil:
Changes made @jkburges

web-app/js/portal/lang/en.js
pblain: @danfruehauf , I think the message should be more specific - i.e. indicate that the cause of the error is an invalid UUID.  'The data collection cannot be loaded because the UUID \'{0}\' does not exist'.

dnahodil: The wording has been updated.

Issue #1878, opened 2015-08-17
Actions in the Portal not currently tracked by Google Analytics


dnahodil:
This is a (possibly incomplete) list of actions in the Portal which are not currently tracked with Google Analytics.

* Change opacity
* Change scale range
* Change style for layer
* Data collection visible/hidden on map
* ncWMS date filtering
* ncWMS data pick which time to see on map
* Remove button (step 3)
* Clear and reset (step 3)

@pblain FYI

danfruehauf:
> This is a (possibly incomplete) list of actions in the Portal which are not currently tracked with Google Analytics.
> 
>     Change opacity
>     Change scale range
>     Change style for layer - #1893
>     Data collection visible/hidden on map - #1893
>     ncWMS date filtering
>     ncWMS data pick which time to see on map
>     Remove button (step 3) - #1893
>     Clear and reset (step 3) - #1893

danfruehauf:
Still left after #1893:
* Change opacity
* Change scale range
* ncWMS date filtering
* ncWMS data pick which time to see on map

jkburges:
How about pan and zoom operations the map, or more interestingly, what areas are people looking at?

PR #1893, opened 2015-08-19
moar google analytics

danfruehauf:
Addresses some of the google analytics metrics we're lacking. See also #1878.

### DIFF
--- a/src/test/javascript/portal/cart/DownloadPanelSpec.js
+++ b/src/test/javascript/portal/cart/DownloadPanelSpec.js
@@ -15,6 +15,7 @@ describe("Portal.cart.DownloadPanel", function() {
         });
         spyOn(downloadPanel, 'generateContent');
         spyOn(downloadPanel, '_generateBodyContentForCollection');
+        spyOn(window, 'trackDataCollectionSelectionUsage');
     });
 
     afterEach(function() {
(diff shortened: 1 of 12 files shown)


dnahodil:
I'll review

dnahodil:
@jkburges, I've written a bunch of code for this PR now, too. So it probably wants a third person to run their eyes over it.

src/test/javascript/portal/cart/DownloadPanelSpec.js
```diff
@@ -81,11 +82,18 @@ describe("Portal.cart.DownloadPanel", function() {
     });
 
     describe('clear all', function() {
-        it('calls to active geo network record store remove all', function() {
+        beforeEach(function() {
             window.setViewPortTab = jasmine.createSpy();
             downloadPanel._clearAllAndReset();
+        });
+
+        it('calls to active geo network record store remove all', function() {
```
jkburges: The name of this `it` is outdated.

jkburges:
In general, I think it's worth investigating re-jigging our GA code alongs the lines of:

* intercept all events
* filter out, by originating component and event type, the ones that we want to send to GA
* one function per component of interest, that can generate an object to send to GA from itself and the event

The benefit of this is that we can more easily add in new GAs, and see in one place what we're collecting.  We also don't end up with GA code intermixed with the other business/UI logic.

dnahodil:
Having just done this work I think that idea has real merit. Maybe before we do more of this backlog item we should try a PoC and see how it might look.

dnahodil:
I've added a new ticket on the board for someone to have a crack at that approach.

PR #1896, opened 2015-08-19
Remove dependency on `getSelectedLayer()` where possible

jkburges:
It's preferable for components to be driven off DataCollection and only off selected layer where absolutely necessary.

### DIFF
--- a/web-app/js/portal/cart/BodaacDownloadHandler.js
+++ b/web-app/js/portal/cart/BodaacDownloadHandler.js
@@ -77,8 +77,7 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
                 filters: collection.getFilters()
             });
 
-            var wmsLayer = collection.getLayerState().getSelectedLayer();
-            return wmsLayer._buildGetFeatureRequestUrl(
+            return OpenLayers.Layer.WMS.prototype._buildGetFeatureRequestUrl(
                 _this._resourceHref(),
                 _this._layerName(),
                 OpenLayers.Layer.DOWNLOAD_FORMAT_CSV,
--- a/web-app/js/portal/cart/BodaacDownloadHandler.js
+++ b/web-app/js/portal/cart/BodaacDownloadHandler.js
@@ -77,7 +77,7 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
                 filters: collection.getFilters()
             });
 
-            return OpenLayers.Layer.WMS.prototype._buildGetFeatureRequestUrl(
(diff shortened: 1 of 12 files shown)


web-app/js/portal/cart/PythonDownloadHandler.js
```diff
@@ -37,7 +37,7 @@ Portal.cart.PythonDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
         var _this = this;
 
         return function(collection) {
-            return collection.getLayerState().getSelectedLayer().getFeatureRequestUrl(
+            return OpenLayers.Layer.WMS.prototype.getFeatureRequestUrl(
```
jkburges: I'm not sure about this myself - i.e. where should this method live?  It's pretty WMS/WFS specific, so tacking it on to `OpenLayers.Layer.WMS`, as it already was, seems ok.  But the way it's called here, and elsewhere, is a bit unusual in our code base.

jkburges:
There's probably some more that can go, but these were the easier ones.

web-app/js/portal/cart/BodaacDownloadHandler.js
```diff
@@ -77,8 +77,7 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
                 filters: collection.getFilters()
             });
 
-            var wmsLayer = collection.getLayerState().getSelectedLayer();
-            return wmsLayer._buildGetFeatureRequestUrl(
+            return OpenLayers.Layer.WMS.prototype._buildGetFeatureRequestUrl(
```
dnahodil: If we're calling out to this method it probably shouldn't be named with an underscore any more. It is no longer "private"/"internal" as an underscore indicates.

web-app/js/portal/cart/PythonDownloadHandler.js
dnahodil: Given that we're not calling this function on a WMS layer it is essentially a static method. So we could have a util class to house it. WmsRequestUtils, or something. At least then we wouldn't be calling it on an object's prototype.

dnahodil:
This looks good, @jkburges. Just those two comments.

web-app/js/portal/cart/PythonDownloadHandler.js
jkburges: I'd prefer to leave it in this file, rather than introduce a `utils` class, as then tend to become dumping grounds for lots of unrelated stuff. 

The namespace still makes sense also, so I'd like to keep that the same.

Maybe I can just drop the `prototype`?

dnahodil: That's fair. I am happy to leave it here (with or without the *prototype* as needed). We can always move it later if we come up with a better place.

jkburges:
image

PR #1902, opened 2015-08-20
Update wording of invalud data collection UUID message

dnahodil:


web-app/js/portal/lang/en.js
```diff
@@ -149,7 +149,7 @@ OpenLayers.Lang.en = OpenLayers.Util.extend(OpenLayers.Lang.en, {
     btnSave: 'Save',
     btnCancel: 'Cancel',
 
-    errorLoadingCollection: 'Error loading collection with UUID \'{0}\'',
+    errorLoadingCollectionMessage: 'The data collection could not be loaded because a data collection with UUID \'{0}\' does not exist.',
```
jkburges: I assume this wording came from above? Isn't it the most likely case that the collection will be currently disabled, rather than not existing at all?

dnahodil: It was based on Pete's recommendation here:
https://github.com/aodn/aodn-portal/pull/1887#discussion_r37492881
(I just tweaked it a bit)

Issue #1870, opened 2015-08-12
Temporal extent on download step not updated after 'clear subset'


jkburges:
**Steps to reproduce:**
1. Add an ncwms collection, e.g. ACORN
2. Edit the "from date" in the subset panel (to a more recent year than the default, for example)
3. Click 'clear subset'
4. Go to step 3.

**What should happen:**
The temporal extent shown on step 3 should reflect the full extent of the collection.

**What does happen:**
The temporal extent shows what was set in point 2 in the steps to reproduce above.

Note: seems as though this bug was introduced in https://github.com/aodn/aodn-portal/commit/6d13c170cc6e746e (according to `git bisect`).

dnahodil:
My work on this is stalled until I can answer this question: https://www.sencha.com/forum/showthread.php?304418-Is-it-possible-to-make-an-Ext.data.Record-class-which-extends-Observable&p=1112488

dnahodil:
Actually, I found a fix that doesn't require DataCollection to be Observable (re: that link).

PR #1903, opened 2015-08-20
Remove DATA_COLLECTION_MODIFIED event as it is unused

dnahodil:


jkburges:
Do we know what this used to be used for? We haven't inadvertently removed some functionality, have we?  Or just a hack-around?

dnahodil:
It used to be used when data collections were rearranged. But we don't do that anymore.

dnahodil:
See 861327666bead2a5b1aa476d47e3c6bd1893e5bb

dnahodil:
Actually. Hold tight on this change.

dnahodil:
There was another usage of this event which was removed. And it might be related to #1870.

dnahodil:
Removed in 7b98dbcd4499c23c9cdd4b0d9f75bd976e0b20e9

PR #1904, opened 2015-08-20
Remove knowledge of ncwmsParams from DataCollection

dnahodil:
We have removed references to ncwmsParams from DataCollection. We now
convert ncwmsParams into a list of filters and store them as such.
Then when they are used (2 places) we decode them back into ncwmsParams.
These conversion steps will be eliminated when
https://github.com/aodn/backlog/issues/191 is completed.

### DIFF
--- a/src/test/javascript/portal/details/NcWmsPanelSpec.js
+++ b/src/test/javascript/portal/details/NcWmsPanelSpec.js
@@ -32,7 +32,7 @@ describe('Portal.details.NcWmsPanel', function() {
 
         dataCollection = {
             getUuid: returns(45678),
-            updateNcwmsParams: jasmine.createSpy('updateNcwmsParams'),
+            setFilters: jasmine.createSpy('setFilters'),
             getLayerState: returns(layerState)
         };
 
@@ -68,7 +68,7 @@ describe('Portal.details.NcWmsPanel', function() {
 
         it('updates the record when panel is created', function() {
             ncwmsPanel._initWithLayer();
-            expect(ncwmsPanel.dataCollection.updateNcwmsParams).toHaveBeenCalled();
+            expect(ncwmsPanel.dataCollection.setFilters).toHaveBeenCalled();
         });
 
         it('updates the date when the start date changes via edit', function() {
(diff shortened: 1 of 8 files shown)


web-app/js/portal/data/DataCollection.js
```diff
@@ -97,50 +97,6 @@ Portal.data.DataCollection = function() {
...
-            name: 'time',
-            value: newFilterValue,
-            visualised: true
-        });
-
-        this.setFilters([
-            dateFilter
-        ]);
-    };
-
```
dnahodil: :arrow_up: This is the headline change

dnahodil: :arrow_up: This is the headline change

src/test/javascript/portal/details/NcWmsPanelSpec.js
```diff
@@ -178,6 +178,85 @@ describe('Portal.details.NcWmsPanel', function() {
...
+        it('updated start date', function() {
+
+            var testStartDate = moment();
+
+            var returnValue = ncwmsPanel._ncwmsParamsAsFilters(testStartDate, moment('invalid date'), null);
+
+            expectedNcwmsParamsFilter.dateRangeStart = testStartDate;
+            expectedDateFilterConfig.value.fromDate = testStartDate.toDate();
+
+            delete returnValue[1].hasValue; // Remove function so it isn't compared against
```
dnahodil: I did this to get the tests to work. The hasValue() functions weren't passing the equality check (I'm guessing because the scope was different). This is a simple way to take them out of the equation.

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -324,7 +324,46 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
         var dateRangeStart = this._getDateFromPicker(this.startDateTimePicker);
         var dateRangeEnd = this._getDateFromPicker(this.endDateTimePicker);
 
-        this.dataCollection.updateNcwmsParams(dateRangeStart, dateRangeEnd, geometry);
+        this.dataCollection.setFilters(this._ncwmsParamsAsFilters(dateRangeStart, dateRangeEnd, geometry));
+    },
+
+    _ncwmsParamsAsFilters: function(dateRangeStart, dateRangeEnd, geometry) {
```
jkburges: Is this perhaps worth of it's own class?  In fact, can we introduce a `Portal.filter.NcwmsDateFilter` (or whatever) - then this starts to be more like the other filter panel - and should make it easier to merge the two.  I'm not really up with the filter and filter UI design at the moment though, so it's just a thought.

dnahodil: I had considered that but as we wouldn't actually end up with a Portal.filter.Ncwms*xxx*Filter (it will use the existing Date and Geographic filters) it seemed that an anonymous object was fine in the interim.

jkburges: Okes.

PR #1909, opened 2015-08-24
Introduced class `DataCollectionLayerAdapter`

jkburges:
This class can be used when we want a data collection to behave somewhat like a single layer (in terms of getting and setting certain attributes, like opacity, for example).

### DIFF
--- a/web-app/js/portal/data/DataCollectionLayerAdapter.js
+++ b/web-app/js/portal/data/DataCollectionLayerAdapter.js
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2015 IMOS
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ *
+ */
+
+Ext.namespace('Portal.data');
+
+//
+// This class implements various OpenLayers.Layer functions using a DataCollection (and its associated layers).
+// It allows client code to essentially treat a DataCollection as it would a Layer, without having to use code
+// such as:
+//
+//    dataCollection.getLayerState().getSelectedLayer()...
+//
+// all over the place.
+//
+Portal.data.DataCollectionLayerAdapter = Ext.extend(Ext.util.Observable, {
+
+    constructor: function(config) {
+        Ext.apply(this, config);
+
+        // TODO: rename - conflicts with `LayerState` class.
+        this._layerState = {};
+
+        this._copyAttributesFromSelectedLayer();
+        this.layerState.on('selectedlayerchanged', function(newlayer) {
+            console.log('copyAttrs');
+            this._copyAttributesFromSelectedLayer();
+        }, this);
+
+        Portal.data.DataCollectionLayers.superclass.constructor.call(this, config);
+    },
+
+    getSelectedLayer: function() {
+        return this.layerState.getSelectedLayer();
+    },
+
+    _eachLayer: function(fn, scope) {
+        this.layerState._eachLayer(fn, scope);
+    },
+
+    isLoading: function() {
+        return this.getSelectedLayer().loading;
+    },
+
+    _setLayerProperty: function(key, value) {
+        this._layerProperties()[key] = value;
+    },
+
+    _getLayerProperty: function(key) {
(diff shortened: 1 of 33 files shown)


dnahodil:
:eyes:

web-app/js/portal/data/DataCollectionLayerAdapter.js
```diff
@@ -0,0 +1,133 @@
...
+            'bboxMinX',
+            'bboxMinY',
+            'opacity',
+            'projection'
+        ], function(attr) {
+            this[attr] = this.getSelectedLayer()[attr];
+        }, this);
+    },
+
+    _is130: function() {
```
dnahodil: This shouldn't be an underscore method any more, we obviously use it from outside the class (as we've now added it to our adapter). It could probably also use a better name...

web-app/js/portal/data/DataCollectionLayerAdapter.js
```diff
@@ -0,0 +1,133 @@
...
+//    dataCollection.getLayerState().getSelectedLayer()...
+//
+// all over the place.
+//
+Portal.data.DataCollectionLayerAdapter = Ext.extend(Ext.util.Observable, {
+
+    constructor: function(config) {
+        Ext.apply(this, config);
+
+        this._layerProperties = {};
```
dnahodil: Renaming this from _layerState is a good change.

web-app/js/portal/data/DataCollectionLayerAdapter.js
```diff
@@ -0,0 +1,133 @@
...
+
+        this._copyAttributesFromSelectedLayer();
+        this.layerState.on('selectedlayerchanged', function(newlayer) {
```
dnahodil: New layer parameter is unused and can be removed

web-app/js/portal/data/DataCollectionLayerAdapter.js
```diff
@@ -0,0 +1,133 @@
...
+            console.log('copyAttrs');
```
dnahodil: Missed one

web-app/js/portal/details/DataCollectionDetailsPanel.js
```diff
@@ -13,7 +13,7 @@ Portal.details.DataCollectionDetailsPanel = Ext.extend(Ext.Panel, {
 
         var tabPanelForLayer = this._initSubsetItemsTabPanel(cfg);
 
-        this.createTools(cfg.dataCollection.getLayerState());
+        this.createTools(cfg.dataCollection.getLayerAdapter());
 
         cfg.dataCollection.getLayerState().on('loadstart', function() {
```
dnahodil: Just a question about using the new adapter vs using the state. Why do we listen to layer state events here but use the adapter to check if the layer is loading?

dnahodil:
There's a few comments and questions there @jkburges. I think overall the adapter looks good, I think I just want to understand a bit better when to use the layerState instead of the adapter. (related to my last question)

web-app/js/portal/data/DataCollectionLayerAdapter.js
jkburges: What *does* it actually mean?

jkburges: Oh, I see - WMS version 1.3.0.

Withdrawn.

jkburges: The only place this function is used is here: https://github.com/aodn/aodn-portal/blob/master/web-app/js/portal/ui/MapPanel.js#L130

W.r.t. the ticket referenced in the comment there, do we actually still need this check (the issue is marked as resolved)?

If so, I suggest to remove all of this code, in a separate PR.

dnahodil: Yeah, I reckon we get rid of this method then. Separate PR is fine.

jkburges:
Updated.

PR #1911, opened 2015-08-24
GeoSelection Facet buttons overlaying the map

### LINKED ISSUE #1861: Geographic boundary usability not ideal
There are a couple of issues with the geographic boundary in my opinion. One of them is that by default it opens in such a way that the `Go` and `Clear` buttons are not visible:
geographicboundary1

The second can be reproduced by:
 * Adding a collection to the map
 * Going back to step 1
 * Opening the geographic boundary facet
 * After drawing a polygon, users will be tempted to click on the `Next` button:
geographicboundary2

I think those are usability issues which should be addressed. Mind you, most of our backend effort is to provide a proper spatial index of the data. By having a suboptimal geographic boundary facet we are defeating that purpose a bit.

pmbohm:
Fixes #1861

### DIFF
--- a/web-app/js/portal/search/GeoSelectionPanel.js
+++ b/web-app/js/portal/search/GeoSelectionPanel.js
@@ -48,11 +48,9 @@ Portal.search.GeoSelectionPanel = Ext.extend(Ext.Panel, {
             toolTemplate: new Ext.Template(''),
             items:[
                 this.facetMap,
-                new Ext.Spacer({
-                    height: 4
-                }),
                 new Ext.Container({
                     layout: 'hbox',
+                    cls: "geoSelectionPanelbuttons",
                     items: [
                         this.goButton = new Ext.Button({
                             text:OpenLayers.i18n("goButton")
--- a/web-app/css/general.css
+++ b/web-app/css/general.css
@@ -934,6 +934,12 @@ div.allwhite {
     float: left;
 }
 
+.geoSelectionPanelButtons {
+    position: relative;
+    top: -35px;
+    right: -210px
+}
+
 .resultsTextBody {
     margin: 5px;
     width: 540px;


pmbohm:
This solves the issue of the GeoSelection Facet buttons being obscured or their meaning being misinterpreted. 
The solution uses CSS to position the buttons over the map. They retain the correct appearance being the same as other button type links in the site. The cursor is correct when mouseovering.
The alternative was programatically difficult if not impossible, adding controls to the Openlayers map which i tried and timed out after 4 hours of work

pmbohm:
Oh and the clear button is now context aware

pmbohm:
screen shot 2015-08-24 at 3 47 15 pm

web-app/js/portal/search/GeoSelectionPanel.js
```diff
@@ -106,6 +108,7 @@ Portal.search.GeoSelectionPanel = Ext.extend(Ext.Panel, {
 
     _clearFacetGeometry: function() {
         this.facetMap.clearGeometry();
+        this.clearButton.setDisabled(!this.facetMap.hasCurrentFeature());
```
dnahodil: Would it be more clear just to say `this.clearButton.setDisabled(true);` in this instance? or is there a time when it might not evaluate to true?

dnahodil:
Currently, if I draw a polygon I can't clear it until I have first clicked 'go'. Not sure if that was intended but it seems odd.

dnahodil:
Changes look good, @pmbohm. I've just made two comments.

pmbohm:
@dnahodil Yes so "Go is really redundant. We should just Go then? There aren't any real overheads as the response seems quick

dnahodil:
My point is just that if I've drawn a polygon, even if I haven't searched with it yet, I should be able to clear it.

re: the go button being redundant, you'd have to talk to Pete.

pmbohm:
@dnahodil This is now revised after speaking to @pblain

PR #1912, opened 2015-08-24
Step 1 free text filter now called 'Keyword'

pmbohm:


dnahodil:
@pmbohm Just checking that this wording change came from talking with Pete?

pmbohm:
Yes it's a @pblain foreign order @dnahodil

dnahodil:
No problems

PR #1913, opened 2015-08-24
Hide toggle arrow for FreeTextSearchPanel

### LINKED ISSUE #1891: Free-text search shows expand/collapse arrow
**Steps to reproduce**
Visit step 1 on a Portal with free-text search enabled

**What does happen**
The free-text search has a expand/collapse arrow
freetext

**What should happen**
To match the other facets it shouldn't have a expand/collapse arrow

pmbohm:
Created an empty template for this panel's tools
Fixes #1891


pmbohm:
It seems that the other facet panel headers can include code like:
```
tools: [
    {
        id: 'toggle',
        hidden: true
   }
]

```
This should be enough to hide it, but maybe as it is the only Tool, this code alone had no effect. The toolTemplate is respected however.

dnahodil:
That's weird that `hidden: true` didn't work.

PR #1916, opened 2015-08-25
Remove get_____Links from DataCollection

dnahodil:
This removes the individual get______Links() methods from DataCollection and has just one method getLinksByProtocol() which you can pass the protocols you are interested in. This new function is also used for getting the point-of-truth link from a DataCollection.


dnahodil:
Imma fix the conflicts

PR #1919, opened 2015-08-26
Additional track usage

pmbohm:


pmbohm:
Fixes #1878

src/test/javascript/portal/details/NcWmsPanelSpec.js
```diff
@@ -75,12 +76,14 @@ describe('Portal.details.NcWmsPanel', function() {
             ncwmsPanel._addTemporalControls();
             ncwmsPanel.startDateTimePicker.fireEvent('change', ncwmsPanel.startDateTimePicker);
             expect(ncwmsPanel._onDateSelected).toHaveBeenCalled();
+            expect(window.trackUsage).toHaveBeenCalled();
```
dnahodil: These tests should check that the correct values are passed in (I think there are 4 here where we don't).

web-app/js/portal/details/LayerControlPanel.js
```diff
@@ -19,37 +19,50 @@ Portal.details.LayerControlPanel = Ext.extend(Ext.Container, {
             this.items.push(layerSelector, { xtype: 'spacer', height: 10 });
         }
 
-        this.items.push(this._newOpacitySlider());
+        this.opacitySlider = new Portal.common.LayerOpacitySliderFixed({
+            layer: this.layer,
```
dnahodil: We specifically moved the creation of the controls into their own methods here to keep the constructor tidy. I suggest we put this control creation back into it's own method.

When we configure it we can pass a `listeners` config option. Something like this:
``` javascript
listeners: {
    'changecomplete': this.doTracking,
    scope: this
}
```

So we wouldn't need to keep a reference to the slider (i.e. `this.opacitySlider`) at all.

web-app/js/portal/details/StylePanel.js
```diff
@@ -152,6 +152,11 @@ Portal.details.StylePanel = Ext.extend(Ext.Container, {
         this.legendImage.show();
 
         this.styleCombo.setValue(styleName);
+
+        if (scaleRange) {
+            var scaleRangeString = toNSigFigs(scaleRange.min, 2) + "," + toNSigFigs(scaleRange.max, 2);
+            trackLayerControlUsage('layerControlTrackingActionStyleRange', scaleRangeString, this.dataCollection.getTitle());
+        }
```
dnahodil: This seems to me to be in the wrong place. It should just go in the NcWmsScaleRangeControls.applyNewScale() method.

dnahodil:
Comments made @pmbohm

pmbohm:
Yes, I can see now that i had a reference to the title right there :/ @dnahodil

pmbohm:
@dnahodil New pull request up now

web-app/js/portal/details/LayerControlPanel.js
dnahodil: I see that we're now passing the listeners as part of the config, @pmbohm. So I'd like to see this component creation moved back into it's own function as it was. We no longer need to assign `this.opacitySlider` because it isn't referenced anywhere else.

pmbohm: @dnahodil Agh yes i did understand you. It was needed for tests, to get a reference to this.dataCollection for the tracking

dnahodil:
Merged. Thanks, @pmbohm

PR #1920, opened 2015-08-26
Increase logging in order to debug travis hanging

jkburges:
e.g.: https://travis-ci.org/aodn/aodn-portal/builds/77101444

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -297,7 +297,7 @@
             <plugin>
                 <groupId>com.github.searls</groupId>
                 <artifactId>jasmine-maven-plugin</artifactId>
-                <version>1.3.1.2</version>
+                <version>1.3.1.4</version>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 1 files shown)


jkburges:
Yes, it took me all morning just to figure out the correct way to configure this.

jkburges:
I'd recommend merging this to master ASAP, even though I'm not sure it's actually fixing the problem.  But, if it *is* merged, and we don't see any travis hangs for several PRs, then perhaps we can say that it is fixed.

dnahodil:
Did you want to rebase your changes to just reflect the final changes?

jkburges:
Yeah that's a good idea.

jkburges:
Done, although Murphy's Law says that this latest commit will hang.

dnahodil:
To Hell with Murphy!

dnahodil:
:8ball: Magic 8-ball says: Outlook is not good

jkburges:
Boo. Don't bother merging this. :sob:

PR #1921, opened 2015-08-26
Remove unused behaviour from MapOptions

dnahodil:


dnahodil:
@pmbohm I've removed the event as per our discussion

PR #1925, opened 2015-08-26
Using generic term - MetadataRecord instead of GeoNetworkRecord

pmbohm:


web-app/js/portal/ui/MainPanel.js
```diff
@@ -46,15 +46,15 @@ Portal.ui.MainPanel = Ext.extend(Ext.Panel, {
 
         Portal.ui.MainPanel.superclass.constructor.call(this, config);
 
-        Ext.MsgBus.subscribe(PORTAL_EVENTS.VIEW_DATA_COLLECTION, this._onViewGeoNetworkRecord, this);
-        Ext.MsgBus.subscribe(PORTAL_EVENTS.DATA_COLLECTION_ADDED, this._onGeoNetworkRecordAdded, this);
+        Ext.MsgBus.subscribe(PORTAL_EVENTS.VIEW_DATA_COLLECTION, this._onViewMetadataRecord, this);
```
jkburges: These probably should be `_onViewDataCollection` and `_onDataCollectionAdded`, respectively.

jkburges: Re `VIEW_DATA_COLLECTION` - usually, event are to indicate that something just happened, not a command to do something, such as this one.  If it wasn't for the multi-select functionality, I think this event could go - with the `MainPanel` simply electing to switch to step 2 on `DATA_COLLECTION_ADDED`.

Maybe there's some other way we could do this - thoughts (cc @dnahodil)?

dnahodil: I agree that it is different to normal events in that way - it's a command rather than an announcement. If people are happy I will have a look into it (as a separate PR) and see if it can be done in a better way.

jkburges: > If people are happy I will have a look into it

image

pmbohm:
@dnahodil @jkburges Merge conflict fixed. This is still up for merging right?

dnahodil:
@pmbohm I'm going to look into alternatives for VIEW_DATA_COLLECTION as a side issue (i.e. it won't stop this getting merged).

There's still a comment there from @jkburges for you to consider.

pmbohm:
@dnahodil @jkburges meme considered

PR #1926, opened 2015-08-26
#1870 - On reset apply the filter values from controls

### LINKED ISSUE #1870: Temporal extent on download step not updated after 'clear subset'

dnahodil:
Fixes #1870

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -44,7 +44,7 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
         this._attachTemporalEvents();
         this._attachSpatialEvents();
         this._removeLoadingInfo();
-        this._applyFilterValuesFromMap();
+        this._applyFilterValuesToCollection();
     },
 
     _onSelectedLayerChanged: function(newLayer) {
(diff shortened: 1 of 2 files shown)


web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -383,6 +383,7 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
         this._updateTimeRangeLabel();
 
         this._setLayerSubsetExtent();
+        this._applyFilterValuesFromMap();
```
jkburges: I think this is a bit confusing.  Really, what we want to express at this level is `_applyFilterValuesToCollection`.  This method should in turn do the bit about getting the constraint from the map (it shouldn't really be getting it from *there*, but that's a different story).

dnahodil:
Changes made. I think I addressed your comment, if you meant that the important thing is that the filters are being applied *to the collection* and not the the filters were *coming from the map*.

PR #1930, opened 2015-08-27
Don't process DataCollection links through the LinkStore

dnahodil:
We no longer process links by feeding them through the LinkStore. Places where they are used have therefore changed, but this is positive in most cases (no need to dereference their `data` property because they are no longer Records). This was the last of the tested behaviour in GeoNetworkRecord so that test spec has been removed.


jkburges:
Does this need a rebase?  No conflicts, just some removals which I think were from another PR.

dnahodil:
I'll rebase it to be sure

dnahodil:
Rebased

PR #1933, opened 2015-08-28
Remove VIEW_DATA_COLLECTION event

dnahodil:


web-app/js/portal/data/MetadataRecordFetcher.js
```diff
@@ -40,7 +38,7 @@ Portal.data.MetadataRecordFetcher = Ext.extend(Ext.util.Observable, {
 
             if (metadataRecord) {
                 _this.dataCollectionStore.add(dataCollection);
-                Ext.MsgBus.publish(PORTAL_EVENTS.VIEW_DATA_COLLECTION, dataCollection);
+                viewport.setActiveTab(TAB_INDEX_VISUALISE);
```
jkburges: Not sure how I feel about adding a dependency on this global... I'm ok to merge this for now though, having said that.

dnahodil: I agree. However, it is something we do already in code [citation needed] so it isn't introducing any new bad practices. It'd be great if we think of something to replace both of these instances!

dnahodil: https://github.com/aodn/aodn-portal/blob/56a57e7a97c63ef6a02d08e790be3dd43808ad3b/web-app/js/portal/search/FacetedSearchResultsDataView.js#L319

PR #1934, opened 2015-08-28
Download handlers not in DataCollection

dnahodil:


web-app/js/portal/cart/DownloadHandler.js
```diff
@@ -66,3 +66,39 @@ Portal.cart.DownloadHandler = Ext.extend(Object, {
...
+
+Portal.cart.DownloadHandler._downloadHandlerConstructorForProtocol = function(protocol) {
+    var mapping = {
+        'OGC:WFS-1.0.0-http-get-capabilities': [
+            Portal.cart.WfsDownloadHandler,
+            Portal.cart.PythonDownloadHandler
+        ],
+        'IMOS:AGGREGATION--bodaac': Portal.cart.BodaacDownloadHandler,
+        'IMOS:AGGREGATION--gogoduck': Portal.cart.GogoduckDownloadHandler
+    };
```
dnahodil: I don't like the fact that this mapping lives here but I had real trouble moving it out as a "static constant". I also tried having a setup where each subclass of DownloadHandler would register itself along with the protocols it supported. I was really happy with this idea but I couldn't get that to work for the same reason (the problem boils down to the fact that DownloadHandler.js must be first so its subclasses can reference it, but that means that it can't reference its subclasses from a static field because they won't exist yet). I'd be happy to pair with someone if they have an idea of how we can improve where this mapping lives.

danfruehauf: That's fine. It's not too painful and probably now in a better place anyway.

PR #1938, opened 2015-08-31
Data Collection loads filters rather than UI component

dnahodil:
These changes make it so that DataCollection loads its own Filters as it is created and the GroupFilterPanel only listens for events (previously it initiated the loading, itself). Also, Filters now fire an even when their value is changed.

### DIFF
--- a/web-app/js/portal/utils/ObservableUtils.js
+++ b/web-app/js/portal/utils/ObservableUtils.js
@@ -1,0 +1,16 @@
+Ext.namespace('Portal.utils');
+
+Portal.utils.ObservableUtils = {
+    FUNCTIONS_TO_FORWARD: ['on', 'un', 'fireEvent'],
+
+    makeObservable: function(target) {
+        target._observable = new Ext.util.Observable();
+
+        Ext.each(this.FUNCTIONS_TO_FORWARD, function(funcName) {
+
+            target[funcName] = function() {
+                return this._observable[funcName].apply(target._observable, arguments);
+            };
+        });
+    }
+};
--- a/web-app/js/portal/utils/ObservableUtils.js
+++ b/web-app/js/portal/utils/ObservableUtils.js
@@ -1,7 +1,20 @@
 Ext.namespace('Portal.utils');
 
 Portal.utils.ObservableUtils = {
-    FUNCTIONS_TO_FORWARD: ['on', 'un', 'fireEvent'],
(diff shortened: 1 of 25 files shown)


dnahodil:
Ready for review.

danfruehauf:
@dnahodil Needs rebasing.

web-app/js/portal/utils/ObservableUtils.js
```diff
@@ -0,0 +1,16 @@
+Ext.namespace('Portal.utils');
+
+Portal.utils.ObservableUtils = {
+    FUNCTIONS_TO_FORWARD: ['on', 'un', 'fireEvent'],
```
danfruehauf: I think it'd be good if those could be controlled by the user of this interface. i.e. one could specify what events to forward. Makes this module more reusable down the track.

danfruehauf:
Altogether good. I think I understand what's happening here and that is a welcome change - separating the UI from the data model while having a clean interface between them.

web-app/js/portal/utils/ObservableUtils.js
dnahodil: Yeah, I just forwarded the events I needed. An alternative which I only thought of because of your comment, would be to get the function names programmatically from the Observable class so that you overwrote them all and didn't need to maintain a list of them all.

dnahodil:
Thanks for the review, Dan.

Yeah, the UI used to drive the loading of the Filters and it was a bit back-to-front. So now a DataCollection takes charge of loading its own filters. Doesn't change anything for the user but I think it models things in a nicer way.

web-app/js/portal/utils/ObservableUtils.js
dnahodil: Thinking a bit more I don't think I'd like to see the events being configurable when choose to make an object Observable with this class. As Observable is an interface we should really guarantee that it implements the functions which all Observable objects should have.

So I think I'll update this list to be complete, rather than just the ones I used.

dnahodil:
@danfruehauf I've made an update. The ObservableUtils.makeObservable() now applies all non-static functions defined by the Observable interface. I know it isn't what you suggested but I think it's the most correct thing to do. Let me know if you disagree!

danfruehauf:
I'd probably like to see a list of functions as a parameter, but that'll do as well. It's more complete, yet not configurable.

PR #1945, opened 2015-09-03
Clean up of zoom to layer

jkburges:
Bounds are taken directly from the data collection, rather than storing in each layer and then taking from there - that's just a round-a-bout and unnecessary way of doing it now that most components are 'data collection centric'.

This also addresses a usability issue whereby the 'zoom to layer' control is only shown for collections having a geo extent defined.  Ordinarily this should be the case for *every* collection, although this is the safer option and can probably also help avoid confusion when developing new collections.  /cc @kereid @pblain change to functionality

### DIFF
--- a/web-app/js/portal/ui/MapPanel.js
+++ b/web-app/js/portal/ui/MapPanel.js
@@ -23,8 +23,6 @@ Portal.ui.MapPanel = Ext.extend(Portal.common.MapPanel, {
             split: true,
             header: false,
             initialBbox: portalConfig.initialBbox,
-            enableDefaultDatelineZoom: portalConfig.enableDefaultDatelineZoom,
-            defaultDatelineZoomBbox: portalConfig.defaultDatelineZoomBbox,
             hideLayerOptions: this.appConfig.hideLayerOptions
         }, cfg);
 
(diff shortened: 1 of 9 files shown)


web-app/js/portal/ui/MapPanel.js
```diff
@@ -109,26 +107,6 @@ Portal.ui.MapPanel = Ext.extend(Portal.common.MapPanel, {
         this.map.setDefaultSpatialConstraintType(this.defaultSpatialConstraintType);
     },
 
-    zoomToLayer: function(openLayer) {
-        if (openLayer && openLayer.hasBoundingBox()) {
-            // build openlayer bounding box
-            var bounds = null;
-            if (openLayer.bboxMinY == -180 && openLayer.bboxMaxY == 180 && this.enableDefaultDatelineZoom) {
```
jkburges: This code was introduced in eef0d5b79f25.  It appears as though `enableDefaultDatelineZoom` is never set to `true` anymore, hence I don't *think* we need this logic.

danfruehauf:
I'll review.

jkburges:
@danfruehauf I just remembered - I meant to delete `enableDefaultDatelineZoom` altogether, since it's never used.

jkburges:
Done in 0a14017

danfruehauf:
Squash them?

jkburges:
Squashed

PR #1947, opened 2015-09-03
fix #1937 - do not show null values in filters

### LINKED ISSUE #1937: Step 2 filtering behaves oddly when empty string is an option
**Steps to reproduce**
Add the Argo collection and go to step 2
Open the drop-down menu for the Platform Number filter

**What does happen**
There is an erroneous option there with no text, which has visual artefacts and might confuse a user as if selected it will hide helpful text. Selecting it has no effect on the filtering of the layer.
string filter empty string
I suspect this is because the empty string is listed as one of the possible options for the filter.

**What should happen**
* The combo filter panel should remove the empty string from the range of values when they are loaded.
* No entry in drop-down for empty option

danfruehauf:


src/test/javascript/portal/filter/ui/ComboFilterPanelSpec.js
```diff
@@ -60,6 +60,14 @@ describe("Portal.filter.ui.ComboFilterPanel", function() {
         });
     });
 
+    describe('_filterData', function() {
+        it('removes empty and null values', function() {
+            var values = [ null, "cake", undefined, "stapler", "", " ", "valid value with space" ];
+            var filteredValues = [ [ "cake"] , [ "stapler"], ["valid value with space"] ];
```
danfruehauf: Let me fix the spacing on this line. I got it a bit wrong.

danfruehauf:
@jonescc Ready again.

PR #1950, opened 2015-09-03
fix #1935 - limit W E bbox to -179.99 179.99

### LINKED ISSUE #1935: No data downloaded when W and E bounding box coordinates are set to -180 and 180 respectively
App version: 4.12.0
Build date: Fri Aug 28 14:56:25 AEST 2015

<h3> Steps to reproduce: <h3>

Add any collection and specify the bounding box coordinates as follows W=-180, E=180, and whatever makes you happy for N and S (-40 and -90 respectively for instance).

Proceed to step 3 and download CSV.

<h3> What should happen: <h3>

The CSV file should have data in it for this set of coordinates.

<h3> What does happen: <h3>

The CSV file only contains the metadata header and data column names. Setting W and E to -179.99 and 179.99 solves this issue.

danfruehauf:
if the bbox is set to W E -180 180, nothing is downloaded

### DIFF
--- a/web-app/js/portal/details/BoxDisplayPanel.js
+++ b/web-app/js/portal/details/BoxDisplayPanel.js
@@ -68,10 +68,10 @@ Portal.details.BoxDisplayPanel = Ext.extend(Ext.Panel, {
     },
 
     _buildBoundingBox: function(config) {
-        this.northBL = this._buildCoord('northBL',-90,90);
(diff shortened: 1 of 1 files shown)


jonescc:
This worries me somewhat.  I think that's around 2 kilometres (at the equator) that can no longer be downloaded.

danfruehauf:
A proper fix will allow us to download around the anti-meridian. It's not a good fix, I agree.

PR #1952, opened 2015-09-04
fix #1944 - use geometry from event

### LINKED ISSUE #1944: Spatial Subset not applied on step 3 for ncWMS collections
App version: 4.12.1
            Build date: Tue Sep 01 15:30:10 AEST 2015
            Build number: 14
            Git ref: d6c409ad34d5a579c4508c81a0767f7f285fef3e

Steps to reproduce:

Add any ncwms collection such as ACORN or SRS.

Draw a bounding box on step 2 at some part over the collection

Proceed to step 3

What should happen:

See indicaqtion of spatial subset applied to collection and the co-ordinates

What does happen:

Only see default temporal extent that is applied

danfruehauf:
use geometry from event for ncwms layers

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -85,10 +85,10 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
             this.map.events.on({
                 scope: this,
                 'spatialconstraintadded': function(geometry) {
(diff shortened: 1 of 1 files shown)


danfruehauf:
Not entirely sure what was meant in ce76a592 (by @dnahodil) but this will fix that bug (which is pretty critical for the release.

I saw that `filter/ui/GeometryFilterService.js` behaves the same, so I assume it is OK.

PR #1953, opened 2015-09-04
issue 1935

### LINKED ISSUE #1935: No data downloaded when W and E bounding box coordinates are set to -180 and 180 respectively

danfruehauf:
fix #1935 by not normalizing longitudes that are valid (-180 <= lon <= 180)


grails-app/views/_js_includes.gsp
```diff
@@ -55,7 +55,7 @@
     <script type="text/javascript" src="${resource(dir: 'js/portal', file: 'portal.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal', file: 'ObjectFactory.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal', file: 'PortalEvents.js')}"></script>
-    <script type="text/javascript" src="${resource(dir: 'js/portal', file: 'WKTNormalised.js')}"></script>
+    <script type="text/javascript" src="${resource(dir: 'js/portal/prototypes', file: 'WKTNormalised.js')}"></script>
```
jonescc: Why prototypes?

danfruehauf: Why not? It's overriding an OpenLayers class. Still cleaner than having it on the root of the portal JS directory!! :)

jonescc: OK, can see why now

PR #1957, opened 2015-09-07
Don't try to use metadataRecord until we know that one has been found

### LINKED ISSUE #1956: Load Data Collection from URL has JavaScript errors
Spotted on master (i.e. not currently on systest) @ 597bfa8f690571ba0fd5aa450eea488ebbebdea6

**Steps to reproduce**
* Try to load a data collection by having the UUID in the URL

**What does happen**
* Errors appear in JavaScript console
* Data collection is not loaded

**What should happen**
* Data collection should load as normal

dnahodil:
Fixes #1956

### DIFF
--- a/web-app/js/portal/data/MetadataRecordFetcher.js
+++ b/web-app/js/portal/data/MetadataRecordFetcher.js
@@ -34,10 +34,11 @@ Portal.data.MetadataRecordFetcher = Ext.extend(Ext.util.Observable, {
             var store = new Portal.data.MetadataRecordStore();
             store.loadData(response.responseXML);
             var metadataRecord = store.getAt(0);
-            var dataCollection = Portal.data.DataCollection.fromMetadataRecord(metadataRecord);
 
             if (metadataRecord) {
+                var dataCollection = Portal.data.DataCollection.fromMetadataRecord(metadataRecord);
                 _this.dataCollectionStore.add(dataCollection);
+
                 viewport.setActiveTab(TAB_INDEX_VISUALISE);
             }
             else {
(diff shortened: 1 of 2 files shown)


web-app/js/portal/data/MetadataRecordFetcher.js
```diff
@@ -34,10 +34,11 @@ Portal.data.MetadataRecordFetcher = Ext.extend(Ext.util.Observable, {
             var store = new Portal.data.MetadataRecordStore();
             store.loadData(response.responseXML);
             var metadataRecord = store.getAt(0);
-            var dataCollection = Portal.data.DataCollection.fromMetadataRecord(metadataRecord);
 
             if (metadataRecord) {
+                var dataCollection = Portal.data.DataCollection.fromMetadataRecord(metadataRecord);
                 _this.dataCollectionStore.add(dataCollection);
+
```
danfruehauf: Why this new line?

dnahodil: Because loading a DataCollection and adding it to the store is a separate thing from setting the tab in a UI component (the line below). So I separated them.

danfruehauf: It's a 3 line block, is it really required?

dnahodil: It's not about how long the block is, it's to separate code that's doing two logically separate operations. The first two lines work together to load the collection and add it to the store (which are data model operations), and the last line sets some UI state.

danfruehauf: Just don't give me a hard time the next time I remove new lines.

Issue #1958, opened 2015-09-07
Load Data Collection from URL fails with fragment identifier


dnahodil:
Spotted by @pmbohm, see https://github.com/aodn/aodn-portal/issues/1956#issuecomment-138183839

**Steps to reproduce**
* Get a direct-link to a data collection in the Portal. 
* Append the hash symbol to the URL
* Try to load the URL in a browser

**What does happen**
* The data collection isn't loaded
* JS error occur in terminal

**What should happen**
* The data collection should be loaded or the user should get feedback that the uuid is unavailable

Note: It is valid for a URL to have a fragment identifier after a query string, so we should handle it.

pmbohm:
@dnahodil As for what it takes to get yourself a free hash char turning up at the end of the url, dunno yet what i do

dnahodil:
*Hash chars, hash chars, get 'em while they're hot. FREE to the first 100 callers!*

It seems you can get one by middle-clicking on the 1-2-3 tabs. There will be other ways, too.

PR #1960, opened 2015-09-07
Avoid generating URLs with hashes

pmbohm:
Stops the issue of appended hashes appearing on the Step1-2-3 links by using button tags instead of links.
See:
https://github.com/aodn/aodn-portal/issues/1958

### DIFF
--- a/grails-app/views/header/_viewPortLink.gsp
+++ b/grails-app/views/header/_viewPortLink.gsp
@@ -6,8 +6,8 @@
 
 --%>
 <div class="viewPortTab viewPortTabDisabled" id="viewPortTab${stepIndex}">
-  <a href="#" >
+  <button>
     <h1>${stepIndex + 1}</h1>
     <h2>${description}</h2>
-  </a>
+  </button>
 </div>
--- a/web-app/css/general.css
+++ b/web-app/css/general.css
@@ -306,6 +306,15 @@ Site navigation
     margin: 0;
 }
 
+.viewPortTab button {
+    background-color: inherit;
+    border: none;
+}
+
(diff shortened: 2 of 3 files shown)


dnahodil:
Change looks fine. Just one thing, I don't get a hand like the tabs are pointers anymore. Ref: http://stackoverflow.com/a/3088012/202224

pmbohm:
@dnahodil Thanks for giving me that 'pointer'

dnahodil:
@pmbohm *(groan)*

danfruehauf:
I think it introduces the following:
buttons

Is it because now those are buttons?

danfruehauf:
Could not reproduce on Chrome. Happens on firefox (40.0).

dnahodil:
I would recommend against removing this. I don't know it's proper name but it indicates which active element on a page has focus. If people navigate with keyboards of some other input device this is how they know what they're interacting with. Hiding it makes a site less accessible.

dnahodil:
I'm not sure why it wasn't present before this change - possibly some CSS somewhere which we've forgotten about or don't know about.

danfruehauf:
I don't think anyone navigates the portal with the keyboard. I might be surprised though.

I'll just not that this PR doesn't quite *fix* anything, yet it introduces a problem which @pblain might want to fix.

pmbohm:
What we have now is more 'correct'. The Anchor tag we had before was an old school hack we hung JS off. This PR fixes the fact that users appeared to be clicking a link to a destination. The URL the users were shown was a furfy. As for making our pages accessible, this should be our aim. We are so lucky its not yet the law here as it is in the USA

PR #1961, opened 2015-09-07
Bookmark hashes now stripped

### LINKED ISSUE #1958: Load Data Collection from URL fails with fragment identifier

pmbohm:
fixes #1958


dnahodil:
@pmbohm. Looks good. Good thing remembering to update the tests, too!

PR #1963, opened 2015-09-08
Tidy spatial Subset info on Step 3

pmbohm:


src/test/javascript/portal/prototypes/OpenLayersSpec.js
```diff
@@ -177,6 +177,12 @@ describe('OpenLayers', function() {
                 expect(OpenLayers.Geometry.fromWKT('POLYGON((1 2,3 4,1 2))').isBox()).toEqual(false);
             });
         });
+
+        describe('getPrettyBounds', function() {
+            it('returns true when box', function() {
```
dnahodil: This message does not match what we're actually testing

src/test/javascript/portal/prototypes/OpenLayersSpec.js
```diff
@@ -177,6 +177,12 @@ describe('OpenLayers', function() {
...
+                expect(OpenLayers.Geometry.fromWKT('POLYGON((1 2,3 2,3 4,1 4,1 2))').getPrettyBounds()).toEqual('1W 2S 3E 4N');
```
dnahodil: We should probably test with values with decimal components to ensure that toNSigFigs() is used.

dnahodil:
The code changes here look really good @pmbohm, and this will be a nice change to have.

I've just made 2 comments.

pmbohm:
@dnahodil tests now changed

dnahodil:
:+1:

PR #1965, opened 2015-09-08
Remove collection as buttons in step 3

pmbohm:


web-app/js/portal/lang/en.js
```diff
@@ -229,7 +228,7 @@ OpenLayers.Lang.en = OpenLayers.Util.extend(OpenLayers.Lang.en, {
     depthLabel: 'Depth:',
     elevationLabel: 'Elevation:',
 
-    removeDataCollection: 'Remove data collection',
+    removeDataCollection: 'Remove this data collection from the Subset and Download pages, the collection can be re-added from the search',
```
dnahodil: I think 'removeButtonTooltip' is a better name for this dictionary entry because it tells you *how* this value is being used (in a tooltip).

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -137,45 +138,42 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
     },
 
-    _getButtonId: function(collection, i18Name) {
+    _getlinkId: function(collection, i18Name) {
         return OpenLayers.i18n(i18Name, {id: collection.uuid});
     },
 
-    _removeButtonOnClick: function(button) {
-        var collectionId = this.getIdFromButtonContainerId(button, "removeButtonId");
+    _removeOnClick: function(containerId) {
```
dnahodil: This should probably be called _removeLinkOnClick because we refer to the element as this.removeLink in other places.

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -137,45 +138,42 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
-    getIdFromButtonContainerId: function(button, i18Name) {
-        var collectionId = button.container.id;
+    getIdFromButtonContainerId: function(containerId, i18Name) {
         var prefix = OpenLayers.i18n(i18Name, {id: ""});
-        return collectionId.replace(prefix, '');
+        return containerId.replace(prefix, '');
     },
 
-    _getButtonId: function(collection, i18Name) {
+    _getlinkId: function(collection, i18Name) {
```
dnahodil: This should have a capital 'l'

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -137,45 +138,42 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
-                }
+                title: OpenLayers.i18n("removeButtonTooltip"),
+                cls: 'x-tool-awesome fa fa-fw fa-close',
+                renderTo: elementId
             });
+            this.removeLink.on('click', function() {
+                this._removeOnClick(elementId);
+            }, this);
+
+
```
dnahodil: Unnecessary whitespace here

dnahodil:
The CSS changes in this PR have changed the height of the grey header bars in step 1. I'm not sure if that was intentional.

dnahodil:
The share URL on step 3 is no longer right-aligned.

dnahodil:
@pmbohm comments made

pmbohm:
@dnahodil I took another look at the CSS. Ideally i would like the grey headers to be identical. There is a small difference due to Step3 having the Bootstrap links. I can live with it at the moment.
There is changes coming next for the share link. Currently the share link messes with the download calculator layout that will fix itself when the share link is relocated.

dnahodil:
@pmbohm I don't have tooltips for the remove 'x' on step 3 or for the 'Clear and Reset' button (also on step 3). Perhaps Ext.ux.Hyperlink doesn't support the 'title' attribute?

We should either:
* Update Ext.ux.Hyperlink so it supports the 'title' attribute
* Add the tooltips in some other way
or
* Remove the `title: xxx` attributes from the config objects for those links

PR #1966, opened 2015-09-09
Sharebutton and remove button reworking

pmbohm:


pmbohm:
@dnahodil This is a reworking of the previous pull request on Step 3 changes plus share button re-working. I needed the changes all in one place for my own sanity and its probably easier to review / see the full picture

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -41,25 +44,48 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
     _getRecordTitle: function(values) {
         return values.title;
     },
 
-    _shareButtonMarkup: function(values) {
-        return String.format(
-            '      <span class="fa fa-fw fa-share-alt fa-lg"></span>' +
-            '      <input readonly onclick="this.focus(); this.select();" title="{0}" value="{1}/home?uuid={2}" />',
-            OpenLayers.i18n('shareButton'),
+    _shareButton: function(collection) {
```
dnahodil: I don't think this is a great name for a function. I know it's basically the same as `_downloadButton()` below but I think function names should have a verb in them...

(It's up to you because it does match code that's already here)

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -41,25 +44,48 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
-            values.uuid
+            collection.uuid
         );
+        var elementId = this._getLinkId(collection, 'shareButtonId');
+
+        if ( Ext.get(elementId)) {
+
+            Ext.fly(elementId).update("");
+
+            this.shareLink = new Ext.ux.Hyperlink({
```
dnahodil: I think this could just be a local variable rather than applying it to `this`. It's not used anywhere else that I can see.

web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -137,45 +163,40 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
...
-        var elementId = this._getButtonId(collection, 'removeButtonId');
+    _createRemoveLink: function(collection) {
+        var elementId = this._getLinkId(collection, 'removeButtonId');
         if (collection.menuItems.length > 0 && Ext.get(elementId)) {
 
             // remove old button
             Ext.fly(elementId).update("");
 
-            new Ext.Button({
+            this.removeLink = new Ext.ux.Hyperlink({
```
dnahodil: Same here, no need to assign it to `this`. Just a local variable should work.

dnahodil:
Mostly all good @pmbohm. Just a few comments and then it's good-to-go.

dnahodil:
I noticed the start of the word 'undefined' flicker up on the page when you remove a data collection. I reckon there's somewhere where we're not returning an empty string but we should be.

dnahodil:
Also, now that we have a popup for the share link we have some more room to play with and it might be worth adding a little sentence to explain what it is and how to use it. Could be the same as the tooltip text, I guess.

dnahodil:
Sorry, @pmbohm, I noticed a couple of other things this morning.

pmbohm:
@dnahodil I don't think its necessary to have additional text. The share button is now a ubiquitous and understood icon.  A nice too have is the text highlighted and or copied into the clipboard but i wont go there either.

dnahodil:
Ok, it's up to you

PR #1979, opened 2015-09-16
fixes #1964

### LINKED ISSUE #1964: Download Estimator - bug waiting to happen?
### To Reproduce
Look at the generated HTML source of a step 3 download estimator

### What happens
Looking at the generated source we see similar too:
```
<div id="downloadEst-undefined-1441684507291">.......
```
The function called to build the div id:
```
getIdElementName: function(uuid) {
        return String.format("downloadEst-{0}-{1}", uuid, this.initTimestampString);
    },
```

### What should happen
uuid should not be undefined or the getIdElementName function should be changed

pmbohm:


web-app/js/portal/cart/BaseInjector.js
```diff
@@ -66,6 +70,7 @@ Portal.cart.BaseInjector = Ext.extend(Object, {
 
     _addDownloadEstimate: function(dataCollection) {
 
+        dataCollection.uuid = this._getCollectionUuid(dataCollection);
```
danfruehauf: I don't think it's a good idea to mutate the `dataCollection` object here.

`dataCollection.uuid` should be set when the `dataCollection` object is first instantiated.

web-app/js/portal/cart/BaseInjector.js
```diff
@@ -25,6 +25,10 @@ Portal.cart.BaseInjector = Ext.extend(Object, {
         };
     },
 
+    _getCollectionUuid: function(collection) {
```
danfruehauf: And this too, it's not the right place for that.

pmbohm:
@danfruehauf Thanks i found the correct built in function to use

web-app/js/portal/cart/BaseInjector.js
```diff
@@ -66,6 +66,7 @@ Portal.cart.BaseInjector = Ext.extend(Object, {
 
     _addDownloadEstimate: function(dataCollection) {
 
+
```
danfruehauf: That newline can go.

danfruehauf: I'll do it.

danfruehauf:
Saw that, I'll merge after tests pass.

Issue #800, opened 2014-01-10
Time filter on step 2 must manage/display UTC time


ggalibert:
What to do :
Select any metadata record linked to a layer which has a time filter enabled (e.g. : 
ANMN National Reference Stations - Near-real-time meteorology, see filter screen shot attached : 
![portal_time_filter](https://f.cloud.github.com/assets/3498686/1885151/c2477150-79c1-11e3-8ada-337e83dc7ff9.png)
) on step 1. Use this time filter on step 2 (see example attached : 
![step2_time_filter](https://f.cloud.github.com/assets/3498686/1885153/e51999c4-79c1-11e3-85bb-73d72940a281.png)
) and go to step 3.

Expected behaviour :
The dates in the sql query and resulted csv file should match the dates selected in step 2.

Experienced behaviour :
The dates don't match and suggest that the dates in step 2 are managed/displayed in local time (whichever it is...) and then converted to UTC time to filter data. See attached screenshot of step 3 : 
![step3_time_filter](https://f.cloud.github.com/assets/3498686/1885164/8193b2da-79c2-11e3-865e-cb0fb3099f03.png)

PR #1980, opened 2015-09-16
fixes #1954

### LINKED ISSUE #1954: Time filtering for aggregatable collections uses "greater/less or equal"
App version: 4.12.1
            Build date: Tue Sep 01 15:30:10 AEST 2015
            Build number: 14
            Git ref: d6c409ad34d5a579c4508c81a0767f7f285fef3e

Same as https://github.com/aodn/aodn-portal/issues/1350

Steps to reproduce:

Add any ncWMS collection to the map such as an ACORN collection

Set the time filter in the subset tab to be the same date for start and end and then choose the end time to be 2 steps after the start ie if the start is 1200 and then there are 1230, 1300, 1330 select 1300.

Proceed to step 2

Click download and select URL

What should happen:

Get a file with 3 URLS

What does happen:

Get a file with only 1 URL

pmbohm:


pmbohm:
fixes #1954

web-app/js/portal/filter/DateFilter.js
```diff
@@ -123,17 +122,14 @@ Portal.filter.DateFilter = Ext.extend(Portal.filter.Filter, {
     _getDateString: function(newDate) {
 
         if (newDate) {
-            newDate.setHours(this.timezoneCorrection);
```
danfruehauf: Not entirely sure about just removing that. Can you tell why this was added in the first place?

pmbohm: No i cant. Someone like @mhidas or @ggalibert would be good at testing this. But they would need it merged to test it? Conundrum....

danfruehauf: From memory (didn't use `git blame`) @dnahodil added this and he had a pretty solid reason. Simply removing that might introduce the problem he was trying to fix at the time.

pmbohm: Got the pull request where this was added. Now to analyse it
https://github.com/aodn/aodn-portal/pull/850

pmbohm: ### To test what is happening

Using the workflow that was given for bug #800

1) Select collection IMOS - ANMN National Reference Stations - Maria and North Stradbroke Islands - Near real-time meteorology and sea surface temperature
2) On Step 2, select 2012/11/16 to 2012/11/17 in temporal extent

### What Happens - On https://github.com/aodn/aodn-portal/pull/1980 pull request

* Step 3 shows:

```
Temporal Extent: 2012/Nov/16-00:00-UTC to 2012/Nov/17-00:00-UTC
```

* Downloaded data as CSV has timestamps from '2012-11-15T13:06:49' to '2012-11-16T12:06:50'


=========================================
On Production

* Step 3 shows:

```
Temporal Extent: 2012/Nov/16-10:00-UTC to 2012/Nov/17-10:00-UTC
```

*This is incorrect. The user would expect that they are picking from the start of 2012/Nov/16 to the start of 2012/Nov/17 all in UTC*


* Downloaded data as CSV has timestamps from  '2012-11-15T23:06:20' to '2012-11-16T22:06:49'

pmbohm:
After way too long confusing myself with the output of console.log's, I'm happy that this pull request is doing the right thing. The screen shot shows actual downloads matching the dates that a user sees throughout the Steps.

http://127.0.0.1:8080/aodn-portal/home?uuid=028b9801-279f-427c-964b-0ffcdf310b59&uuid=4402cb50-e20a-44ee-93e6-4728259250d2

screen shot 2015-09-21 at 11 35 03 am

pmbohm:
@danfruehauf Ive checked this again. As far as i can see its good. New push removes a required line

danfruehauf:
@pmbohm This looks more like a proper fix, I think. How should I test it?

pmbohm:
@danfruehauf Ive given an example above with a link to layers I used. Try similar layers, a WMS and a ncWMS. Basically the user should see date values remain consistant from Step 2 to download.
Dont try printing out to console date values as i did for a while as the browser will play tricks on you like converting to localtime.

danfruehauf:
From the best of my knowledge, this is a good fix for the problem. I'll merge it, but also tag @dnahodil so when he comes back he can also comment on that.

PR #1981, opened 2015-09-16
fixes #1977

### LINKED ISSUE #1977: Cannot tell step buttons are clickable
App version: 4.14.0
            Build date: Tue Sep 15 12:12:50 AEST 2015
            Build number: 16
            Git ref: 46c86164f7408e5001457edf483241bad348f0a9
        -->

Steps to reproduce:

Add any collection and then go back to step 1 hover mouse over step 2 and step 3 bar at top

What should happen:

Mouse pointer turns into hand when hovering over the buttons to proceed to next step

What does happen:

It stays as a pointer

pmbohm:


danfruehauf:
Looks good, I'll just verify the behaviour.

PR #1982, opened 2015-09-17
fixes #1976

### LINKED ISSUE #1976: Satellite collections show layer names in get feature info popup
App version: 4.14.0
            Build date: Tue Sep 15 12:12:50 AEST 2015
            Build number: 16
            Git ref: 46c86164f7408e5001457edf483241bad348f0a9

This is also in production

Steps to reproduce

Narrow down results by selecting Satellite from the Platform facets.

Select any collection from the list

On step 2 click any point on the collection display on the map

WEhat should happen:

Tabs show collection name - "human readable"

What does happen:

They show layer name - I think

ie IMOS - OceanCurrent - Gridded sea level anomaly - Near real time shows as surface_geostrophic_sea_water_velocity(NRT)

pmbohm:


pmbohm:
When the WMS link is missing title then the collection title is used.

web-app/js/portal/data/LayerSelectionModel.js
```diff
@@ -57,7 +57,11 @@ Portal.data.LayerSelectionModel = Ext.extend(Ext.util.Observable, {
 
     // TODO: unit tests?
     _linkToOpenLayer: function(layerLink, dataCollection) {
-        var layerDisplayName = dataCollection.get('title');
+
+        var layerDisplayName;
+        if (layerLink.title == "") {
+            layerDisplayName = dataCollection.get('metadataRecord').get('title');
```
jkburges: Can be `layerDisplayName = dataCollection.getTitle();` I think.  Better, because we're not accessing the internals of `DataCollection`.

jkburges:
I'm still seeing the layer name (I think):

image

pmbohm:
I will take a look at that. If its correct then we want to strip underscores and title case it

pmbohm:
The Geonetwork titles needs cleaning up. I'll expect the unexpected.

```
<link>
GSLA_DM/surface_geostrophic_sea_water_velocity|surface_geostrophic_sea_water_velocity (DM)|http://ncwms.aodn.org.au/ncwms/wms|OGC:WMS-1.1.1-http-get-map|application/vnd.ogc.wms_xml
</link>
```

jkburges:
> The Geonetwork titles needs cleaning up. I'll expect the unexpected.

So, is this PR doing as it should (and there's just some content issues)?

web-app/js/portal/data/LayerSelectionModel.js
pmbohm: `layerDisplayName = dataCollection.getTitle();` works but my IDE didnt help me find it. You did

pmbohm:
Yes but i will add some cleaning because we can expect lazy configuration from AODN providers

jkburges:
> Yes but i will add some cleaning because we can expect lazy configuration from AODN providers

I tend to think that we *shouldn't* do any cleaning, otherwise we end up with mess in both geonetwork *and* our code.

What do you think? 

Also, @pblain?

pblain:
Add an additional check to the monitor (that @danfruehauf  has just finished). Identify and hide any data collections with "lazy" config - and notify the responsible party.  i.e. Treat bad config as a failing data collection.  The portal code can then remain clean.

pmbohm:
My cleanup is only for lazy config. It title cases and removes underscores from a title. 
I think Dan's checker should blacklist bad config, that is config that just wont work.
But I +1 a response from @pblain

pmbohm:
@jkburges Removed titlecase & underscore removal

PR #1984, opened 2015-09-21
Share link in a input field

pmbohm:


pmbohm:
The share link is now behaving as commonly expected

PR #1989, opened 2015-09-21
Added WPS Download handler

pmbohm:
Works with http://127.0.0.1:8080/aodn-portal/home?uuid=8c5f876e-7394-4641-b2d0-3f6c1664662f&name=TestAnmnTS
Needs `geonetwork.url = "http://catalogue-rc.aodn.org.au/geonetwork"` in config.groovy

### DIFF
--- a/web-app/js/portal/cart/WpsDownloadHandler.js
+++ b/web-app/js/portal/cart/WpsDownloadHandler.js
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2014 IMOS
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ *
+ */
+
+Ext.namespace('Portal.cart');
+
+Portal.cart.WpsDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
+
+    ASYNC_DOWNLOAD_URL: 'asyncDownload?aggregatorService=wps&',
+
+    getDownloadOptions: function() {
+
+        var downloadOptions = [];
+        if (this._hasRequiredInfo()) {
+
+            downloadOptions.push({
+                textKey: 'downloadAsWpsLabel',
+                handler: this._getUrlGeneratorFunction(),
+                handlerParams: {
+                    asyncDownload: true,
+                    collectEmailAddress: true,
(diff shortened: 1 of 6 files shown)


jkburges:
I have broken this PR with [my commit](https://github.com/aodn/aodn-portal/commit/c4d92003886cbcc69bb268e4c9745926ddcee32a).  I will fix it here, since to be fair, this PR was up first!

web-app/js/portal/cart/WpsDownloadHandler.js
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2014 IMOS
+ *
+ * The AODN/IMOS Portal is distributed under the terms of the GNU General Public License
+ *
+ */
+
+Ext.namespace('Portal.cart');
+
+Portal.cart.WpsDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
```
jkburges: This class is *very* similar to `GogoduckDownloadHandler`, apart from some small differences in labels, URL construction, and subset parameters.  But overall, the logic is the same.

Consider extracting out the common code (to a `Portal.cart.AsyncDownloadHandler` class, perhaps).

pmbohm: @jkburges Yes now that the smoke has cleared, I agree with extracting the common code but i couldnt finish all tests in time. Feel free to take a look at https://github.com/aodn/aodn-portal/issues/1997 and make any changes to get it in if you wish

danfruehauf: Don't bother too much extracting common code with that and the gogoduck handler - gogoduck handler should die sooner than later.

pmbohm: Well the choices are there now on a separate branches, so take your pick tomorrow with @jkburges . Get healthy!

jkburges:
The test record is in geonetwork-rc, not 123 - I have changed that in the OP.

jkburges:
I don't seem to have a download option which triggers WPS?

jkburges:
> I don't seem to have a download option which triggers WPS?

Take that back - yes I do.

PR #1990, opened 2015-09-22
Developer happiness

jkburges:


pmbohm:
Sounds like a set of nice things. I'll take a punt

jkburges:
Not that exciting.  I just wasn't getting logs or reloads - pretty painful when trying to write code!

grails-app/conf/Config.groovy
```diff
@@ -291,14 +291,10 @@ marvl {
 // WFS indexed file info
 indexedFile.fileSizeColumnName = "size"
 
-// Something else is adding an appender (what?!) - get rid of it, else we end up
-// with duplicate logs in production.
-org.apache.log4j.Logger.rootLogger.removeAllAppenders()
```
danfruehauf: This will actually still result in double logging in prod. I tested it at the time...

jkburges: You mean with appenders removed we were still getting double logs?  In any case, I just checked on edge and we are *not* getting double logs currently, whoot!

danfruehauf: Alright, then forget about my comment then :)

Because at the time I realy wanted to remove all that, but I noticed that it actually resulted in double log lines.

PR #1991, opened 2015-09-22
Wps service

jkburges:
This PR makes it possible for the portal to send WPS requests to a WPS enabled geoserver, with a request such as the following:

```
curl "http://localhost:8080/aodn-portal/asyncDownload?aggregatorService=wps&jobParams.namespace=imos&jobParams.typeName=anmn_ts_timeseries_data&jobParams.cqlFilter=INTERSECTS(geom%2CPOLYGON((113.33%20-34.09%2C113.33%20-30.98%2C131.11%20-30.98%2C131.11%20-34.09%2C113.33%20-34.09)))%20AND%20TIME%20%26gt%3B%3D%20%272009-01-13T23%3A00%3A00Z%27%20AND%20TIME%20%26lt%3B%3D%20%272009-02-01T00%3A00%3A00Z%27&server=http://geoserver-systest.aodn.org.au/geoserver/wps"
```

Note: this request format has changed slightly due to c4d92003886cbc -  specifically the extra level of hierarchy in the request `jobParams`.

### DIFF
--- a/grails-app/services/au/org/emii/portal/WpsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsService.groovy
@@ -1,0 +1,30 @@
+package au.org.emii.portal
+
+import grails.gsp.PageRenderer
+import groovyx.net.http.*
+
+import static groovyx.net.http.Method.POST
+import static groovyx.net.http.ContentType.XML
+
+class WpsService extends AsyncDownloadService {
+
+    def groovyPageRenderer
+
+    def getConnection(params) {
+        def conn = new HTTPBuilder(params.server)
+        conn.contentType = XML
+
+        return conn
+    }
+
+    def getBody(params) {
+        def body = groovyPageRenderer.render(template: '/wps/asyncRequest.xml', model: params)
+        log.debug("Request body: ${body}")
+
+        return body
+    }
+
+    def onResponseSuccess = { resp, xmlReader ->
+        return new groovy.xml.StreamingMarkupBuilder().bindNode(xmlReader).toString()
+    }
+}
--- a/grails-app/views/wps/_asyncRequest.xml.gsp
+++ b/grails-app/views/wps/_asyncRequest.xml.gsp
@@ -9,28 +9,16 @@
   <ows:Identifier>gs:NetcdfOutput</ows:Identifier>
 
   <wps:DataInputs>
-    <wps:Input>
-      <ows:Identifier>namespace</ows:Identifier>
-      <wps:Data>
-        <wps:LiteralData>imos</wps:LiteralData>
-      </wps:Data>
-    </wps:Input>
-    <wps:Input>
-      <ows:Identifier>typeName</ows:Identifier>
-      <wps:Data>
-        <wps:LiteralData>${typeName}</wps:LiteralData>
-      </wps:Data>
-    </wps:Input>
+    <g:each in="${jobParams}" var="id, value">
 
-    <g:if test="${cqlFilter}">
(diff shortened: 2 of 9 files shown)


pmbohm:
The `Portal.cart.WpsDownloadHandler`will send you also emailAddress and url. You would need these right?

jkburges:
> The Portal.cart.WpsDownloadHandlerwill send you also emailAddress and url. You would need these right?

Correct, not quite up to that bit yet :worried:

grails-app/views/wps/_asyncRequest.xml.gsp
```diff
@@ -0,0 +1,45 @@
...
+      </wps:Data>
+    </wps:Input>
+    <wps:Input>
+      <ows:Identifier>typeName</ows:Identifier>
+      <wps:Data>
+        <wps:LiteralData>${typeName}</wps:LiteralData>
+      </wps:Data>
+    </wps:Input>
+
+    <g:if test="${cqlFilter}">
```
danfruehauf: I thought here instead, we could iterate on the parameters that the user provided, apart from email, which is mandatory. So same goes to `typeName` and `namespace`. So the using controller passes to the view email + a map of parameters and the view encodes them one by one without asking questions. This makes the portal dumb about WPS - and it should be, with the logic and flexibility kept on JS side, which is where all the download logic resides.

jkburges: Yes, that's a good idea.  Nice.

danfruehauf:
@jkburges I think that's altogether a very clean and concise implementation. Good job!

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+import static groovyx.net.http.Method.POST
+import static groovyx.net.http.ContentType.XML
+
+class WpsService extends AsyncDownloadService {
+
+    def groovyPageRenderer
+
+    def getConnection(params) {
+        // TODO: get geoserver URL from request?
```
jkburges: Oops I need to remove this TODO comment.

grails-app/views/wps/_asyncRequest.xml.gsp
jkburges: Done.

danfruehauf:
GTG?

jkburges:
GTG :+1:

grails-app/views/wps/_asyncRequest.xml.gsp
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<wps:Execute service="WPS" version="1.0.0"
+  xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1"
+  xmlns:ogc="http://www.opengis.net/ogc" xmlns:xlink="http://www.w3.org/1999/xlink"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 http://schemas.opengis.net/wps/1.0.0/wpsExecute_request.xsd">
+
+
+  <ows:Identifier>gs:NetcdfOutput</ows:Identifier>
+
```
danfruehauf: I'm just noting that this view will have to have email as a parameter that is mandatory. Without an email, the job cannot be submitted.

PR #1995, opened 2015-09-24
Wps status page

jkburges:
Access the page with URL of the form:

http://localhost:8080/aodn-portal/wps/jobReport?server=\<WPS server\>&uuid=\<UUID\>

e.g.:

http://localhost:8080/aodn-portal/wps/jobReport?server=http%3A%2F%2Fgeoserver-systest.aodn.org.au%2Fgeoserver%2Fows&uuid=e8c4ce84-3e18-4c37-9790-746dcf2c79cb

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -137,6 +137,8 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
+
+        grails.mail.port = com.icegreen.greenmail.util.ServerSetupTest.SMTP.port
     }
 
     test {
@@ -150,6 +152,8 @@ environments {
         // URLs
         grails.serverURL = "http://myaodn.example.com"
         geonetwork.url = "http://catalogue-123.aodn.org.au/geonetwork"
+
+        greenmail.disabled = true
     }
 }
 
@@ -316,7 +320,8 @@ log4j.main = {
 
     debug   'grails.app.job',
         'grails.app.tagLib',
-        'grails.app.controller.au.org.emii.portal.SystemController'
+        'grails.app.controller.au.org.emii.portal.SystemController',
+        'grails.plugin.mail'
 }
 
 /**
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -126,6 +126,12 @@ enabledFacets = [
 // Google Analytics
 googleAnalytics.trackingId = null
 
+// Let's be careful about not being open to spamming.
+// This must be specifically enabled in order for emails to be sent from the app.
+// This can be changed once we have some other form of protection in place (e.g. apache locking
+// down which IP address are allowed to access certain email-producing URLs).
+grails.mail.disabled = true
+
 environments {
 
     development {
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -143,8 +143,6 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
-
-        grails.mail.port = com.icegreen.greenmail.util.ServerSetupTest.SMTP.port
     }
 
     test {
@@ -158,8 +156,6 @@ environments {
         // URLs
         grails.serverURL = "http://myaodn.example.com"
         geonetwork.url = "http://catalogue-123.aodn.org.au/geonetwork"
-
-        greenmail.disabled = true
     }
 }
 
--- a/grails-app/services/au/org/emii/portal/WpsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsService.groovy
@@ -2,6 +2,7 @@ package au.org.emii.portal
 
 import grails.gsp.PageRenderer
 import groovyx.net.http.*
+import org.codehaus.groovy.grails.web.mapping.LinkGenerator
 
 import static groovyx.net.http.Method.POST
 import static groovyx.net.http.ContentType.XML
@@ -9,6 +10,15 @@ import static groovyx.net.http.ContentType.XML
 class WpsService extends AsyncDownloadService {
 
     def groovyPageRenderer
+    LinkGenerator grailsLinkGenerator
+
+    String registerJob(params) throws HttpResponseException {
+        def registerResponse = super.registerJob(params)
+        params.uuid = _getUuidFromRegisterResponse(registerResponse)
+        _notifyRegistrationViaEmail(params)
+
+        return registerResponse
+    }
 
     def getConnection(params) {
         def conn = new HTTPBuilder(params.server)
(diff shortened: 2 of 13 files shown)


jkburges:
Some work which can possibly follow on from this is to show a 'report' field, which would contain more detailed information taken from the status element in the WPS response.

However, each element has a different structure depending on the status, so it's not completely trivial.  Hence, it's not been done (yet).

Anyway, this at least gives us enough to go on with for the full portal - WPS integration.

grails-app/views/wps/show.gsp
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <div class="jumbotronFooter voffset5">
+            <div class="container">
+                <footer class="row">
+                    <div class="col-md-4">
+                        <p>If you've found this information useful, see something wrong, or have a suggestion,
+                           please let us
+                           know.
+                           All feedback is very welcome. For help and information about this site
+                           please contact <a href="mailt:info@emii.org.au">info@emii.org.au</a></p>
```
pmbohm: mailto:

jkburges: Nice pick up - this bug exists in gogoduck also!

grails-app/views/wps/show.gsp
```diff
@@ -0,0 +1,51 @@
...
+                <a class="btn" role="button" href="https://imos.aodn.org.au/imos123/home">
+                    <img src="https://static.emii.org.au/images/logo/IMOS-Ocean-Portal-logo.png" alt="IMOS logo">
+                </a>
+            </div>
+        </div>
+        <div class="container">
+            <h2><g:message code="job.status" default="WPS Status"/></h2>
+            <dl>
+                <g:labelledContent labelCode="job.id.label">${job.uuid}</g:labelledContent>
+                <g:labelledContent labelCode="job.createdTimestamp.label" if="${job.createdTimestamp}" >
```
pmbohm: If the job is finished then I wouldn't want to see 'Submitted:' header at all


screen shot 2015-09-24 at 1 54 56 pm

jkburges: That's a bug.  It should be showing the date of submission.

c.f.: https://gogoduck.aodn.org.au/gogoduck/job/f61e4003

grails-app/views/wps/show.gsp
```diff
@@ -0,0 +1,51 @@
...
+                  <joda:format value="${job.createdTimestamp}" />
+                </g:labelledContent>
+                <g:labelledContent labelCode="job.status.label">
+                  <g:message code="job.status.${job.status}" default="${job.status.toString()}" />
```
pmbohm: I think if the job isn't completed it should say for example

**Status:** Submitted at 19/02/2015 3:30pm

grails-app/views/wps/show.gsp
```diff
@@ -0,0 +1,51 @@
...
+        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css">
+        <link rel="stylesheet" href="${resource(dir: 'css', file: 'imosTheme.css')}" type="text/css">
+    </head>
+    <body>
+        <div class="imosHeader">
+            <div class="container">
+                <a class="btn" role="button" href="https://imos.aodn.org.au/imos123/home">
+                    <img src="https://static.emii.org.au/images/logo/IMOS-Ocean-Portal-logo.png" alt="IMOS logo">
+                </a>
+            </div>
```
pmbohm: Should we have links to the portal? or IMOS?
Sorry, this has probably been discussed at length with Gogoduck

jkburges: My brief, in this respect, was to "just do what GoGoDuck does" - so, better probably to ask @pblain!

grails-app/views/wps/show.gsp
jkburges: As above, "just do what gogoduck does"!

grails-app/views/wps/show.gsp
pmbohm: Ok so that makes sense if it always shows the submission time. That's useful even

jkburges: Hmm I can't recreate this.  Seems to work for me both for jobs which are 'started' and 'succeeded'.

grails-app/conf/Config.groovy
```diff
@@ -137,6 +137,8 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
+
+        grails.mail.port = com.icegreen.greenmail.util.ServerSetupTest.SMTP.port
```
danfruehauf: What's greenmail?

We might use here AWS SES

jkburges: See https://grails.org/plugin/greenmail

Basically, enables one to test sending emails without having to rely on an external SMTP server.

danfruehauf: Before we roll this, remind me to subscribe for SES, so we can use it to send emails and drive this config from chef.

If greenmail is only for dev, then that's all good.

grails-app/views/wps/jobComplete.gsp
```diff
@@ -0,0 +1,11 @@
+Hi,
+
+Your IMOS download is ready. You can download it by clicking here: ${jobReportUrl}
```
danfruehauf: This may need to change. That is because you can't tell the user the data is ready as it might have failed. So need to tell the user the job has completed and she can click the URL to get more information and potentially also download the data.

grails-app/conf/Config.groovy
jkburges: FWIW, I've also been testing using the same email config as with gogoduck (i.e. sendmail) and it works fine with that.  So we can use the same chef code for this (SES aside).

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -27,4 +37,36 @@ class WpsService extends AsyncDownloadService {
     def onResponseSuccess = { resp, xmlReader ->
         return new groovy.xml.StreamingMarkupBuilder().bindNode(xmlReader).toString()
     }
+
+    def notifyViaEmail(params) {
```
danfruehauf: If I merge this, I need to make sure that either:
 * This action is mocked (i.e., doesn't send email)
 * Protected - cannot be accessed except by authorized computers

Alternatively we can protect it with apache, but we have to remember to do so...

grails-app/conf/Config.groovy
danfruehauf: I want SES because it'll reduce by 1 the amount of services we depend on...

But yes, we can also start with sendgrid. Essentially it's just a simple data bag.

grails-app/services/au/org/emii/portal/WpsService.groovy
danfruehauf: Scratch that comment from here, look at `jobComplete` in the controller, but same comment.

grails-app/views/wps/jobComplete.gsp
jkburges: Yeah it does need to change, you're right.  I thought of that but forgot.  @pblain, what do you want the wording to be (this is the email that users get when the job has completed - be it successfully or not).

The other option is for the portal server to interrogate geoserver to find the status, but personally I would advise against this, at least for now, because it adds another point of failure (for a small amount of benefit).

danfruehauf: I'm with you on that. I advised against it also in the original design, as you said, due to added complexity with not much added value for the user.

grails-app/services/au/org/emii/portal/WpsService.groovy
jkburges: How's about I have the default email destination be the greenmail "server" - i.e. emails will not be sent.  chef can override this with a real SMTP destination (just as it already does with ggd) when the security is ready.  

My feel is that it would be better to do this with apache (or something else off-the-shelf) rather than code our own logic - but you may have a different opinion on this?

jkburges:
@danfruehauf have disabled email by default. Also, decided not to include greenmail, easy enough to test with a debug log or using a real SMTP server (see the comment in 5728470).

So, we can either wait for @pblain's advice on email wording changes, or merge and do that later (my preference, especially that this feature isn't "live" on the front-end yet).

danfruehauf:
@jkburges Lets get the wording right, so we don't need to revisit it. I'm going to talk now to @pblain.

grails-app/views/wps/jobComplete.gsp
pblain: @danfruehauf , the text is pretty much OK.  Perhaps change slightly to read:
Your IMOS subset has been created. You can download it by clicking here:

jkburges: @pblain, this wording (or the original here) doesn't really make sense if the subset failed (because there will be nothing to download).

It needs to be something more like:

> Your IMOS subset has completed.  Access the results here:

or something better, but that leaves open the possibility of failure...

But if you understand all that and https://github.com/aodn/aodn-portal/pull/1995/files#r40636889 is definitely what you want, no problems!

danfruehauf:
We can fix the wording later if it's a problem. Better have that in to avoid painful rebases.

grails-app/views/wps/jobComplete.gsp
pblain: Obfuscating the text will not soften the blow

PR #1996, opened 2015-09-24
Support collection availability in Geonetwork

danfruehauf:
* Do not pass any `filters=` values by default, for backward
   compatibility
 * Pass `filters=collectionavaiability` if passed `healthy=true`
   this will show all healthy layers
 * Pass `filters=!collectionavaiability` if passed `healthy=false`
   this will show all broken layers

### DIFF
--- a/web-app/js/portal/service/CatalogSearcher.js
+++ b/web-app/js/portal/service/CatalogSearcher.js
@@ -212,6 +212,21 @@ Portal.service.CatalogSearcher = Ext.extend(Ext.util.Observable, {
         return deepestFacets;
     },
 
+    _getCollectionAvailabilityParams: function(url) {
+        var getParams = url.split("#")[0].split("?");
+        var params = Ext.urlDecode(getParams[1]);
+
+        if (params.healthy === "true") {
+            return { filters: "collectionavailability" };
+        }
+        else if (params.healthy === "false") {
+            return { filters: "!collectionavailability" };
+        }
+        else {
+            return {};
+        }
+    },
+
     _getParams: function(page) {
         //--- Add search defaults (e.g. map layers only)
         var params = Ext.apply({}, this.defaultParams);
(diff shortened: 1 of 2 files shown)


danfruehauf:
Merge only after geonetwork link monitoring is in too, otherwise things will break.

Generally speaking I think the logic of assembling the search query (or at least most of its parameters) should move to the server side, but I don't have time to do it now.

The nice things about this is that we will be able to find all the broken collections by navigating to:
https://imos.aodn.org.au/imos123/home?broken=true

danfruehauf:
Can be reviewed. GN is being deployed to systest as we speak...

Oh, but edge?! What will we do with edge???

jkburges:
> Oh, but edge?! What will we do with edge???

Perhaps we need to have toggling for this feature?

web-app/js/portal/service/CatalogSearcher.js
```diff
@@ -212,6 +212,21 @@ Portal.service.CatalogSearcher = Ext.extend(Ext.util.Observable, {
         return deepestFacets;
     },
 
+    _brokenMode: function(url) {
+        var getParams = url.split("#")[0].split("?");
```
jkburges: http://extjs.cachefly.net/ext-3.3.1/docs/source/Ext.html#method-Ext-urlDecode ?

danfruehauf:
I'm working on a feature toggle...

web-app/js/portal/service/CatalogSearcher.js
danfruehauf: I don't think you should `urlDecode` before splitting the `#` and `?`. I took some code from:
https://github.com/aodn/aodn-portal/blob/master/web-app/js/portal/data/MetadataRecordFetcher.js#L54

jkburges: Yeah, you're right.

danfruehauf:
This is not a feature toggle, but it is backward compatible. Now we can continue browsing as usual but have the option of passing either `healthy=true` or `healthy=false`.

jkburges:
Want to raise an issue/backlog item for this somewhere so that we can change it over (to be default behaviour) once all geonetworks support it?

PR #2000, opened 2015-09-30
Fix #1998 - use correct uuid

### LINKED ISSUE #1998: Download Estimator - not completing
### Whats Happening (current Master)
On http://portal-edge.aodn.org.au/imos/home?uuid=4402cb50-e20a-44ee-93e6-4728259250d2&health=all the Download Estimator on Step 3 never seems to populate. 
### To reproduce
1. select a small area on Step 2
2. Note the request to ....aodn-portal/download/estimateSizeForLayer is made and returned with a numeric value not an error message or status
3. On Step 3 the estimator status is never shown

danfruehauf:
Fix download estimator - can now find element to update after
download estimator returns

### DIFF
--- a/src/test/javascript/portal/cart/DownloadPanelItemTemplateSpec.js
+++ b/src/test/javascript/portal/cart/DownloadPanelItemTemplateSpec.js
@@ -21,7 +21,7 @@ describe('Portal.cart.DownloadPanelItemTemplate', function () {
         Portal.app.appConfig.grails = {serverURL: "munt"};
 
         mockDataInjection = {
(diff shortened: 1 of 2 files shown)


danfruehauf:
I'd like to hotfix systest with that, so we can release.

pmbohm:
@danfruehauf I get `Uncaught TypeError: collection.getUuid is not a function`
Added Argo layer, after a pull and server restart on Chrome

danfruehauf:
I assumed a few wrong things. Works now.

PR #2002, opened 2015-09-30
Correct additional links now appear in Step 3

pmbohm:


pmbohm:
Compare Argo on production vs master to test. They should now be the same.
Not many collection have any other links appearing apart from metadata on Step 3

jkburges:
If I understand this correctly, we now use something else for these "deprecated" link types, am I right?  Is there anything in the backlog or wherever to update the affected records with whatever types we are using now?

Issue #1988, opened 2015-09-21
Date range on Step 2 not working


pblain:
## Steps to reproduce
1. Select AusCPR and go to Step 2
2. Enter the date range 2010/01/11 in both the "from" and "to" fields

## What happens 
The map is empty

## What I expect to happen
Data from the entire day (of the 11th) displayed on the map - not just data at 00-UTC

PR #2004, opened 2015-10-01
WmsDateFilter now adds 24 hours to end date

pmbohm:


pmbohm:
Fixes #1988

web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -138,7 +138,7 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Ext.Container, {
 
         var panelOrder = [
             Portal.filter.GeometryFilter,
-            Portal.filter.DateFilter,
```
jkburges: Any reason why we need to retain `DateFilter`, along with `WmsDateFilter`?  Can it not just be rolled in to the same class, if it's always the same behaviour that we need?

src/test/javascript/portal/filter/DateFilterSpec.js
```diff
@@ -55,11 +55,6 @@ describe("Portal.filter.DateFilter", function() {
             expect(filter.getCql()).toBe("column_name >= '1999-01-01T00:00:00Z'");
         });
 
-        it('gives data layer CQL', function() {
```
jkburges: These are fairly important unit tests which no longer appeared to be covered anywhere (as far as I can see)?

jkburges:
I'm not sure if I understand this PR properly.  My thought is that is we just update the existing `DateFilter` class to do what we want (i.e. include the whole day of the end date), then the changes should be pretty minimal.

e.g. the deleted unit tests can stay, the only change really would need to be a line of code that adds 24 hrs to the `to date`?  But maybe I'm missing something, it is Monday morning after all.

pmbohm:
For ncWMS we dont want to do anything with the end date as they contain times. 
ncWMS also uses `DateFilter` so that requires the new class @jkburges

src/test/javascript/portal/filter/DateFilterSpec.js
pmbohm: I set out to leave all the same existing tests in `WmsDateFilter` but could not work out a way to get the resulting call to `getEndOfToDate` to work

dnahodil:
This filtering code should really be moving towards generic code for everything. That's been something we've been trying to do for ages. The ncWMS using DateFilter is a temporary workaround to allow us to transition towards unified filtering. If anything needs to change it should be that workaround. I can give you a hand to look at it if you like.

dnahodil:
If we're happy to lose a second we can just set the time component of the end date date picker to 11:59.59. Then all the other code should work unchanged.

pmbohm:
@dnahodil Well i dont want to go against any grand plan and I dont really want another week off so if you would like to close this pull request and put up an alternative, please do.

PR #2011, opened 2015-10-05
fixes #2006

### LINKED ISSUE #2006: If no data to download then collection does not have remove function on step 3
App version: 4.15.3
            Build number: 20
        -->
It has been like this for a very very long time just hadn't reported it

Steps to reproduce:

Add a collection that will not have data to download such as IMOS - AATAMS Facility - Acoustic Tagging Data Repository or IMOS - ANMN Passive Acoustic Observatories Sub-facility. Proceed to step 3

What should happen:

Once I realise I cannot download any data I should be able to remove the collection on step 3 like any other collection

What does happen:

There is no remove button, so as a user i have to inconveniently go back to step 2 to remove it.

pmbohm:


dnahodil:
The fix works well, @pmbohm.

Just one question, though. Why are we getting rid of that test?

pmbohm:
`_createRemoveButtonAfterPageLoad` always returned the same expected result.
`describe('apply',.....` seems to have it covered

dnahodil:
Ok then. It'd be nice to mention that in the commit message so it's clear that it wasn't removed by accident.

PR #2014, opened 2015-10-05
Fixes #2007

### LINKED ISSUE #2007: Clear subset does not reset display time for ncWMS
App version: 4.15.3
            Build number: 20
Steps to reproduce:

Add an ACORN collection.

Set start date to be something different to default.

Note displaying date/time is updated.

Click clear subset

What should happen:

Display date/time is reset to match the default date/time that just reset

What does happen:

Date/time selected are reset but display time is still the previous date/time that was selected.

pmbohm:


dnahodil:
This change resets the date as requested, however the map layer doesn't appear to update to that date (i.e. the map still shows the data from the previously-selected date)

pmbohm:
@dnahodil Now calling the pre-existing and now more intuitively named function that fixes everything

dnahodil:
Looks good @pmbohm

PR #2016, opened 2015-10-05
Issue #1988 alternative

### LINKED ISSUE #1988: Date range on Step 2 not working

dnahodil:
This is another way to fix #1988

### DIFF
--- a/web-app/js/portal/filter/ui/DateFilterPanel.js
+++ b/web-app/js/portal/filter/ui/DateFilterPanel.js
@@ -101,7 +101,13 @@ Portal.filter.ui.DateFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel, 
 
         this.filter.setValue({
             fromDate: this.fromDate.getValue(),
-            toDate: this.toDate.getValue()
+            toDate: this._endOfDay(this.toDate.getValue())
         });
+    },
+
+    _endOfDay: function(date) {
+        if (date) {
+            return moment(date).endOf('day').toDate();
+        }
     }
 });
--- a/src/test/javascript/portal/filter/ui/DateFilterPanelSpec.js
+++ b/src/test/javascript/portal/filter/ui/DateFilterPanelSpec.js
@@ -13,8 +13,6 @@ describe("Portal.filter.ui.DateFilterPanel", function() {
     beforeEach(function() {
         Portal.filter.ui.DateFilterPanel.prototype._createControls = noOp;
 
-        Portal.filter.ui.DateFilterPanel.prototype._getDateString = noOp;
-
         filterPanel = new Portal.filter.ui.DateFilterPanel({
             filter: {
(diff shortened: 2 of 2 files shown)


dnahodil:
@pmbohm this is the sort of thing I was thinking in my comment (https://github.com/aodn/aodn-portal/pull/2004#issuecomment-145403435).

web-app/js/portal/filter/ui/DateFilterPanel.js
```diff
@@ -101,7 +101,14 @@ Portal.filter.ui.DateFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel,
 
         this.filter.setValue({
             fromDate: this.fromDate.getValue(),
-            toDate: this.toDate.getValue()
+            toDate: this._endOfDay(this.toDate.getValue())
         });
+    },
+
+    _endOfDay: function(date) {
```
jkburges: Can use http://momentjs.com/docs/#/manipulating/end-of/ ?

Except this is a javascript date, not a moment. But maybe we can still use it?

pmbohm: I tried that function. The results were not really what is wanted as i dont think it handles the fact that we want the last moment of a UTC day not the last second of the date in client timezone expressed as UTC

pmbohm: I am a fan of using moment though
```
        var momentDate = moment.utc(date);
        if (date && momentDate.isValid()) {
            return momentDate.add(1, 'd').toDate();
        }
```

web-app/js/portal/filter/ui/DateFilterPanel.js
```diff
@@ -101,7 +101,13 @@ Portal.filter.ui.DateFilterPanel = Ext.extend(Portal.filter.ui.BaseFilterPanel,
...
+        if (date) {
+            return moment(date).endOf('day').toDate();
```
dnahodil: How's that? or is it too much for one line?

dnahodil:
@jkburges changes made for suggestion

web-app/js/portal/filter/ui/DateFilterPanel.js
jkburges: Is this going to work w.r.t. https://github.com/aodn/aodn-portal/pull/2016/files#r41110785 ?

dnahodil: It seemed to work as I expected it to. I did a test download on the Argo collection and the downloaded data was all within the correct day.

PR #2017, opened 2015-10-05
Fixes #1999

### LINKED ISSUE #1999: Select button does not take user to step 2 after ncwms collection added
pp version: 4.15.3
            Build number: 20
        -->

**Steps to reproduce:**

1. Refresh portal
2. proceed to bottom of first page of results and type 12 in the page number box
3. Add the first record "IMOS - SRS SATELLITE - SST L3S - 01 day composite - day and night time composite" (goes to step 2)
4. (Attempt to) add the second record "IMOS - SRS SATELLITE - SST L3S - 03 day composite - day time".

**What should happen:**

Taken to step 2

**What does happen:**

Select button greys - collection is added but user is kept on step 1 like ctrl is being held down.

jkburges:
[Alt text for your video](http://www.youtube.com/watch?v=n1EymCZSqFg)

Don't attempt to load filters from the server for NcWms - currently, they are hard-coded in the client and trying to load from server just wipes out the desired filters with an empty list.  This in turn causes problems such as #1999.

Note that the proper fix is to treat NcWms collections like any other, i.e.:

* load filters from server
* model filters in exactly the same way as non-ncwms
* use the same UI components as non-ncwms

This can be reverted when *all* collections, [NcWms and non-NcWms, are treated the same as far as filtering is concerned](https://github.com/aodn/backlog/issues/191).

### DIFF
--- a/src/test/javascript/portal/data/DataCollectionSpec.js
+++ b/src/test/javascript/portal/data/DataCollectionSpec.js
@@ -8,9 +8,11 @@
 describe("Portal.data.DataCollection", function() {
 
     var dataCollection;
+    var isNcwmsSpy;
 
     beforeEach(function() {
         spyOn(Portal.data.DataCollection.prototype, '_loadFilters');
+        isNcwmsSpy = spyOn(Portal.data.DataCollection.prototype, 'isNcwms').andReturn(true);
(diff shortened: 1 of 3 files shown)


dnahodil:
I guess this is okay until we go through with the proper fix.

Probably worth writing a test, though.

PR #2023, opened 2015-10-06
WPS jobComplete email wording change

pmbohm:


grails-app/views/wps/jobComplete.gsp
```diff
@@ -1,6 +1,8 @@
 Hi,
 
-Your IMOS download is ready. You can download it by clicking here: ${jobReportUrl}
+Your IMOS subset has been created. You can download it by clicking here: ${jobReportUrl}
+
+Note: The link will only be available for the next 24 hours.
```
jkburges: Not sure that I explained myself properly before, but the "24" should be coming from `Config.groovy`.  This will then make it easy for chef to overwrite (you don't have to worry about *that* for now, but it'd be nice if this was configurable to begin with).

It's called `wps.resourceExpirationTimeout` in geoserver, not sure if that's descriptive enough.

pom.xml
```diff
@@ -233,6 +233,12 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
```
jkburges: Not sure how this was tested, but `grails run-app` is now broken (because the dependency is present only for maven).  Happily, it seems as though [there is a prettytime plugin](http://grails.org/plugin/pretty-time), so should be easy enough to pull it in that way (in `BuildConfig.groovy`).

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -39,10 +43,12 @@ class WpsService extends AsyncDownloadService {
     }
 
     def notifyViaEmail(params) {
+
+        params.expirationPeriod = _getExpirationPeriod()
```
jkburges: If we make sure that there's always a value set in the config, and using joda, then I think this could be as simple as:

```groovy
params.expirationDateTime = new DateTime().plusSeconds(grailsApp.config.wpsResourceExpirationTimeout)
```

Then in the gsp:

```xml
... <prettytime:display date="${expirationDateTime}" />.
```

Not sure about the need for unit exclusion as in https://github.com/aodn/aodn-portal/pull/2023/files#diff-cf1774ac4be79e524ea06d4aaffd87efR72 - I would've thought this library would pick something sensible?

pom.xml
jkburges: `prettytime` looks like a pretty cool library, btw.

grails-app/services/au/org/emii/portal/WpsService.groovy
pmbohm: I didnt try the tag in the GSP. I didnt even know about it. However the simple pretty time format function didnt give me an accurate enough time which i assume the tag would do. 

The `_getExpirationPeriod` function is now quite bullet proof which im happy about. I'd also rather have it so config can be forgotten or left off by us or others.

pom.xml
pmbohm: 50% chance of getting this wrong :(

pmbohm:
@jkburges Good to go now

PR #2026, opened 2015-10-09
Use regex to match url replacement

danfruehauf:


danfruehauf:
@dnahodil Seems to work OK (tested with Argo).

PR #2027, opened 2015-10-13
Added to metadataProtocols in config.groovy

pmbohm:
@atkinsn discovered an additional protocol is being used by metadata

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -263,7 +263,9 @@ portal {
 
         supplementary = [
             "WWW:LINK-1.0-http--link",
-            "WWW:LINK-1.0-http--downloaddata"
+            "WWW:LINK-1.0-http--downloaddata",
+            "WWW:LINK-1.0-http--related",
+            "WWW:DOWNLOAD-1.0-ftp--download"
         ]
 
         metadataRecord = [


dnahodil:
@atkinsn could you please give an example of a data collection which uses this protocol?

atkinsn:
Uuid: 8af21108-c535-43bf-8dab-c1f45a26088c

grails-app/conf/Config.groovy
```diff
@@ -263,7 +263,8 @@ portal {
 
         supplementary = [
             "WWW:LINK-1.0-http--link",
-            "WWW:LINK-1.0-http--downloaddata"
+            "WWW:LINK-1.0-http--downloaddata",
+            "WWW:LINK-1.0-http—related"
```
dnahodil: This looks like it might be the wrong hyphen/dash character @pmbohm?

dnahodil:
Thanks, Nat

grails-app/conf/Config.groovy
dnahodil: Yeah, this seems to be incorrect. Perhaps a text editor somewhere along the way turned a double hyphen into a dash of some kind.

pmbohm:
@dnahodil additional additional protocol now added

PR #2029, opened 2015-10-14
Tiny modifications required to work with ncwms plugin

danfruehauf:
* Add Z when requesting time to denote UTC (thredds need that)
 * Pass service=ncwms, so Geoserver uses ncwms plugin instead of
   the default wms plugin. Doesn't affect regular ncwms
   interaction

### DIFF
--- a/web-app/js/portal/ui/openlayers/layer/NcWms.js
+++ b/web-app/js/portal/ui/openlayers/layer/NcWms.js
@@ -21,6 +21,7 @@ OpenLayers.Layer.NcWms = OpenLayers.Class(OpenLayers.Layer.WMS, {
 
         this.pendingRequests = new Portal.utils.Set();
 
+        params['SERVICE'] = "ncwms";
         OpenLayers.Layer.WMS.prototype.initialize.apply(this, [name, url, params, options]);
 
         this._setExtraLayerInfoFromNcwms();
@@ -273,7 +274,7 @@ OpenLayers.Layer.NcWms = OpenLayers.Class(OpenLayers.Layer.WMS, {
     },
 
     _getTimeParameter: function(dateTime) {
-        return dateTime.clone().utc().format('YYYY-MM-DDTHH:mm:ss.SSS');
+        return dateTime.clone().utc().format('YYYY-MM-DDTHH:mm:ss.SSS[Z]');
     },
 
     _getExtraLayerInfoFromNcwms: function() {
--- a/src/test/javascript/portal/ui/openlayers/layer/NcWmsSpec.js
+++ b/src/test/javascript/portal/ui/openlayers/layer/NcWmsSpec.js
@@ -44,7 +44,7 @@ describe("OpenLayers.Layer.NcWms", function() {
 
         it('time specified', function() {
             cachedLayer.setTime(time);
-            expect(cachedLayer.params['TIME']).toBe(time.utc().format('YYYY-MM-DDTHH:mm:ss.SSS'));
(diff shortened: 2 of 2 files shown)


danfruehauf:
These are *all* the portal changes required to work with the abomination ncwms plugin.

web-app/js/portal/ui/openlayers/layer/NcWms.js
```diff
@@ -278,7 +279,7 @@ OpenLayers.Layer.NcWms = OpenLayers.Class(OpenLayers.Layer.WMS, {
     },
 
     _getTimeParameter: function(dateTime) {
-        return dateTime.clone().utc().format('YYYY-MM-DDTHH:mm:ss.SSS');
+        return dateTime.clone().utc().format('YYYY-MM-DDTHH:mm:ss.SSS') + 'Z';
```
dnahodil: The `Z` character can just be part of the format string, no need to append

dnahodil:
Are these changes backwards-compatible?

dnahodil:
Also, broken tests need to be fixed, obviously.

danfruehauf:
Tests needs to be fixed, true. I overlooked it...

And yes, tested in front of the original ncwms it was all good.

BTW, no need to merge it right now, it's here so @pmbohm can pick it up and see what was needed to be modified in portal...

dnahodil:
> tested in front of the original ncwms it was all good.
:+1:

> no need to merge it right now
If it's all the changes required then is there any reason not to merge it?

danfruehauf:
The only reason I can think of is if the plugin in geoserver will not be called `ncwms`. I was thinking more along the lines of `abomination-ncwms` or something like that ;)

dnahodil:
Ok, I'll hold off

danfruehauf:
As a sidenote I'll add that having the extra `Z` is the correct way. The 'Z' is to denote UTC, correct me if I'm wrong.

dnahodil:
> ... having the extra Z is the correct way. The 'Z' is to denote UTC ...

Yeah, that's my understanding

pmbohm:
As its a Geoserver plugin, calling the service ncwms should be enough

dnahodil:
Ok. Happy to merge once @danfruehauf has considered the first suggestion (https://github.com/aodn/aodn-portal/pull/2029#discussion_r42058716).

danfruehauf:
@dnahodil Tested with our ncwms and it works (WMS, date/time selectors, styles). Doesn't seem to be affected by the `service=` line. It probably just gets ignored by ncwms.

dnahodil:
Sorry, @danfruehauf, I meant this comment:
> The Z character can just be part of the format string, no need to append

web-app/js/portal/ui/openlayers/layer/NcWms.js
danfruehauf: Ah, I missed that. Yeah, I'll amend it.

danfruehauf: `console.log(time.utc().format('YYYY-MM-DDTHH:mm:ss.SSSZ'));` results in `2011-07-08T03:32:45.000+00:00`

So not good. Leave as is.

dnahodil: You'll want to escape the Z character. I think it's square brackts:
`console.log(time.utc().format('YYYY-MM-DDTHH:mm:ss.SSS[Z]'));`

danfruehauf:
@dnahodil Fixed.

PR #2030, opened 2015-10-15
Enable WPS downloads

dnahodil:
**This should probably wait to be merged until after we've got the WPS stuff working end-to-end**

Just in case...


danfruehauf:
:+1:

I feel like having PRs to remove netcdf_subset_service and gogoduck webapps too! :)

PR #2039, opened 2015-10-26
Focus the text field when the free text search facet is expanded

jkburges:
A small but persistent annoyance.


dnahodil:
A nice little fix @jkburges

Issue #305, opened 2013-07-31
Have default style selected in Right Details Panel when ncWMS Layer added


dnahodil:
Currently when you add an ncWMS Layer it is displayed with its default style (because ncWMS knows what that is). But our right details panel does not properly reflect what style is being shown. This is because the Portal doesn't know what the default style is (and ncWMS is just using it anyway because we haven't specified a different style to use).

So the enhancement is to have it such that the Portal will be able to display the default style. There are a few possible ways to achieve this, including:
- We could modify ncWMS to declare in the GetCapabilities document what the default style is for each layer (if one is set). The Portal could then capture that information and use it accordingly.
- We could add the ability to manually choose a Style in the Portal for any particular Layer. This would mean some manual work but doesn't rely on us updating the ncWMS software.

PR #2040, opened 2015-10-27
Now handles null defaultPalettes from thredds ncWMS server

pmbohm:


pmbohm:
It is not possible to get the default Palette information cheaply from a Thredds ncWMS server. So the portal changes here can gracefully handle this.

pmbohm:
Referencing #305 - ncWMS can give us defaultPalette however Thredds ncWMS is missing this functionality, as no index is built

danfruehauf:
I'll add that generally `GetCapabilities` is lacking that. Gives the styles, but not the default one. Not sure if the `GetMetadata` request that ncwms actually does is part of OGC WMS. @jonescc ??

dnahodil:
Nah, `GetMetadata` is not part of the WMS spec. It's something that ncWMS added.

danfruehauf:
So I assume it's OK for us to interact with `GetMetadata` on a "best effort" basis. We cannot actually get all the information we need out of thredds, as @pmbohm mentioned.

dnahodil:
I don't think we use `GetMetadata` for much. It might be simpler to assume it's *never* available and just do the same for stand-alone ncWMS as we do for THREDDS ncWMS.

danfruehauf:
@dnahodil `GetMadata` is used for 3 things (I don't think more):
 * Getting styles (`REQUEST=GetMetadata&item=layerDetails`)
 * Getting all available days with data (`REQUEST=GetMetadata&item=layerDetails`)
 * Getting timesteps for day (`REQUEST=GetMetadata&item=timesteps`)

dnahodil:
Okay, so the only one that we can't get with the ncWMS plugin is styles (default style/palette, really). So maybe we change this code to behave as if we can never have this information?

danfruehauf:
As discussed - I'm happy with that. So a blank box to start with after load?

dnahodil:
I think @pmbohm has set some default text which is probably better than blank. I think this is the code that can be simplified: https://github.com/aodn/aodn-portal/pull/2040/files#diff-f562570827825bf3c38891d4fb397849R119

pmbohm:
Heres what my pull will request will show when using Thredds and the Geoserver plugin: 

screen shot 2015-10-27 at 12 02 06 pm
Otherwise it loads as before. Job done.

danfruehauf:
@dnahodil OK if I leave it with you to review?

I have quite a bit on my plate now...

dnahodil:
Yeah. Code all looked good to me other than that one suggestion.

PR #2041, opened 2015-10-27
fixes #2032

### LINKED ISSUE #2032: Clear and reset does not clear and reset everything
App version: 4.16.1
            Build date: Tue Oct 13 14:49:06 AEDT 2015
            Build number: 23
            Git ref: 152a05741ae9815e67080afa3b47dbc875c9d8aa

Steps to reproduce:

Select any facet on step one.
Select a collection to be taken to step 2.
Select polygon and draw a polygon
Proceed to step 3 and click clear and reset
On step 1 view search selections
Add another collection from step 1 and when taken to step 2 look at bounding box options

What should happen:
User taken back to step 1 and search is reset - no facets selected
On step 2 Bounding box should be on default with indication that user can type co-ords if they want

What does happen:
On step 1 previous facet/search is still selected.

On step 2 the polygon co-ordinates are still visible (does not affect drawing of bb) but if it is showing as set to polygon then it should be on draw polygon but if user draws on the map they will draw a bounding box and the co-ordinate change correctly to 4 for bounding box but the drop down still says polygon.

pmbohm:


dnahodil:
@pmbohm I am getting JavaScript errors in the console when I open this in the browser.

dnahodil:
The first commit works in the browser so it'll be a change in the second commit.

dnahodil:
This change incorrectly clears the facets if I remove a single collection from the Portal. It should only clear them on 'Clear and Reset'.

pmbohm:
@dnahodil Now correctly using the `PORTAL_EVENTS.RESET` instead of `DATA_COLLECTION_REMOVED`
The javascript bug was introduced by the work on tests. Lost the place now but it was a incorrect reference to the map

dnahodil:
Changes looks good @pmbohm and works as expected

PR #2043, opened 2015-10-28
Encode ncwms layer name

danfruehauf:
Ncwms layer name will look like `workspace:layer#time,file_url/variable`.

This introduces the `#`, which will allow users to control from
metadata which `time` and `file_url` columns names they would
like to use when using the abomination ncwms controller.

After introducing the `#`, it means we need to encode the layer
name properly in JS land. The server side changes ensure that
when accessing geoserver backends (ncwms abomination controller),
the layer name will be encoded.

This does not affect regular ncwms layers where the layer does not
need to be encoded.


danfruehauf:
@dnahodil As per our discussion.

See also:
https://github.com/aodn/geoserver-build/commit/79a332ed2af0dbae8c24d8782cd1a072bd6bccd9

PR #2047, opened 2015-11-02
URL encode layer in GetLegendGraphic

danfruehauf:
Amend 41b46c4, also GetLegendGraphic needs to encode the layer as
it will contain `#`s.


danfruehauf:
This is why the legend graphic didn't work on Friday review.

PR #2050, opened 2015-11-04
New contact link in home page footer

pmbohm:


danfruehauf:
@pmbohm We decided to have it configurable in chef. Now it will be deployed also on the IMAS portals.

@markhepburn FYI.

pmbohm:
@danfruehauf You have decided to make it configurable in Chef. I on the other hand spent 10 minutes adding text in plain old HTML configurable via the Config.groovy file. That is after spending a further 4 hours trying to decipher what are the required Chef changes and then trying (unsuccessfully) to test with vagrant (using the PO box).
There is minimal gain to have this Chef configured, and maximum time wasting trying to get this into Chef. If I would have succeeded, then there would be further cost for anyone who wants to make changes to seemingly simple block of HTML. @jonescc

danfruehauf:
@pmbohm Fine. OK. But lets then not pretend next planning meeting we care about "AODN scalability" if this is how we choose to fix things.

markhepburn:
(If I can weigh in, since I was cc-d and representing IMAS / non-AODN users)

We do rely on the chef infrastructure to maintain a number of instances; I can do a PR if you like if that means our own contact info won't be overwritten?  (I have a feeling I might even have done the top-link configurability, but I can't remember)

markhepburn:
Actually -- it has been a while -- it looks like we're using the branding_base option in chef, so I think this is separate.  (I guess it's a toss-up whether that sort of info should be configurable or is part of the branding).  Anyway, I think external users won't be affected.

danfruehauf:
@markhepburn :+1: Thanks for that.

dnahodil:
@pmbohm's change has not made anything worse. He's just continued along the path the code is already on. He's not obligated to change the way this works to fix his bug.

danfruehauf:
In this case, this can be overwritten. True.

But if it wasn't the case, then are you suggesting that we should encourage other people to fork the portal so they can reuse that? That's quite the opposite of what you say in planning meetings...

dnahodil:
I think all new work we do should move towards allowing the Portal to be easily and successfully reused. But that doesn't mean that any time anyone touches any code they have to go and change the entire way the code works before they make their small change. They're not obliged to change it and, if they don't feel confident about how to make the change, it's unreasonable to expect them to.

pblain:
To summarise then:
1. The change @pmbohm made is fine.
2. Any new work must ensure that the portal remains configurable so that others (especially existing portal operators) do not need to fork it.
3. It's preferable (but not a requirement) to look for ways to further extend the re-usability of the portal.

PR #2056, opened 2015-11-10
WPS work

dnahodil:
This PR is preparation for the GoGoDuck WPS Portal work.

The major changes:
- Rename GoGoDuck code to include 'V1'. This code can be retired after we've migrated all the GN records.
- Renamed some code which is called 'WPS...' but is actually specific for the NetCDF Subset Service.
- Made some WPS code generic to allow reuse.

### DIFF
--- a/grails-app/services/au/org/emii/portal/WpsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsService.groovy
@@ -1,16 +1,14 @@
 package au.org.emii.portal
 
-import grails.gsp.PageRenderer
+import groovy.xml.StreamingMarkupBuilder
 import groovyx.net.http.*
 import org.codehaus.groovy.grails.web.mapping.LinkGenerator
 
 import grails.converters.JSON
 
-import static groovyx.net.http.Method.POST
 import static groovyx.net.http.ContentType.XML
 
 import org.ocpsoft.prettytime.PrettyTime
-import org.ocpsoft.prettytime.units.*
 
 class WpsService extends AsyncDownloadService {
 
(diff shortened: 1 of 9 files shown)


grails-app/views/wps/_asyncRequest.xml.gsp
```diff
@@ -13,7 +13,7 @@
             <wps:Reference mimeType="application/octet-stream" xlink:href="http://geoserver/wps" method="POST">
                 <wps:Body>
                     <wps:Execute version="1.0.0" service="WPS">
-                        <ows:Identifier>gs:NetcdfOutput</ows:Identifier>
+                        <ows:Identifier>gs:${jobIdentifier}</ows:Identifier>
```
jkburges: `jobType`?  ID makes it sound like UUID, i.e. an individual instance of a job.

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -1,16 +1,14 @@
 package au.org.emii.portal
 
-import grails.gsp.PageRenderer
+import groovy.xml.StreamingMarkupBuilder
```
jkburges: image

grails-app/views/wps/_asyncRequest.xml.gsp
dnahodil: I wasn't really sure of what to call it so I want with 'identifier' because that is what was in the XML tag. `jobType` is better, though.

grails-app/services/au/org/emii/portal/WpsService.groovy
dnahodil: I removed a bunch of unused imports when I had to edit this file.

There was also a fully-qualified class name here https://github.com/aodn/aodn-portal/pull/2056/files#diff-cf1774ac4be79e524ea06d4aaffd87efL47 which seemed a bit odd as you generally only do that (IMO) if you have naming conflicts within a file (e.g. two classes called JSON but from different packages), so I moved it to an import statement.

dnahodil:
Comments addressed.

grails-app/services/au/org/emii/portal/WpsService.groovy
jkburges: :+1:

PR #2058, opened 2015-11-11
New WPS GoGoDuck download handler

dnahodil:


web-app/js/portal/cart/DownloadHandler.js
```diff
@@ -89,7 +89,8 @@ Portal.cart.DownloadHandler._downloadHandlerConstructorForProtocol = function(pr
             Portal.cart.PythonDownloadHandler
         ],
         'IMOS:AGGREGATION--bodaac': Portal.cart.BodaacDownloadHandler,
-        'IMOS:AGGREGATION--gogoduck': Portal.cart.GogoduckV1DownloadHandler
+        'IMOS:AGGREGATION--gogoduck': Portal.cart.GogoduckV1DownloadHandler,
+        'OGC:WPS--gogoduck': Portal.cart.GogoduckDownloadHandler
```
dnahodil: I think this is what we agreed on for the protocol. For the NetCDF Subset Service the protocol will be `OGC:WPS--netcdf-subset-service`.

danfruehauf: jodaac?

Just a bit shorter...

web-app/js/portal/cart/GogoduckDownloadHandler.js
```diff
@@ -0,0 +1,71 @@
...
+        };
+    },
+
+    _buildWpsUrl: function(filters, layerName, serverUrl, notificationEmailAddress) {
+
+        var aggregationParams = filters.filter(function(filter) {
+            return filter.isNcwmsParams;
+        })[0];
+
+        var subset = String.format(
```
danfruehauf: Not critical, but can become a function.

web-app/js/portal/cart/DownloadHandler.js
dnahodil: I thought we weren't using the JODAAC name any more. Might be mistaken?

The GeoServer code is 'NetcdfGenerator' so maybe it should be that?

web-app/js/portal/cart/GogoduckDownloadHandler.js
```diff
@@ -0,0 +1,71 @@
...
+        return function(collection, handlerParams) {
+            var wpsUrl = _this._buildWpsUrl(
+                collection.getFilters(),
+                _this._resourceName(),
+                _this._resourceHref(),
+                handlerParams.emailAddress
+            );
+
+            if (handlerParams.challengeResponse) {
+                wpsUrl += String.format("&challengeResponse={0}", encodeURIComponent(handlerParams.challengeResponse));
```
danfruehauf: This can possibly go to the base class (`AsyncDownloadHandler`)

web-app/js/portal/cart/DownloadHandler.js
danfruehauf: Yeah, I'm not too concerned about that. Leave it like that then.

dnahodil:
Thanks for the review, @danfruehauf. Because of your comments I went back and had a look at what more could be extracted thanks to the inheritance hierarchy. There's still more that can be done but I think these changes reduce duplication and make the various handlers more consistent.

PR #2059, opened 2015-11-11
Maps - navigation and drawing buttons re-configured

pmbohm:


dnahodil:
I've found a few bugs that seem to be related to this change:
* Possible to single-click and have a blue box being drawn (should only normally be drawn for click-drag)
* Possible to get two GetFeatureInfo popups open
* 'No more information found at this point' message from GetFeatureInfo also appears in shopping cart widget

@pmbohm please let me know if you'd like me to demonstrate any of these.

pmbohm:
@dnahodil Just spent the last few hours, (three) getting rid of the shopping cart widget bug. You wernt spozed to se et. That was the definitive bug if there ever was one. Point two is fixed. 
Point one is a feature?

dnahodil:
Cool, I'll check out your changes.

dnahodil:
Just discussed with @pmbohm. Those issues seem to be better, two others are:
* There is a flickering as the GFI pop-up opens. It appears that the GFI requests are possibly happening twice, causing the flickering as the second lot of data arrives.
* Clicking near the edge of the map opens a GFI pop-up but as the map moves to give it space the arrow moves where it is pointing to.

I wouldn't mind betting the second is actually caused by the first, so I'd investigate the first one first.

pmbohm:
@dnahodil removal of the Vector layer was causing problems in the same way that removal of the control was. All handled now

PR #2062, opened 2015-11-18
Enable NetCDF subset service WPS

dnahodil:


pmbohm:
Really needs all available download handlers to be placed into config.groovy

http://stackoverflow.com/questions/1366127/instantiate-a-javascript-object-using-a-string-to-define-the-class-name

dnahodil:
> Really needs all available download handlers to be placed into config.groovy

Agreed :+1:

dnahodil:
This configuration is no longer baked into the code and now lives in application config (config.groovy).

PR #2063, opened 2015-11-18
Handling wps job report failures

pmbohm:


grails-app/controllers/au/org/emii/portal/WpsController.groovy
```diff
@@ -1,31 +1,50 @@
...
+                _renderExecutionFailed(execResponse)
+            }
+            else if (execResponse.ProcessOutputs.Output.Reference.@href != null){
+                params.status = "Preparing download"
+                _renderExecutionStatus(execResponse)
+            }
+            else {
+                params.status = "Download ready"
+                _renderExecutionStatus(execResponse)
+            }
```
dnahodil: Lines 17 - 24 look a bit odd to me; I don't think we need to determine the status here. We can just do it inside `renderExecutionStatus()` which will cut out a bit of code here.

grails-app/controllers/au/org/emii/portal/WpsController.groovy
```diff
@@ -34,21 +53,36 @@ class WpsController extends RequestProxyingController {
...
 
+    def _renderExecutionFailed(execResponse) {
+        render(
+            view: 'show',
+            model: [
+                job: [
+                    uuid: params.uuid,
+                    status: "ProcessFailed",
+                    downloadTitle: "IMOS download ERROR - ${params.uuid}",
+                    errorMessage: "message"
```
dnahodil: Are we meant to be loading the error message from somewhere?

grails-app/i18n/messages.properties
```diff
@@ -79,6 +79,8 @@ job.id.label=Job Id
 job.createdTimestamp.label=Submitted
 job.status.label=Status
 job.download.label=Download
+job.error.label=Error message
+job.error.message=Unfortunately there has been an error with the download.<br/>Please try again at a latter time or try again with a smaller download.
```
dnahodil: Typo 'latter' -> 'later'

grails-app/views/wps/_asyncRequest.xml.gsp
```diff
@@ -9,31 +9,30 @@
...
     <wps:DataInputs>
         <wps:Input>
-            <ows:Identifier>notifiable</ows:Identifier>
+            <ows:Identifier>wrappedProcessResponse</ows:Identifier>
             <wps:Reference mimeType="application/octet-stream" xlink:href="http://geoserver/wps" method="POST">
                 <wps:Body>
                     <wps:Execute version="1.0.0" service="WPS">
                         <ows:Identifier>gs:${jobType}</ows:Identifier>
-
+                        <ows:Identifier>gs:NetcdfOutput</ows:Identifier>
```
dnahodil: I don't think we need this line. We have an Identifier on line 16 (loading dynamically from the `jobType` variable)

grails-app/controllers/au/org/emii/portal/WpsController.groovy
dnahodil: Oh. I see we're using this value to build up a message code to load from messages.properties. It looks like this is the only place that uses it so maybe passing a boolean called `showErrorMessage` might make it more clear what is going on?

Or, if you didn't want to do that, at least renaming this to be `errorMessageCode` to make it clear that it's not the message itself, but the code to load the message.

grails-app/controllers/au/org/emii/portal/WpsController.groovy
```diff
@@ -1,31 +1,50 @@
...
+        def execResponse = _performProxyingIfAllowed()
+        if (execResponse != null) {
+            if (execResponse.name() == "ExceptionReport") {
+                 wpsService._notifyErrorViaEmail(params)
+            }
+            else {
+                 wpsService._notifyDownloadViaEmail(params)
+            }
+        }
         render status: 200
```
dnahodil: I know you didn't write this, but this is a good opportunity to change this 200 literal value. We have constants defined for the HTTP status codes which we use throughout the Portal. See LayerController for an example.

pmbohm:
@dnahodil Changes made from your comments. @jonescc has said that he will look at this pull request, as he can test the whole stack

dnahodil:
Okay, cool, I'll leave the rest up to @jonescc

dnahodil:
Those 3 commits look good @jonescc. Are there more problems you're looking into? (last on IRC I think was in GeoServer rather than the Portal)

PR #2066, opened 2015-11-23
Remove download handler for non-WPS GoGoDuck

dnahodil:
This change will remove the download handler for GoGoDuck v1 (non-WPS).

**Required timeline of events**
1. WPS GoGoDuck tested and deployed to production
2. Online resources in metadata for GoGoDuckable data collections updated so that the protocol  changes from `IMOS:AGGREGATION--gogoduck` to `OGC:WPS--gogoduck`. (This doesn't need to happen for all metadata records at the same time. However, at any given time, a single metadata record should only have _either_ of those protocols, but not both.)
3. This PR can be merged
4. Use WPS GoGoDuck in prod for a bit and determine that we are happy with the behaviour
5. Decommission the old GoGoDuck web-app


julian1:
@dnahodil! Are you cleaning up before you come back?

dnahodil:
Ha ha. I saw this in my list and figured it _probably_ could be closed. It's okay. I don't mind my work going to waste. Sob.

😉

Issue #2053, opened 2015-11-05
Click point has to be too precise


kereid:
Steps to reproduce:

Add a soop collection or argo.

While zoomed out click on the area where there might be a cluster of floats or a coupe of soop tracks

Zoom all the way in and click a track or a float

What should happen:

When there is a click in a general area of a float or track get some information

What does happen:

Have to be ridiculously precise to get any detail

When it's zoomed out is there someway to set the click area to encompass a larger surface? Even if this means getting details in the GFI for multiple floats or tracks it's better than the user clicking and thinking there is not information. 

Same for zooming in maybe the floats and tracks can be displayed bigger not still so small
![click presicion](https://cloud.githubusercontent.com/assets/2906602/10985197/5e1c5588-8473-11e5-8f16-3f25612f8db4.jpg)

pmbohm:
Seeing errors that are the result of the portal not sending the correct click point when clicking over the date line. These requests will all silently fail. This may be responsible for a lot of misses
```
org.geoserver.platform.ServiceException: -965, 122 not in dimensions of image: 1535, 436
	at org.geoserver.wms.featureinfo.GetFeatureInfoKvpReader.read(GetFeatureInfoKvpReader.java:181)
```

pmbohm:
Tested with styling changes for Argo and this seems to fix the issue
![screen shot 2015-11-23 at 5 31 07 pm](https://cloud.githubusercontent.com/assets/1066981/11331145/c22f6502-920a-11e5-94d4-e1c08c0d3a64.png)
Content issue created https://github.com/aodn/content/issues/154

kereid:
![click point](https://cloud.githubusercontent.com/assets/2906602/11886457/6252ba9e-a57e-11e5-8927-3cdab148b435.jpg
Re-opening as it still seems to be hard to click on a track it's just specific to one or 2 collections.

pmbohm:
Yes this specific to one or two collections. The bug needs to be reported for the collection on the content repo and the fix applied as described at aodn/content#154

PR #2067, opened 2015-11-23
Send correct click point values to GFI requests

pmbohm:
By itself this does not solve the issue that the GFI requests dont seem to return hits often enough. It does solve the issue where requests returned server errors as the value for the 'X' click point became a negative number and this always failed. Possibly upgrades to Geoserver and or Openlayers produced this outcome?
#### To Test
The error can be seen by adding the code below to the `findFeatures` function in `FeatureInfoPopup.js`. Whenever x is negative there will have been a 500 error returned. _Click on the right of the map_
```
console.log(event.xy);
this.loc = this.map.getLonLatFromViewPortPx(event.xy);
var pix = this.map.getViewPortPxFromLonLat(this.loc);
console.log(this.loc);
console.log("x: " + pix.x );
```

#### Note
This bug also comes from styling issues where lines are styled too small for the feature to be considered in GFI responses. Reported at https://github.com/aodn/content/issues/154

### DIFF
--- a/src/test/javascript/portal/ui/FeatureInfoPopupSpec.js
+++ b/src/test/javascript/portal/ui/FeatureInfoPopupSpec.js
@@ -87,17 +87,17 @@ describe("Portal.ui.FeatureInfoPopup", function()
    });
 
     describe("_clickPoint", function(){
-        it("returns integer x and y values", function()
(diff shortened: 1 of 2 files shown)


pmbohm:
fixes #2053

PR #2068, opened 2015-11-24
Controller improvements

dnahodil:
The WPS controller previously extended from RequestProxyingController to get the known servers URL verification for free. However, it was a bit messy because we were overwriting internal logic because it isn't really proxying requests for the Portal client. So I extracted the host-validation logic into a new controller so that controllers can just subclass that for the behaviour.

So instead of (mis)using `_performProxyingIfAllowed()` we can just use the new method `_ifAllowed(url)`.

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/WpsController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/WpsController.groovy
@@ -1,26 +1,29 @@
 package au.org.emii.portal
-import au.org.emii.portal.proxying.RequestProxyingController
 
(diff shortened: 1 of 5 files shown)


dnahodil:
@jonescc would you mind checking this out for me? I'd like to get it in for this round of testing so that if anything goes wrong with it, I can be here to help. Let me know if any of it isn't clear.

dnahodil:
(or let me know if you're too busy)

jonescc:
Looks good to me @dnahodil!  Will just test with remove chaining changes

PR #2072, opened 2015-11-25
Minimaps use zoomToExtent function now

### LINKED ISSUE #2048: Incorrect geographical bounding box display in metadata record display of Step1
Data bounding boxes do not display correctly. e.g. none of the satellite bboxes are complete; sst satellite data, from the metadata record spans 70°E to 170°W, 20°N to 70°S but this is not shown in the image with the metadata record, e.g. L3S 1-day composites, which cuts off eastern Australia and the Tasman Sea to NZ; Johnson southern ocean chl covers 0-360 degrees but not in the image. Others too, e.g. SOOP XBT which spans the Pacific and down to Antarctica is likewise limited.

pmbohm:
Essentially the the bug is that the whole extent of the collection should now show for all collections. Still reliant on Openlayers to do the best fit

Fixes #2048


danfruehauf:
That is solving the behaviour correctly, however it highlights another issue - the bathymetry layer used for the minimaps is incomplete and will show some gaps. But that is another problem.

PR #2078, opened 2015-11-30
Default mail config. Created for WPS

pmbohm:


danfruehauf:
Why is that required? Should be switched on only for production/test environments. By default, for dev it should be disabled as it is now.

pmbohm:
No idea. I'm copying closely https://github.com/aodn/go-go-duck/blob/master/grails-app/conf/Config.groovy#L50 for WPS
Does that mean go-go-duck is incorrect/ unsafe?

danfruehauf:
Probably the config there was redundant. Who has a mail service running on their host anyway? :)

pmbohm:
I'll put up a pull request to remove it from go-go-duck then

PR #2083, opened 2015-12-04
Ensure gogoduck always gets spatial subset specified with at least on…

jonescc:
…e decimal place

### DIFF
--- a/src/test/javascript/portal/cart/GogoduckV1DownloadHandlerSpec.js
+++ b/src/test/javascript/portal/cart/GogoduckV1DownloadHandlerSpec.js
@@ -61,7 +61,7 @@ describe('Portal.cart.GogoduckV1DownloadHandler', function () {
                         isNcwmsParams: true,
                         dateRangeStart: moment.utc('2000-01-01T01:01:01'),
                         dateRangeEnd: moment.utc('2014-12-23T23:59:59'),
-                        latitudeRangeStart: -42,
+                        latitudeRangeStart: -42.35,
                         latitudeRangeEnd: -20,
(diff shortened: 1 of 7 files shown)


src/test/javascript/portal/cart/GogoduckV1DownloadHandlerSpec.js
```diff
@@ -61,7 +61,7 @@ describe('Portal.cart.GogoduckV1DownloadHandler', function () {
                         isNcwmsParams: true,
                         dateRangeStart: moment.utc('2000-01-01T01:01:01'),
                         dateRangeEnd: moment.utc('2014-12-23T23:59:59'),
-                        latitudeRangeStart: -42,
+                        latitudeRangeStart: -42.35,
```
danfruehauf: Was the test wrong in that case?

danfruehauf: Ah, never mind, that's the input. Forget about it.

web-app/js/portal/prototypes/Number.js
```diff
@@ -0,0 +1,3 @@
+Number.prototype.toDecimalString = function() {
+	return this.toString().contains('.') ? this.toString() : this + '.0';
```
danfruehauf: So just what I did in GGD :)

src/test/javascript/portal/cart/GogoduckV1DownloadHandlerSpec.js
jonescc: Thought I should have something with a decimal place

PR #2084, opened 2015-12-04
Fix netcdf subset ui

danfruehauf:


danfruehauf:
@jonescc Give this a shot. It works for me.

PR #2085, opened 2015-12-04
Default to source content-disposition if not overridden when proxying

jonescc:


danfruehauf:
Hacky, but clear and succinct. `ProxyController` is one big hack to begin with.

PR #2100, opened 2016-01-06
AODN theme

pmbohm:


jonescc:
Not sure how this is meant to look.  In chrome, I have to scroll down to see acknowledgements etc.  I also get circles with strange symbols in them that aren't obvious to me what they should do.

grails-app/conf/Config.groovy
```diff
@@ -221,7 +221,8 @@ portal {
     siteHeader = "Open Access to Ocean Data"
     motdUrl = "https://static.emii.org.au/motd"
 
-    logo = "images/IMOS_logo.png"
+    logo = "images/AODN/AODN_logo.png"
+    secondaryLogo ="https://static.emii.org.au/images/logo/IMOS_logo-stacked3.png"
```
danfruehauf: I'm not reviewing the PR per se, but here you're missing a space.

But the more important thing is that this will become the default now, unless overriden. So either communicate it with IMAS (config changes on their side), or make the default logo blank so we can override it in our config.

grails-app/conf/Config.groovy
```diff
@@ -252,7 +327,7 @@ portal {
         wms = [
             "OGC:WMS-1.1.1-http-get-map",
             "OGC:WMS-1.3.0-http-get-map",
-            "IMOS:NCWMS--proto"
+            "IMOS:NCWMS--abom"
```
jonescc: We don't want to revert to abom in this PR!  This may go away with a rebase

grails-app/conf/Config.groovy
jonescc: Would be good if we could remove IMOS specific stuff from the base code.  Not including an IMOS specific logo here would be a good start.

jonescc: On that note - I'm now wondering why the landing page is bundled with the portal.  Landing pages will by default be organisation specific.  Something to seperate out in future?

danfruehauf: > On that note - I'm now wondering why the landing page is bundled with the portal. Landing pages will by default be organisation specific. Something to seperate out in future?

Portal branding lets the contributors set their own page. See also:
http://portal-rls.imas.utas.edu.au/

It redirects to:
http://reeflifesurvey.imas.utas.edu.au/static/landing.html

Which is a simple static page.

pmbohm: This AODN/IMOS stuff is for development and wont be included in a production war. I think its useful to list here all the config items that are used by the portal. Its like a template for external sysadmins to modify and inject and of course for eMII developers its essential.

grails-app/views/landing/index.gsp
```diff
@@ -1,96 +1,172 @@
...
+    <g:render template="/landing_theme_includes"></g:render>
+
+    <script type="text/javascript">
+        <portal:motdPopup />
+
+        $(document).ready(function() {
+            $window = $(window);
+
+            // simple parallax for <section data-type="background">...</section> tags
+            $('section[data-type="background"]').each(function() {
```
jonescc: Doesn't sound like a data-type...

grails-app/views/landing/index.gsp
```diff
@@ -1,96 +1,172 @@
...
+                }, 500);
+                return false;
+            });
+        });
+
+    </script>
+</head>
+
+<body>
+<nav class="">
```
jonescc: Any reason why we lost the indent?

jonescc:
Hmm - wish I knew more about writing css/html - not much I can comment on.   Looks good to me apart from minor comments above.  Thanks for clarifying config stuff.

pmbohm:
@jonescc new review changes pushed

grails-app/conf/Config.groovy
```diff
@@ -138,6 +138,80 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
+
```
danfruehauf: @pmbohm Why was that chunk copied under `development`? It's redundant.

pmbohm:
@danfruehauf Fixed the config.groovy file
@jonescc renamed tag attributes in landing index page

danfruehauf:
> @danfruehauf Fixed the config.groovy file

It's not really fixed. It'll break quite a lot of things actually.

danfruehauf:
@pmbohm Please in `Config.groovy`, add under `development` the secondary logo config line and leave the rest untouched. If that is not clear to you why, please come and see me.

pmbohm:
@danfruehauf Changes as per your request

danfruehauf:
> @danfruehauf Changes as per your request

That looks good. Worth testing that when secondary logo is undefined, to make sure it works as expected - should be no logo.

pmbohm:
@danfruehauf  You can see there is a gsp if tag where it is used

grails-app/views/header/_mainPortalHeader.gsp
```diff
@@ -1,13 +1,17 @@
...
     <div id="logoContainer">
-        <a href="landing"><img src="${portalBranding.logoImage}" alt="logo" id="logo" />
+        <a href="landing"><img src="${portalBranding.logoImage}" alt="main logo" id="logo" height="80" />
         </a>
     </div>
     <div id="landingHeaderContainer" class="headerHeightOverlord">
         <h1 id="headerTitle">${portalBranding.siteHeader}</h1>
     </div>
-
+    <g:if test="${portalBranding.secondaryLogoImage}">
```
danfruehauf: I'd recommend to not having such logic in a gsp, but rather have `portalBranding` return the default value, which can be just a blank image, or something similar. The logic already exists there and it's one less control path to test.

pmbohm: @danfruehauf Sorry if `secondaryLogoImage` is an empty string the user will see the text from the alt attribute. The image tag should not be rendered. The test in the gsp will need to stay. 
There is no such test on the main `logoImage` because it is expected to exist.

danfruehauf: Yep, OK.

PR #2104, opened 2016-01-13
GetFeatureInfo now uses jQuery to be draggable. Fixes #2094

### LINKED ISSUE #2094: 2 GFI open at a time
App version: 4.19.2
            Build date: Fri Dec 18 15:56:45 AEDT 2015
            Build number: 31
            Git ref: 93adc9b7627059e2fc0dba037e319ff29797f1e8

Steps to reproduce:

Add any collection
Click once to open GFI
Select pan control
Click pin on still open GFI
Click and hold GFI as though to move it 
Move mouse around really quickly 
When mouse loses connection with GFI let go of mouse button then click again on map

What does happen:

2 Get feature windows are open at the same time and have to refresh page to remove

What should happen:

Only one get feature info panel is open at a time

gfi double

pmbohm:


anguss00:
Nice work @pmbohm

PR #2109, opened 2016-01-18
Aodn landing page review

pmbohm:


grails-app/conf/Config.groovy
```diff
@@ -224,7 +225,7 @@ portal {
     siteHeader = "Open Access to Ocean Data"
     motdUrl = "https://static.emii.org.au/motd"
 
-    logo = "images/AODN/AODN_logo.png"
+    logo = "https://static.emii.org.au/images/aodn/AODN_logo.png"
```
danfruehauf: Still subject to change.

grails-app/views/_core_theme_includes.gsp
```diff
@@ -1,4 +1,6 @@
 <link rel="stylesheet" type="text/css" href="${resource(dir: 'js', file: 'ext-3.3.1/resources/css/ext-all-notheme.css')}"/>
 <link rel="stylesheet" type="text/css" href="${resource(dir: 'js', file: 'ext-3.3.1/resources/css/xBaseTheme.css')}"/>
 
-<link rel="shortcut icon" href="${resource(dir: 'images', file: 'favicon.ico')}" type="image/x-icon"/>
+<g:if test="${grailsApplication.config.portal.icon}" >
+    <link rel="shortcut icon" href="${grailsApplication.config.portal.icon}" type="image/x-icon"/>
+</g:if>
```
danfruehauf: Missing EOL.

grails-app/views/_core_theme_includes.gsp
```diff
@@ -1,4 +1,6 @@
 <link rel="stylesheet" type="text/css" href="${resource(dir: 'js', file: 'ext-3.3.1/resources/css/ext-all-notheme.css')}"/>
 <link rel="stylesheet" type="text/css" href="${resource(dir: 'js', file: 'ext-3.3.1/resources/css/xBaseTheme.css')}"/>
 
-<link rel="shortcut icon" href="${resource(dir: 'images', file: 'favicon.ico')}" type="image/x-icon"/>
+<g:if test="${grailsApplication.config.portal.icon}" >
```
danfruehauf: I think we can go through `portalBranding` here, eliminating the `if` statment.

danfruehauf:
The AODN logo image is still here, but I guess that's not too bad. It just means we can get rid of the rest of the IMOS branding I suppose.

grails-app/views/_core_theme_includes.gsp
pmbohm: The condition is useful here. If there is no `grailsApplication.config.portal.icon` there will be no broken tag created. The browser will still look in the root directory for a file called `favicon.ico` which could be desirable.

danfruehauf: Exactly, this is why you want `portalBranding`. Instead of handling the logic here in a template, you have a class.

PR #2113, opened 2016-01-22
Allow bypassing HostVerifier

danfruehauf:
If `config.allowAnyHost` is set to true, bypass HostVerifier and allow
interaction with any backend host.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -138,6 +138,10 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
+
+        // Set to true if you want to test interaction with new servers. This turns
+        // your portal instance into an open proxy and can be dangerous.
+        allowAnyHost = true
     }
 
     test {
--- a/src/groovy/au/org/emii/portal/HostVerifier.groovy
+++ b/src/groovy/au/org/emii/portal/HostVerifier.groovy
@@ -12,6 +12,9 @@ class HostVerifier {
     def allowedHosts = null
 
     def allowedHost(address) {
+        if (grailsApplication.config.allowAnyHost) {
+            return true
+        }
 
         initializeAllowedHostsIfNeeded()
 


danfruehauf:
So one might ask why this is necessary. And I say - *testing*. @smancini and @xhoenner wanted to add some servers on their PO box but had to define each individual server in their data bags. This will enable them to just play with new remote servers. Of course if they have more sophisticated download options, they'll need to define them.

So once this is merged, I'll have a PR on the chef repo to enable this on po boxes.

grails-app/conf/Config.groovy
```diff
@@ -162,6 +162,10 @@ baselayerServer = [
     wmsVersion: '1.1.1'
 ]
 
+// Set to true if you want to test interaction with new servers. This turns
+// your portal instance into an open proxy and can be dangerous.
+allowAnyHost = false
```
pmbohm: Why not put this in the `environments -> development` section and set to true? 
Great work BTW!

grails-app/conf/Config.groovy
```diff
@@ -138,6 +138,10 @@ environments {
         grails.serverURL = "http://${localhostAddress}:8080/$appName"
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-123.aodn.org.au/geonetwork"
+
+        // Set to true if you want to test interaction with new servers. This turns
+        // your portal instance into an open proxy and can be dangerous.
+        allowAnyHost = true
```
danfruehauf: @pmbohm Good idea, moved to be under `development`.

src/groovy/au/org/emii/portal/HostVerifier.groovy
```diff
@@ -12,6 +12,9 @@ class HostVerifier {
     def allowedHosts = null
 
     def allowedHost(address) {
+        if (grailsApplication.config.allowAnyHost) {
+            return true
+        }
 
         initializeAllowedHostsIfNeeded()
```
pmbohm: There may be value in placing this new code latter, after the test for an empty address? You could also then remove a return?
```
if (allowedHosts[host] || grailsApplication.config.allowAnyHost) {
            return true
        }
```

danfruehauf: I think this bypass is best at the beginning. It's allowing any host, also empty host. Everything is allowed. This is practically short-circuiting the function.

PR #2114, opened 2016-01-28
New AODN theme

pmbohm:


grails-app/conf/Config.groovy
```diff
@@ -135,7 +134,7 @@ environments {
 
         // URLs
         def localhostAddress = java.net.InetAddress.getLocalHost().getHostAddress()
-        grails.serverURL = "http://${localhostAddress}:8080/$appName"
+        grails.serverURL = "http://${localhostAddress}:9090/$appName"
```
anguss00: Why the port change?

grails-app/conf/Config.groovy
```diff
@@ -152,7 +151,7 @@ environments {
 
     production {
 
-        // URLs
+        // overwritten by Chef in production
         grails.serverURL = "http://myaodn.example.com"
         geonetwork.url = "http://catalogue-123.aodn.org.au/geonetwork"
```
anguss00: lets point this to catalogue-portal. I know it redirects, but we're not using this domain name anymore

grails-app/views/wps/show.gsp
```diff
@@ -2,14 +2,14 @@
 <html>
     <head>
         <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css">
-        <link rel="stylesheet" href="${resource(dir: 'css', file: 'imosTheme.css', absolute: true)}" type="text/css">
+        <link rel="stylesheet" href="${resource(dir: 'css', file: 'AODNTheme.css', absolute: true)}" type="text/css">
```
anguss00: We don't have to do it here and now, but given we can configure css in chef should we be looking at moving the css path to Config.groovy?

anguss00:
Will merge when Trav is happy @pmbohm

Issue #2116, opened 2016-01-29
Geographic Search - no way to reset or clear


jachope:
App Version: 4.21.0
Build Date: Wed Jan 27 15:51:57 AEDT 2016
Build Number:
GitRef:

Systest

Steps to reproduce:
1. Complete a geographic search for data collections where the map is moved by the pan hand i.e. searching in the antarctic region.
2.Select a collection
3. Select Step 3 and then 'Clear & Reset' OR Select Step 1 and select 'New Search'

What should happen:
The Geographic facet should re-set so the the map is centered and the default drawing tool activated.

What does happen:
The Geographic facet does not reset - the pan hand remains and the map will not re-center.  Need a new session for the defaults to be returned.

PR #2123, opened 2016-02-08
Reset to default bbox on geo filter when all filters are reset

anguss00:
Fixes issue #2116


jonescc:
I'll have a look at this

PR #2136, opened 2016-03-01
Citation Feature: populate download confirmation from metadata record

markhepburn:
There's a few dependencies involved before this can be usefully deployed:

* A bugfix on the otherConstraints indexing (aodn/schema-plugins#30)
* Inclusion of the licence image link in indexing (aodn/schema-plugins#31)
* dump both the otherConstraints and image link fields in geonetwork (aodn/core-geonetwork#58)

Otherwise though, it's essentially just a matter of formatting.  Since there are repeated and optional fields, I've switched to using a Extjs template instead of OL templates.

### DIFF
--- a/src/test/javascript/portal/cart/DownloadConfirmationWindowSpec.js
+++ b/src/test/javascript/portal/cart/DownloadConfirmationWindowSpec.js
@@ -11,10 +11,18 @@ describe("Portal.cart.DownloadConfirmationWindow", function() {
     var downloadFilename = "imos:argo profiles";
     var superclass;
 
+    var mockCollection = {
+        getMetadataRecord: returns({ data: '' })
+    };
+
     beforeEach(function() {
 
         confirmationWindow = new Portal.cart.DownloadConfirmationWindow();
         superclass = Portal.cart.DownloadConfirmationWindow.superclass;
+
+        // Mock UI-related methods invoked in show()
+        confirmationWindow.contentPanel.update = function() {};
+        confirmationWindow.syncShadow = undefined;
 
         spyOn(superclass, "show");
         spyOn(superclass, "hide");
@@ -24,7 +32,7 @@ describe("Portal.cart.DownloadConfirmationWindow", function() {
 
         beforeEach(function() {
 
-            confirmationWindow.show(downloadUrl, downloadFilename);
+            confirmationWindow.show({ collection: mockCollection });
         });
 
         it('shows window', function() {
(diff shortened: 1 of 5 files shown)


danfruehauf:
@markhepburn Some unit tests fail. See to that. In the mean time I'll be looking at your work.

danfruehauf:
Looks good to me. Perhaps another person should have a look at that as well? @pmbohm @anguss00 ?

pblain:
Let me have a quick look before merging

danfruehauf:
@pblain Cannot be merged yet anyway - tests need to pass.

pblain:
@danfruehauf - other than my previous comment (and failing test) it looks good to me.  I'm assuming the licence (e.g. Creative Commons) image will render OK - couldn't check that.

danfruehauf:
> I'm assuming the licence (e.g. Creative Commons) image will render OK - couldn't check that.

Is that coming from GN?

I could probably just fix the tests here and we can merge it? Is it supposed to go to prod?

pblain:
> Is that coming from GN?

Yes, it's referenced in the metadata with the other licence info

> Is it supposed to go to prod?

It will go to prod - as part of our usual deployment cycle

web-app/js/portal/cart/DownloadConfirmationWindow.js
```diff
@@ -115,6 +110,15 @@ Portal.cart.DownloadConfirmationWindow = Ext.extend(Ext.Window, {
         this.onAcceptCallback = params.onAccept;
 
         Portal.cart.DownloadConfirmationWindow.superclass.show.call(this);
+
+        var metadataRecord = params.collection.getMetadataRecord();
```
danfruehauf: @markhepburn Some mocking required for that in unit tests.

Not going to debate the quality of the unit tests just quite yet, but... You can probably plug it in an adequate way.

danfruehauf:
Not going to fix the tests here. I'll wait @markhepburn to do that...

markhepburn:
Yeah, I had to mock a couple of things (and looks like they're a tad out of date too; the second `beforeEach` didn't make much sense).  I have them passing now, but I'll take another look in the morning.  Ugh.

danfruehauf:
> I have them passing now, but I'll take another look in the morning. Ugh.

Not quite :)

markhepburn:
Passing locally I mean, nothing committed yet :) (just feeling a tad dirty with the approach, and want to see if it can be tidied up a bit)

src/test/javascript/portal/cart/DownloadConfirmationWindowSpec.js
```diff
@@ -11,11 +11,23 @@ describe("Portal.cart.DownloadConfirmationWindow", function() {
     var downloadFilename = "imos:argo profiles";
     var superclass;
 
+    var mockCollection = {
+        getMetadataRecord: function() {
```
danfruehauf: You can do:
```
getMetadataRecord: returns({data: ''})
```

PR #2140, opened 2016-03-07
Fixes #2132

### LINKED ISSUE #2132: Double accounting in Google Analytics
## How to reproduce
1. Go to Google Analytics - Portal.
1. Click Events > Top Events > DataCollections

## What happens
The stats for GoGoDuck downloads are doubled up leading to an overstated number of downloads:
image

## What should happen
Total events should give an accurate indication of the number of downloads

pmbohm:
Each download handler now calls trackUsage individually

### DIFF
--- a/web-app/js/portal/cart/BodaacDownloadHandler.js
+++ b/web-app/js/portal/cart/BodaacDownloadHandler.js
@@ -12,27 +12,34 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
 
         if (this._hasRequiredInfo()) {
 
+            var generatorFunction = this._getUrlGeneratorFunction();
+            var urlFieldName =  this._urlFieldName();
+
             downloadOptions.push({
                 textKey: 'downloadAsAllSourceNetCdfLabel',
                 type: 'WFS',
-                handler: this._getUrlGeneratorFunction(),
+                handler: this.downloadClickHandler,
                 handlerParams: {
+                    generatorFunction: generatorFunction,
+                    collectionFiltersAsTextFunction: this.getCollectionFiltersAsText,
                     filenameFormat: '{0}_source_files.zip',
                     downloadControllerArgs: {
                         action: 'downloadNetCdfFilesForLayer',
-                        urlFieldName: this._urlFieldName()
+                        urlFieldName: urlFieldName
                     }
                 }
             });
 
             downloadOptions.push({
                 textKey: 'downloadAsUrlsLabel',
-                handler: this._getUrlGeneratorFunction(),
+                handler: this.downloadClickHandler,
                 handlerParams: {
+                    generatorFunction: generatorFunction,
+                    collectionFiltersAsTextFunction: this.getCollectionFiltersAsText,
                     filenameFormat: '{0}_URLs.txt',
                     downloadControllerArgs: {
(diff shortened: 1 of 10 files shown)


web-app/js/portal/cart/BodaacDownloadHandler.js
```diff
@@ -12,27 +12,34 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
 
         if (this._hasRequiredInfo()) {
 
+            var generatorFunction = this._getUrlGeneratorFunction();
```
jonescc: Is this used?

web-app/js/portal/cart/BodaacDownloadHandler.js
```diff
@@ -12,27 +12,34 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
...
+            var urlFieldName =  this._urlFieldName();
+
             downloadOptions.push({
                 textKey: 'downloadAsAllSourceNetCdfLabel',
                 type: 'WFS',
-                handler: this._getUrlGeneratorFunction(),
+                handler: this.downloadClickHandler,
                 handlerParams: {
+                    generatorFunction: generatorFunction,
```
pmbohm: Turns out that its used here

web-app/js/portal/cart/BodaacDownloadHandler.js
jonescc: Yes it is! Ignore this comment.

web-app/js/portal/cart/DownloadPanel.js
```diff
@@ -239,11 +239,6 @@ Portal.cart.DownloadPanel = Ext.extend(Ext.Panel, {
 
         params.onAccept = function(callbackParams) {
             self.downloader.download(collection, generateUrlCallbackScope, generateUrlCallback, callbackParams);
-            trackDownloadUsage(
-                OpenLayers.i18n('downloadTrackingActionPrefix') + OpenLayers.i18n(textKey),
-                collection.getTitle(),
-                undefined
-            );
         };
```
jonescc: I reckon this is the right place to track download usage.  Otherwise we are burying this in a downloader, or a callback to generate the download url.  This doesn't seem like the right place to put UI action tracking.

I'd leave it here.

To include collection filter information applicable for the download we could just

a) add the url generated by the callback as the tracking details, 
b) include the filters applied to the collection at this point,
c) if you want the filters formatted just loop through the filters getting the human readable form

I don't think we need to format tracking related info in the download handler as it isn't tied to that and I don't think we should use the insertion service
which seems to be designed to generate formatted filter details for insertion into the download page.  If we use that then whenever we change the format there we would have 
to change getCollectionFiltersAstext to remove the page formatting.  I'd have a look at Portal.filter.combiner.HumanReadableFilterDescriber which is what the insretion services uses under the hood to
get humanreadable filter descriptions

pmbohm: What would be the indicator of which download option was chosen?
The url generated by the callback?

jonescc: textKey?

pmbohm:
Replaced by https://github.com/aodn/aodn-portal/pull/2144

Issue #2134, opened 2016-02-25
ncDownloadService does not perform subsetting correctly


smancini:
Issue found in portal-systest

App version: 4.21.5
Build date: Tue Feb 16 11:46:01 AEDT 2016
Build number: 41
Git ref: cf0e67855d506c0cd8b721f6323f9a0fca973071

### Steps to reproduce:

Select the following dataset collection:
IMOS - Australian National Mooring Network (ANMN) Facility - Temperature and salinity time-series
http://portal-systest.aodn.org.au/imos123/home?uuid=7e13b5f3-4a70-4e31-9e95-335efa491c5c
Apply the following subset:
Spatial Subset: 151.344W -33.972S 151.411E -33.917N
Temporal Extent: 2013/Mar/09-00:00-UTC to 2013/Apr/20-23:59-UTC 

On step 3, select the "NetCDF" download option.

### What should happen:

By selecting the "NetCDF" download option, I should get 1 netCDF file per instrument and deployment with data only for the selected period.

### What does happen:

I get 1 netCDF file per instrument and deployment but with data for the entire deployment.
The NcDownloadService did not perform subset on the time dimension.

In my example, there are two deployments for the mooring and the period I am interested in.
The first deployment start from 05/01/2013 and finish on the 27/03/2013.
The second deployment start from the 22/03/2013 and finish on the 05/07/2013.
The output files will contain either data from January to March or April to July depending on the deployment while I was expecting data for the period March-April.

jonescc:
Currently filtering on time_coverage_start and time_coverage_end rather than on "TIME" resulting in all data in the files with the right time coverage being returned.

PR #2143, opened 2016-03-09
Don't convert date filter from local time to UTC

jonescc:
When testing https://github.com/aodn/aodn-portal/pull/2141 for https://github.com/aodn/aodn-portal/issues/2134, I noticed that dates selected in the date filter for non-gridded data are treated as being in local time and converted to UTC before being sent geoserver for subsetting (WFS and WPS).

So 2013/3/26 would be converted to UTC and sent to Geoserver as 2013-03-25T13:00:00Z resulting in my Netcdf files (or CSV files) including data from this timestamp onwards.

Date/Times selected for gridded data are treated as being at UTC and so the treatment of date/times is inconsistent between gridded and non-gridded data.

This PR makes the treatment of the date selected for non-gridded data consistent with that of gridded data.  It also makes it work the way I was expecting - i.e. I selected data for 26/3/2013 so when I look at the file I don't expect to see data from the previous day.

### DIFF
--- a/src/test/javascript/portal/utils/TimeUtilSpec.js
+++ b/src/test/javascript/portal/utils/TimeUtilSpec.js
@@ -1,0 +1,10 @@
+describe("Portal.utils.TimeUtil", function() {
+    it('returns date formatted in iso8601 format at UTC', function() {
+        var tu = new Portal.utils.TimeUtil();
+        var testDate = new Date('2013/3/26');
+
+        var formattedDate = tu.toUtcIso8601DateString(testDate);
+
+        expect(formattedDate).toEqual('2013-03-26T00:00:00Z');
+    });
+});
--- a/web-app/js/portal/utils/TimeUtil.js
+++ b/web-app/js/portal/utils/TimeUtil.js
@@ -1,46 +1,27 @@
 Ext.namespace('Portal.utils');
 
 Portal.utils.TimeUtil = Ext.extend(Object, {
-    constructor:function(config) {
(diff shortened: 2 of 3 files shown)


web-app/js/portal/utils/TimeUtil.js
```diff
@@ -1,46 +1,27 @@
...
     },
 
-    _toUtcIso8601DateString : function(date, timeString) {
-        if (timeString) {
-            return this._toUtcIso8601DateString(Date.parseDate(date
-                            .format(this.DATE_FORMAT)
-                            + ' ' + timeString, this.DATE_TIME_FORMAT));
-        }
-        return this._toUtcDateString(date) + 'T' + this._toUtcTimeString(date);
+    toUtcIso8601DateString : function(date, timeString) {
```
pmbohm: `timeString` now not used

src/test/javascript/portal/utils/TimeUtilSpec.js
```diff
@@ -0,0 +1,10 @@
+describe("Portal.utils.TimeUtil", function() {
+    it('returns date formatted in iso8601 format at UTC', function() {
+        var tu = new Portal.utils.TimeUtil();
+        var testDate = new Date('2013/3/26');
+
+        var formattedDate = tu.toUtcIso8601DateString(testDate)
```
pmbohm: ;
Thats not a wink. Its a missing wink on this line

jonescc:
@pmbohm  ; added, timestring removed

PR #2144, opened 2016-03-10
Fixes #2132 trackDownloadUsage is only done on accepting license dialog box.

### LINKED ISSUE #2132: Double accounting in Google Analytics

pmbohm:
@jonescc Track download changes now in one place with the filters supplied as human readable text as the data

### DIFF
--- a/src/test/javascript/portal/cart/DownloadPanelSpec.js
+++ b/src/test/javascript/portal/cart/DownloadPanelSpec.js
@@ -214,6 +214,7 @@ describe("Portal.cart.DownloadPanel", function() {
                 filenameFormat: "{0}.csv"
             };
             var testCollection = makeTestCollection();
+            testCollection.getFilters = returns();
             var callbackScope = downloadPanel;
             var callback = noOp;
             var testKey = "downloadAsCsvLabel";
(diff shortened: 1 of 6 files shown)


jonescc:
Much simpler!

Small problem with unsubsetted download of gridded data specifying temporal extent - this is what is recorded

Download - Un-subsetted NetCDFs
trackUsage.js:37 IMOS - ACORN - Rottnest Shelf HF ocean radar site (Western Australia, Australia) - Delayed mode sea water velocity
trackUsage.js:38 Temporal Extent (undefined): 2014/Jun/01-10:30-UTC to 2014/Jun/06-12:30-UTC

pmbohm:
@jonescc I worked around the issue here. I cant see when this date filter ever sets 'label'

jonescc:
For GeoServer layers, the label attribute is set to the label returned by the getFilters call.  The label looks to be used when there is more than one time filter associated with the layer.

In the tested layer above, the layer is gridded and so no label is set on the filter.  A label could have been added when creating the datefilter in this case NcWmsPanel, so everything has a label, but I don't think its used anywhere for gridded layers so just handling it not being there is also OK.

PR #2145, opened 2016-03-10
Revert "Don't convert date filter from local time to UTC"

jonescc:
Reverts aodn/aodn-portal#2143

Needs more work.


jonescc:
Actually, spoke to Pete.  I'll try and fix it instead

PR #2167, opened 2016-04-01
Added failure message in case of timeout.

bhasin85:


web-app/js/portal/cart/DownloadEstimator.js
```diff
@@ -30,6 +30,13 @@ Portal.cart.DownloadEstimator = Ext.extend(Object, {
                 this._createDownloadEstimate(result, collection.getUuid(), callback);
             },
             failure: function(result, values) {
+                if (result.isTimeout) {
+                    var estimaterId = "downloadEst-" + collection.getUuid();
+                    var listOfElements = $('[id^='+estimaterId+']');
+                    if (listOfElements.length == 1) {
+                        listOfElements[0].innerHTML = '<div><span class="fa fa-fw fa-warning"></span>Unable to calculate download size.</div><div class="clear"></div>';
```
pmbohm: This is a mix of jQuery and old school it seems.
It could be reduced to (untested):
`jQuery("#downloadEst-" + collection.getUuid()).html('<div><span class="fa fa-fw........')`

web-app/js/portal/cart/DownloadEstimator.js
```diff
@@ -30,6 +30,13 @@ Portal.cart.DownloadEstimator = Ext.extend(Object, {
...
+                    }
+                }
                 this._createFailMessage(result, collection.getUuid());
```
pmbohm: @bhasin85 
Ive not tested this but it seems that the fail message should be created by `this._createFailMessage`
`createFailMessage` is creating the html in the same way as success `createDownloadEstimate` which should work.

`createFailMessage` function should be fixed here

web-app/js/portal/cart/DownloadEstimator.js
```diff
@@ -66,7 +66,7 @@ Portal.cart.DownloadEstimator = Ext.extend(Object, {
     _addDownloadEstimate: function(sizeEstimate, uuid, callback) {
 
         var htmlAddition = this._generateEstHtmlString(sizeEstimate);
-        var sizeDiv = Ext.get(this.getIdElementName(uuid));
+        var sizeDiv = Ext.select(this.getIdElementName(uuid));
```
pmbohm: This would be the first time `Ext.get` is used in our code. Are you sure you need to change this?

web-app/js/portal/cart/DownloadEstimator.js
```diff
@@ -30,7 +30,7 @@ Portal.cart.DownloadEstimator = Ext.extend(Object, {
                 this._createDownloadEstimate(result, collection.getUuid(), callback);
             },
             failure: function(result, values) {
-                this._createFailMessage(result, collection.getUuid());
+                this._createDownloadEstimate(result, collection.getUuid(), callback);
```
pmbohm: Agh !!! The missing bit. Nice

web-app/js/portal/cart/DownloadEstimator.js
bhasin85: Ext.get is unable to resolve the element and is throwing error.

pmbohm:
Could you squash your commits please

PR #2172, opened 2016-04-12
Step 1 icons using the iconSourceUuid from Geonetwork.

pmbohm:
This is a temporary solution.


pmbohm:
@pblain has asked that the logos are less prominent. I will add that feature to this pull request

PR #2192, opened 2016-05-18
Rename home controller

pmbohm:


application.properties
```diff
@@ -1,5 +1,6 @@
 #Grails Metadata file
-app.build.date=UNKNOWN
+#Wed May 18 09:37:10 AEST 2016
+app.build.date=18/05/2016 09\:37
```
anguss00: Changes to here are generated by a grails war build right? Should we back these out of the PR maybe?

anguss00:
Ok looks great aside from the one comment

Issue #2188, opened 2016-05-11
GFI incorrect for Temperature Collection


kereid:
App version: 4.22.6
            Build date: Mon May 09 16:06:32 AEST 2016
            Build number: 50
            Git ref: 85c570ea125a6292d17718a46d1c480afee997e2

The fix for #2156 although does now show a value, it is still incorrect 

Steps to reproduce:

Add any SRS SST collection, Click a point on map display to open GFI look at temperature

What does happen:

Shows temperature in Kelvin but labelled as degrees - should be labelled as Kelvin and does not have the degrees Celsius conversion 

What should happen

See numeric temperature with label beside indicating kelvin or correct measurement for display and then in parenthesis show Celsius conversion with number and label

jonescc:
Fixed by aodn/geoserver-build#138 and #2225

PR #2195, opened 2016-05-18
#2188 Fix to correct GFI for temprature collection

bhasin85:


jonescc:
This gives me:

    Value at Thu, 19 May 2016 16:59:43 GMT 303.0432 degrees 

for SRS SST L3U

This measurement is in kelvin so this doesn't make sense

pmbohm:
@bhasin85 Would you like me to take on this task?

PR #2197, opened 2016-05-19
Share link can handle grails.serverUrl with a trailing slash

pmbohm:


pmbohm:
pinched from http://stackoverflow.com/questions/6680825/return-string-without-trailing-slash

PR #2198, opened 2016-05-19
Selenium setup

bhasin85:


jonescc:
Not clear what the changes are that your proposing here - the diff here makes a lot more changes than I think you were intending.  Do you need to rebase your branch against master?

PR #2201, opened 2016-05-20
Add browser check back to landing page

jonescc:


julian1:
Are we sure we don't want to make an exception for MS Edge?

julian1:
It seems to work pretty well.

jonescc:
@pblain?

web-app/js/portal/utils/BrowserCheck.js
```diff
@@ -0,0 +1,16 @@
+function browserCheck() {
+	var supported = false; // exclusive
```
julian1: would it make sense to move this default var assigment to the place where it's actually being used - in the (isChrome || isDirefox) tlogic?

julian1: i don't really mind either way

web-app/js/portal/utils/BrowserCheck.js
```diff
@@ -0,0 +1,16 @@
...
+	var isChrome = !!window.chrome;
```
julian1: what is the double negative for?

pblain:
Let's not make an exception for MS Edge because it's not officially supported (event though it works).

PR #2216, opened 2016-06-06
Added ajax call to update the popularity counter at GeoNetwork

bhasin85:


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -311,6 +311,11 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
         this.addRecordWithUuid(uuid, multiSelect);
 
         btn.addClass(this.CSS_CLASS_BUTTON_SELECTED);
+
+        // Updating the popularity counter of the metadata record at GeoNetwork
+        Ext.ux.Ajax.proxyRequestXML({
+            url: Portal.app.appConfig.geonetwork.url + '/srv/eng/metadata.show!?uuid=' + uuid
```
jonescc: Perhaps use '/srv/eng/xml_iso19139?uuid=' instead of '/srv/eng/metadata.show!?uuid='

Results in a lot less info being transferred to the client - but has the same result (popularity is increased).

pmbohm:
@bhasin85 Did you sort out the comment from @jonescc ?

bhasin85:
@pmbohm  Look like I missed it. Working on it now.

pmbohm:
@jonescc says it will work, then it works. Nice :) 
@bhasin85

PR #2218, opened 2016-06-08
Titles are now trimmed when excessively long. Hover shows full title with default cursor

pmbohm:
part fixes https://github.com/aodn/backlog/issues/361

### DIFF
--- a/web-app/js/portal/search/FacetedSearchResultsDataView.js
+++ b/web-app/js/portal/search/FacetedSearchResultsDataView.js
@@ -20,7 +20,7 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
             '<div class="resultsHeaderBackground">',
             '    <div class="x-panel-header">',
             '        <div class="resultsRowHeaderTitle">',
-            '            <h3>{[values.title]}</h3>',
+            '            {[this.getHtmlTitle(values)]}',
             '        </div>',
             '        <div class="facetedSearchBtn" id="{[this.buttonElementId(values.uuid)]}">',
             '            {[this.getButton(values)]}',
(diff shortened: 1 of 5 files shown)


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -57,6 +57,10 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
                 },
                 getIconUrl: function(values) {
                     return Ext.ux.Ajax.constructProxyUrl(this.getGeonetworkImageUrl(values.iconSourceUuid));
+                },
+                getTrimmedTitle: function(values) {
+                    var title = values.title;
+                    return String.format("<h3 title=\"{0}\">{1}</h3>", title, Ext.util.Format.ellipsis(title, 180, true));
```
bhasin85: Do you think it will be better if we return just `Ext.util.Format.ellipsis(title, 180, true)` and update the call `{[this.getTrimmedTitle(values)]}` to `<h3 title="{[values.title]}">{[this.getTrimmedTitle(values)]}</h3>`  

Something like this
```
getTrimmedTitle: function(title) {
                    return Ext.util.Format.ellipsis(title, 180, true);
                  }
```

pmbohm: My thinking was that the adding of the title attribute to the html was best done in the function 'getTrimmedTitle'. Initially I thought 'getTrimmedTitle' would call a global function and the heading size (H3) would be an parameter to that global function, but that was going too far.
So is there a good reason to change it too your sugestion?

bhasin85: I thought it would be nice to have all HTML tags inside `_getHtmlContent` and make the function `getTrimmedTitle` do just what the name suggests. Does that make sense?

pmbohm: Ok, then if the function name was `getTitleHtml` i could call it as is?

bhasin85: That sounds good to me :+1:

pmbohm:
@bhasin85 Comments addressed now

PR #2222, opened 2016-06-23
Selenium tests

bhasin85:


pmbohm:
@bhasin85 Do you think it would be best to merge in https://github.com/aodn/aodn-portal/pull/2198
to this pull request? Maybe it is part of this, then the other pull request should be closed. This pull request should then be squashed down to `Added Selenium Project`

bhasin85:
@pmbohm @jonescc  I will create a fresh pull request and will close  #2198 and #2222

Issue #2156, opened 2016-03-22
GFI reads unkown units for temperature


kereid:
App version: 4.22.1
            Build date: Thu Mar 17 11:32:41 AEDT 2016
            Build number: 45
            Git ref: e4226d3eb99a6ea6ed28cb92c95bca91afaa37a5
        -->

Steps to reproduce:

Add any temperature collection ie IMOS - SRS SATELLITE - SST L3S - 03 day composite - day time
Once collection is laoded on step 2 click any point on the collection

What does happen:
GFI opens and see Value at Thu, 31 Dec 2015 03:20:00 GMT 297.92636 (unknown units) 

What should happen:

GFI opens and see Value at Thu, 31 Dec 2015 03:20:00 GMT 297.92636kelvin (19.85 celcius)

Could have been the other way around where it had kelvin in parenthesis can't remember now that I can't see it

bhasin85:
Issue Fixed. Pull request created.

kereid:
Reopening as pull request was not merged and is now stale

pmbohm:
fixed by https://github.com/aodn/aodn-portal/pull/2166

pmbohm:
#2166 Incorrectly used the layer.units (http://dev.openlayers.org/docs/files/OpenLayers/Layer-js.html#OpenLayers.Layer.units) as a fall back when metadata units are missing.

jonescc:
Fixed by https://github.com/aodn/geoserver-build/pull/138 and https://github.com/aodn/aodn-portal/pull/2225

PR #2225, opened 2016-06-28
Fixes 2188 - unit information comes from metadata only

pmbohm:
Units will not be stated if the metadata does not contain units.

Related work to correctly return units comes from https://github.com/aodn/geoserver-build/pull/138


pmbohm:
fixes https://github.com/aodn/aodn-portal/issues/2156

jonescc:
Looks good.  Should merge with aodn/geoserver-build#138

Issue #2224, opened 2016-06-27
Download at a point does not download a CSV file 


kereid:
App version: 4.23.7
            Build date: Fri Jun 24 11:34:38 AEST 2016
            Build number: 59
            Git ref: 6d7ad76b52ceba0e4ebbaab92c222731feb19bda
        -->

Steps to reproduce:

Add a collection that has point co-ordinates enabled, subset to one-two days, enable check box and proceed to step 3.

Select dropdown that is csv to download, try to open downloaded file

What should happen:

Downloaded file is a CSV file

What does happen:

Download is not a csv file

pmbohm:
@kereid Can you tell me which layer and server please? 
A share link would be nice :)

kereid:
@pmbohm harvesters ran last night so got point options dissapeared @jonescc is manually running to get it back on there but it will be on http://portal-systest.aodn.org.au/search?uuid=8b469535-eb48-4f1f-9612-bdf2e83cdce8

jonescc:
Transform has been re-run - looks like it will get wiped out every day with the way harvesters are currently setup.

I'm currently looking at this issue - the response is not being returned with mimetype text/csv which is why the browser doesn't offer to download it like other downloads.  The contents are csv.

jonescc:
Fixed by https://github.com/aodn/aodn-portal/pull/2226 and https://github.com/aodn/geoserver-build/pull/141

PR #2226, opened 2016-06-28
Addresses #2224 - proxy content type of WPS outputs

jonescc:
Requires https://github.com/aodn/geoserver-build/pull/141


jonescc:
@pmbohm  requires supporting changes in aodn/geoserver-build#141

pmbohm:
@jonescc Merging untested (by myself) because the worst that could happen, is still better than existing behaviour

jonescc:
@pmbohm, requires supporting changes in aodn/geoserver-build#141.  That pull request should have been reviewed/merged with this one.

PR #2235, opened 2016-07-07
Selenium Setup Files

bhasin85:


bhasin85:
@jonescc  @pmbohm  New pull request for the selenium setup

selenium/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>au.org.emii.portal</groupId>
+    <artifactId>aodn-portal-selenium</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <build>
+        <testSourceDirectory>src/test/java</testSourceDirectory>
```
jonescc: This is the default - Don't normally have to specify this

selenium/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+        <resources>
+            <resource>
+                <directory>src/test/resources</directory>
```
jonescc: This is the default - don't normally have to specify this

selenium/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.uncommons</groupId>
+            <artifactId>reportng</artifactId>
+            <version>1.1.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
```
jonescc: Perhaps use either junit or testng.  Seems unnecessary to use both.   I haven't used testng myself before - but it looks good and won't take much to change to.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
```diff
@@ -0,0 +1,19 @@
+package au.org.emii.portal.tests;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.testng.annotations.Test;
+
+public class BrowserStackTest extends BaseTest {
+
+    @Test(invocationCount = invocationCount, threadPoolSize = threadPoolSize)
+    public void BrowserStackTest() {
```
jonescc: Tests should normally be methods not constructors e.g. testBrowserStackSearch.  Using a constructor makes it difficult to use many of the other features of testng such as @Before and @After which all assume you are writing tests as methods.   Also, I wouldn't normally put a doing operation like this into a constructor - it would normally be used for initialising the instance/setting instance variables.

This seems to be testing that browserstack integration is working.  I don't think we need to test that explicity - we'll find that out when we try to perform other tests.

Would also not put a println statement in a test method like that except during local debugging.

jonescc: Actually its named like a constructor which is confusing - the above is actually a method.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
```diff
@@ -0,0 +1,19 @@
+package au.org.emii.portal.tests;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.testng.annotations.Test;
+
+public class BrowserStackTest extends BaseTest {
+
+    @Test(invocationCount = invocationCount, threadPoolSize = threadPoolSize)
```
jonescc: Perhaps only use invocationcount and threadPoolSize were the test require it (none of yours do so I'd purge these references from the project altogether)..

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public static String BROWSER_STACK_BUILD;
+    public static String AODN_PORTAL_HOME_PAGE;
+    public static String AODN_PORTAL_SEARCH_PAGE;
+    public static Properties properties = new Properties();
+    private static Logger log = Logger.getLogger(BaseTest.class.getName());
+
+    private WebDriver driver;
+
+    @Parameters({"browser", "browser_version", "os", "os_version", "device", "platform", "resolution"})
+    @BeforeClass
```
jonescc: Would it make better sense to start afresh before each test?

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
```diff
@@ -0,0 +1,128 @@
...
+            driver = new RemoteWebDriver(new URL(BROWSER_STACK_LOCAL_URL), capability);
+        } else {
+            driver = new RemoteWebDriver(new URL(BROWSER_STACK_URL), capability);
+        }
+
+        driver.get(AODN_PORTAL_HOME_PAGE);
+        wait(5);
+    }
+
+    @org.testng.annotations.AfterClass
```
jonescc: Would it make better sense to refresh after each test

Also perhaps using an import statement would make this easier to read.

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testng.annotations.Test;
+
+//Link to IMOS website
+public class ImosLinkCheckTest extends BaseTest {
+
+    private static Logger log = Logger.getLogger(ImosLinkCheckTest.class.getName());
+
+    //Click on the IMOS link at the top right of the screen (22/08 now at bottom of screen)
+    @Test(invocationCount = invocationCount, threadPoolSize = threadPoolSize)
+    public void ImosLinkCheckTest() {
```
jonescc: Again named like a constructor but is a method.  java method should start with a lower case letter and should not be named the same as the class

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public static Properties properties = new Properties();
+    private static Logger log = Logger.getLogger(BaseTest.class.getName());
+
+    private WebDriver driver;
+
+    @Parameters({"browser", "browser_version", "os", "os_version", "device", "platform", "resolution"})
+    @BeforeClass
+    public void setup(String browser, String browser_version, String os, String os_version, String device, String platform, String resolution) throws MalformedURLException, InterruptedException {
+
+        BROWSER_STACK_AUTOMATE_KEY = System.getProperty("browserstack.automateKey");
```
jonescc: Setting statics this way is unusual.  I'd make them instance fields or load them up in a static initialiser.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
```diff
@@ -0,0 +1,128 @@
...
+        capability.setCapability("browserstack.video", BROWSER_STACK_VIDEO);
+        capability.setCapability("build", BROWSER_STACK_BUILD);
+        return capability;
+    }
+
+    public WebDriver getDriver() {
+        return driver;
+    }
+
+    public boolean isElementClickable(By locator) {
```
jonescc: Perhaps move into one of your utils classes - utility functions like this seem to mostly be defined there.

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+        // Go to home page
+        getDriver().get(AODN_PORTAL_HOME_PAGE);
+        int invalidLinksCountHomePage = SeleniumUtil.validateInvalidLinks(getDriver());
```
jonescc: Seem to be passing the driver to every invocation of methods on this class - may be better to instantiate the class with the driver up front and change static methods to instance methods so you don't have to pass it all the time.

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
```diff
@@ -0,0 +1,135 @@
...
+
+public class SeleniumUtil {
+
+    private static Logger log = Logger.getLogger(SeleniumUtil.class.getName());
+
+    public static String executeCurl(String command) {
+        Process curlProc;
+        String curlResponse = null;
+        try {
+            curlProc = Runtime.getRuntime().exec(command);
```
jonescc: I'd use native java methods to test links rather than calling an operating system command such as curl

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
```diff
@@ -0,0 +1,135 @@
...
+            BufferedReader stdInput = new BufferedReader(new
+                    InputStreamReader(curlProc.getInputStream()));
+            while ((curlResponse = stdInput.readLine()) != null) {
+                System.out.println(curlResponse);
+            }
+
+            // read any errors from the attempted command
+            BufferedReader stdError = new BufferedReader(new
+                    InputStreamReader(curlProc.getErrorStream()));
+            while ((curlResponse = stdError.readLine()) != null) {
```
jonescc: Trying to drain the stdout and sterr of a process this way will cause tests to block waiting for the wrong output.  Far easier to use java to test links.

selenium/src/test/java/au/org/emii/portal/tests/ImosLogoCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testng.annotations.Test;
+
+//IMOS Logo should be present
+public class ImosLogoCheckTest extends BaseTest {
+
+    private static Logger log = Logger.getLogger(ImosLogoCheckTest.class.getName());
+
+    //Look at logo in top right corner of any step including landing page
+    @Test(invocationCount = invocationCount, threadPoolSize = threadPoolSize)
+    public void ImosLogoCheckTest() {
```
jonescc: See previous comments about method naming

selenium/src/test/java/au/org/emii/portal/utils/WebElementUtil.java
```diff
@@ -0,0 +1,208 @@
+package au.org.emii.portal.utils;
+
+import org.apache.log4j.Logger;
+import org.junit.Assert;
```
jonescc: Perhaps use the testng assert class so we don't have to use junit as well

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+        Assert.assertEquals(invalidLinksCountHomePage, 0);
+
+        // Go to search page - Step 1
+        getDriver().get(AODN_PORTAL_SEARCH_PAGE);
+        wait(2);
+        int invalidLinksCountSearchPageStep1 = SeleniumUtil.validateInvalidLinks(getDriver());
+        Assert.assertEquals(invalidLinksCountSearchPageStep1, 0);
```
jonescc: Perhaps  get a list of links and assert each link works so that invalid links appear in the report rather than using log.error and just getting an invalid link somewhere message

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
```diff
@@ -0,0 +1,135 @@
+package au.org.emii.portal.utils;
+
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.conn.ssl.SSLContexts;
```
jonescc: This is deprecated - perhaps use an undeprecated option

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
```diff
@@ -0,0 +1,135 @@
...
+            SSLContext sslContext = SSLContexts.custom()
+                    .loadTrustMaterial(null, new TrustStrategy() {
+
+                        @Override
+                        public boolean isTrusted(final X509Certificate[] chain, final String authType) throws CertificateException {
+                            return true;
+                        }
+                    })
+                    .useTLS()
+                    .build();
```
jonescc: The methods used here ar deprecated - perhaps use undeprecated alternatives instead.

jonescc:
Would it have been better to integrate the tests here into the portal code?   Did you look at using grails/selenium support?

selenium/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+            <artifactId>commons-lang3</artifactId>
+            <version>3.0</version>
+            <scope>test</scope>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/com.beust/jcommander -->
+        <dependency>
+            <groupId>com.beust</groupId>
+            <artifactId>jcommander</artifactId>
+            <version>1.48</version>
+        </dependency>
```
jonescc: Should this be scoped to test?

selenium/src/test/resources/reportng-1.1.4/CHANGELOG.txt
```diff
@@ -0,0 +1,57 @@
+Changes in version 1.1.4
```
jonescc: Not clear what this reporting directory is used for.  Its unusual for jars to be found in the resources directory for example.   Has this been accidently included - or were you using it for command line report generation or something?

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
```diff
@@ -0,0 +1,19 @@
...
+    public void BrowserStackTest() {
+        getDriver().get("http://www.google.com");
```
jonescc: Perhaps move to utils class

jonescc:
Hmm, thinking about this more, integrating with portal code would only make sense if we were running the tests as a part of the maven build - e.g. during integration testing, and we would need to be doing that against defined test data.   That would be really good - but that's not how we are intending to use this is it.   Is this then system testing tests?  Given the dynamic nature of content in systest (new collections added and changed etc) how well will that work?  Should we be looking at setting up controlled testing data?

jonescc:
@bhasin85 initial review/comments made.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
bhasin85: @jonescc  Fixed it.  You are right, I was using the println for local testing and forgot to delete it.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
bhasin85: Done. Good Suggestion.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
bhasin85: When we will be running multiple tests in parallel against different environments, the driver object will be different for tests and will therefore need to initialize it separately for every test.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
bhasin85: @jonescc  Do you mean refreshing the url? 

Fixed the import statement.

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
bhasin85: Fixed.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
```diff
@@ -0,0 +1,128 @@
...
+        DesiredCapabilities capability = getDesiredCapability(browser, browser_version, os, os_version, device, platform, resolution);
+
+        if (BROWSER_STACK_LOCAL.equals("true")) {
+            driver = new RemoteWebDriver(new URL(BROWSER_STACK_LOCAL_URL), capability);
+        } else {
+            driver = new RemoteWebDriver(new URL(BROWSER_STACK_URL), capability);
+        }
+
+        driver.get(AODN_PORTAL_HOME_PAGE);
+        wait(5);
```
bhasin85: Here I am just waiting to allow the resources to load up before beginning the test. I do not have any condition, but we can surely reduce the time to 3 seconds. In my experience, it takes that much time for slow browsers like IE.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
bhasin85: Done.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
bhasin85: Done.

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
bhasin85: Good idea. Done.

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
bhasin85: It is not getting used anywhere. I have deleted it. Also, I am using the native java methods to test link in 

`verifyURLStatus` method

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
bhasin85: Removed the curl method.

selenium/src/test/java/au/org/emii/portal/tests/ImosLogoCheckTest.java
bhasin85: Fixed.

selenium/src/test/java/au/org/emii/portal/utils/WebElementUtil.java
bhasin85: Removed Junit from dependency.

selenium/src/test/java/au/org/emii/portal/tests/ImosLinkCheckTest.java
bhasin85: Added error log.

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
bhasin85: Fixed.

selenium/src/test/java/au/org/emii/portal/utils/SeleniumUtil.java
bhasin85: Replaced the methods.

selenium/pom.xml
bhasin85: Removed the junit dependency.

selenium/pom.xml
bhasin85: Yes and Fixed

selenium/src/test/resources/reportng-1.1.4/CHANGELOG.txt
bhasin85: Removed the folder reportng-1.1.4. I added it while experimenting and left it there.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
bhasin85: @jonescc  Not sure why you suggested to move this to utils?

bhasin85:
So basically the Selenium test is "Automated Functional" tests. The plan is to add an additional step in the portal pipeline to run the selenium tests and automate the Kate's test cases (to begin with).

If the test case needs the test data which is not available in the systest environment then the test will fail. Setting up test data can be done while provisioning the systest environment. We can have a chat about this on Monday.

bhasin85:
@jonescc  I have not yet looked into the integration with grails. We do not necessarily need to tie this up with portal. I think it is easier to maintain and run if it is independent of portal code.

bhasin85:
@jonescc  initial changes made.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
jonescc: No probs, can see now that that this is how parameterized tests are setup in testng/browserstack examples.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
jonescc: As above, I can see now that this is as per browserstack/testng examples for parameterized tests so this looks fine.

jonescc:
Looks to me like in grails you would use Geb (a previous team member has done this in aatams).   There's quite a bit of functionality that Geb provides that may be useful.  In the past we have preferred grails/groovy over java except where we wish to contribute to java open source projects.  So just noting there's a preferred technology switch here which should be highlighted for other team members benefit.  I personally would have liked to have seen how this looked using Geb.  Often its a bit harder work in java.

selenium/src/test/java/au/org/emii/portal/tests/BrowserStackTest.java
jonescc: Just seems cleaner to me using a general utility class throughtout rather than using the lower level collaborators but its not a biggie.

selenium/src/test/java/au/org/emii/portal/tests/BaseTest.java
jonescc: Using hard coded waits like this is normally a bad idea.  In this case you have to wait for the worst possible start time for every test. Refer http://www.seleniumhq.org/docs/04_webdriver_advanced.jsp.

bhasin85: Changed it to implicit wait.

jonescc:
Re controlled test data - if we need to do that in systest then we are probably diverging from what systest is meant to be used for.  But can adopt a try it and see approach for the moment.

bhasin85:
> Looks to me like in grails you would use Geb (a previous team member has done this in aatams). There's quite a bit of functionality that Geb provides that may be useful. In the past we have preferred grails/groovy over java except where we wish to contribute to java open source projects. So just noting there's a preferred technology switch here which should be highlighted for other team members benefit. I personally would have liked to have seen how this looked using Geb. Often its a bit harder work in java.

@jonescc  I totally see your point. I understand that there will be some advantages for using grails/groovy. Advantages of using Java will be that more support is available (which made it easy for me to integrate with BrowserStack). I don't mind switching to Geb if there are more benefits than challenges.  I will raise a discussion item and we can get everyone's thoughts on this.

Issue #2230, opened 2016-07-05
Sign off in email from GoGODuck Version 2 incorrect


kereid:
App version: 4.23.8
            Build date: Mon Jul 04 10:58:28 AEST 2016
            Build number: 60
            Git ref: 6daffbcb2eb03fb82930ed54c770325371297c72

Steps to reproduce:

Select any gridded data collection to add to step 2, on step 2 subset to a small amount, on step 3 select download as netcdf and wait for emails to come through. 

What should happen:

Email sign off reads:

Kind Regards,
The AODN Team.

What does happen:

Email sign off reads:

Kind Regards,
The eMII Team

PR #2237, opened 2016-07-12
Fixes 2230

pmbohm:
Corresponding Chef changes wont be required by AODN but this is could be of interest to @markhepburn and or @fxmzb123  ?

### DIFF
--- a/grails-app/services/au/org/emii/portal/WpsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsService.groovy
@@ -49,15 +49,25 @@ class WpsService extends AsyncDownloadService {
 
         params.expirationPeriod = _getExpirationPeriod()
         params.jobReportUrl = _getJobReportUrl(params)
-        params.contactEmail = grailsApplication.config.portal.contactEmail
+        params.portalConfig = grailsApplication.config.portal
 
-        mailService.sendMail {
-            async true
-            to params.email.to
-            subject params.email.subject
-            body(
-                view: "/wps/${params.email.template}",
-                model: params
+        if (grailsApplication.config.grails.mail.disabled != true) {
+            mailService.sendMail {
+                async true
+                to params.email.to
+                subject params.email.subject
+                body(
+                    view: "/wps/${params.email.template}",
+                    model: params
+                )
+            }
+        }
+        else {
+            log.info(
+                groovyPageRenderer.render (
+                    view: "/wps/${params.email.template}",
+                    model: params
+                )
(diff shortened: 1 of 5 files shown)


grails-app/views/wps/jobComplete.gsp
```diff
@@ -1,13 +1,13 @@
 Hi,
 
-Your IMOS subset has been created. You can download it by clicking here: ${jobReportUrl}
+Your ${grailsApplication.config.portal.siteAcronym} subset has been created. You can download it by clicking here: ${jobReportUrl}
```
jonescc: @pmbohm - refer https://github.com/aodn/aodn-portal/pull/2220 - grailsApplication is not available in this template and the above results in a null pointer exception.

jonescc: Perhaps pass the config as a parameter

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -49,15 +49,26 @@ class WpsService extends AsyncDownloadService {
...
-        mailService.sendMail {
-            async true
-            to params.email.to
-            subject params.email.subject
-            body(
-                view: "/wps/${params.email.template}",
-                model: params
+        params.portalConfig = grailsApplication.config.portal
+
+        if (grailsApplication.config.mail.disabled == true) {
```
jonescc: Shouldn't it be the other way around?

pmbohm: `grailsApplication.config.mail.disabled` never existed either, so i always got what i expected!

pmbohm:
Fixes #2230

grails-app/services/au/org/emii/portal/WpsService.groovy
```diff
@@ -49,15 +49,26 @@ class WpsService extends AsyncDownloadService {
...
+                subject params.email.subject
+                body(
+                    view: "/wps/${params.email.template}",
+                    model: params
+                )
+            }
+        }
+        else {
+            // email is disabled for dev's
+            println (
```
jonescc: Perhaps use a logging statement here instead so output can be controlled by logging framework like other logging statements.

jonescc:
Looks good in vagrant - just one comment re using a println.

pmbohm:
`log.info` works fine. It always has and always will, but I'll need to be aware maybe it wont work in all files. @jonescc

PR #2240, opened 2016-07-14
Issue 2233

bhasin85:


web-app/js/ext-3.3.1/ext-all-debug.js
```diff
@@ -40651,19 +40651,11 @@ Ext.form.DateField = Ext.extend(Ext.form.TriggerField,  {
...
-    
-    
-    
 
-    
-    defaultAutoCreate : {tag: "input", type: "text", size: "10", autocomplete: "off"},
+    compareText : "From date {0} should be before To date {1}",
 
-    
+    defaultAutoCreate : {tag: "input", type: "text", size: "10", autocomplete: "off"},
```
pmbohm: **You cant change any Extjs (core) file!!!!!!!!!**

Notice that  production includes
`<script src="/js/ext-3.3.1/ext-all.js" type="text/javascript"></script>`
Whereas while developing we are using
`<script type="text/javascript" src="/js/ext-3.3.1/ext-all-debug.js">`

The solution may be to use `js/portal/prototypes/Ext.js` if essential?
(It seem this files purpose has morphed anyhow)

anguss00:
There's a bug that I found where if you follow the steps replicate the bug, and then select a from date before the to date (correct pattern) the red error boxes remain even after clicking out of the field

bhasin85:
@anguss00  Give it a go now! :+1:

PR #2250, opened 2016-08-03
fixes #2231

### LINKED ISSUE #2231: Using share link can result in errors
### What happens
http://portal-systest.aodn.org.au/search?uuid=8b469535-eb48-4f1f-9612-bdf2e83cdce8
- Loads IMOS - ACORN - Rottnest Shelf HF ocean radar site (Western Australia, Australia) - Real-time sea water velocity collection but produces a fatal error.
- Portal is broken
- Load the same ACORN layer without the direct link and the layer loads as expected

### What should happen
The share links for all collections should always suceed

pmbohm:


jonescc:
Looks good.

jonescc:
Just noticed there's a client log error in my grails log:

ERROR 2016-08-04 10:53:42,973 [http-bio-8080-exec-2] client-log - session ID: 896A7F794, message: Can't create Filter for dataCollection 'IMOS - ACORN - Rottnest Shelf HF ocean radar site (Western Australia, Australia) - Real-time sea water velocity' from data: {"label":"Time","type":"TimeSeries","name":"timesteps","possibleValues":["2010-03-10T00:00:00Z","2010-03-11T00:00:00Z","2010-03-12T00:00:00Z","2010-03-13T00:00:00Z","2010-03-14T00:00:00Z","2010-03-15T00:00:00Z","2010-03-16T00:00:00Z","2010-03-17T00:00:00Z","2010-03-18T00:00:00Z","2010-03-19T00:00:00Z","2010-03-20T00:00:00Z","2010-03-21T00:00:00Z","2010-03-22T00:00:00Z","2010-03-23T00:00:00Z","2010-03-24T00:00:00Z","2010-03-25T00:00:00Z","2010-03-26T00:00:00Z","2010-03-27T00:00:00Z","2010-03-28T00:00:00Z","2010-03-29T00:00:00Z","2010-03-30T00:00:00Z"

pmbohm:
@jonescc So now we are attempting to create a filter for the type "TimeSeries" and there is no filter (`FilterService.js:36`).
The question is should this output now only be `log.info` and therefore ignored, or should the filter be created and loaded via this mechanism?

jonescc:
Perhaps down the track the ncwms panel should be built in the same way as the geoserver subset panels - but that's a bit of a refactor.  For the moment, if this is not actually causing any problems - I wouldn't report this for ncwms filters - its noise.

PR #2256, opened 2016-08-08
Google style popups - image based pointer

pmbohm:
A foreign order requested by @pblain


pmbohm:
@pblain Styling as per discussion

PR #2260, opened 2016-08-15
Updated portal config to run selenium tests against portal edge

bhasin85:


selenium/pom.xml
```diff
@@ -118,7 +118,7 @@
         <browserstack.automateKey>hp3vg22JUprFFZzESJrz</browserstack.automateKey>
         <browserstack.username>developers44</browserstack.username>
         <browserstack.local>false</browserstack.local>
-        <aodnPortal>https://portal-systest.aodn.org.au/</aodnPortal>
+        <aodnPortal>https://portal-edge.aodn.org.au/</aodnPortal>
```
jonescc: Shouldn't this be configurable outside of the project?

bhasin85: @jonescc  This can be overridden as runtime parameter. Our default environment will be edge instead of systest and therefore I updated the config.

jonescc: OK so I also notice that we are publishing our username and automateKey details in a public repo  - is this an issue?

bhasin85: Yes, that is an issue. I will remove those configurations. :+1:

PR #2272, opened 2016-08-25
Test rail 1

bhasin85:
Merge Before #2290

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/landing/AcknowledgementTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/landing/AcknowledgementTest.java
@@ -1,0 +1,17 @@
+package au.org.emii.portal.tests.landing;
+
+import au.org.emii.portal.tests.BaseTest;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+
+public class AcknowledgementTest extends BaseTest {
+
+    @Test
+    public void linkTest() {
+        webElementUtil.clickLinkContainingText("Acknowledgement");
+        webElementUtil.switchToNewTab();
+        webElementUtil.verifyPageTitle("Data Use Acknowledgement | Portal User Guide");
+        webElementUtil.verifyInnerHtml("<h2>Data Licencing</h2>");
+     }
+}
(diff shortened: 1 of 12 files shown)


bhasin85:
Please ignore the codacy/pr check. Thanks

bhasin85:
@jonescc  @julian1  @anguss00  @pmbohm  Can someone please look:telescope: into this PR?

Thanks  :+1:

selenium/src/test/java/au/org/emii/portal/tests/landing/AcknowledgementTest.java
```diff
@@ -0,0 +1,17 @@
...
+import au.org.emii.portal.tests.BaseTest;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+
+public class AcknowledgementTest extends BaseTest {
+
+    @Test
+    public void linkTest() {
+        webElementUtil.clickLinkContainingText("Acknowledgement");
```
anguss00: Is there any way to link these text comparisons with the values in en.js within the main portal? Otherwise, as soon as the wording changes we're going to break a stack of tests. It would be best not to duplicate simple text config if at all possible.

selenium/src/test/java/au/org/emii/portal/tests/landing/ContactLinkTest.java
```diff
@@ -0,0 +1,35 @@
...
+    public void linkTest() {
+        goToHome();
+        WebElement contact = webElementUtil.findElement(By.linkText("Contact"));
+        Assert.assertNotNull(contact);
+        webElementUtil.clickLinkContainingText("Get Ocean Data Now");
+
+        // Step 1
+        contact = webElementUtil.findElement(By.linkText("Contact"));
+        Assert.assertNotNull(contact);
+        webElementUtil.clickButtonWithTitle("Add this collection");
```
anguss00: ^ Same applies for all of these text comparisons

selenium/src/test/java/au/org/emii/portal/tests/landing/AcknowledgementTest.java
bhasin85: It is definitely possible to do that, but it defeats the purpose of black box testing. Functional testing is from user point of view and therefore I think it is not a good idea to bind it with the project config.  

I am happy to have a chat with you regarding the same and we can find a solution together.

PR #2290, opened 2016-08-29
Selenium test for point time series lat/lon update

bhasin85:
Merge After #2272


selenium/src/test/java/au/org/emii/portal/tests/landing/AodnLinkTest.java
```diff
@@ -0,0 +1,14 @@
...
+import au.org.emii.portal.tests.BaseTest;
+import org.testng.annotations.Test;
+
+public class AodnLinkTest extends BaseTest {
+
+    @Test
+    public void linkTest() {
+        webElementUtil.clickLinkContainingText("AODN");
+        webElementUtil.switchToNewTab();
+        webElementUtil.verifyPageTitle("Australian Ocean Data Network: IMOS.org.au");
```
pmbohm: Are these test intended to be run in the dev environment?

bhasin85: These tests are intended for portal-edge(for now). We can also run it against dev environment if we want.

selenium/src/test/java/au/org/emii/portal/tests/landing/AcknowledgementTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.ArrayList;
+
+public class AcknowledgementTest extends BaseTest {
+
+    @Test
+    public void linkTest() {
+        webElementUtil.clickLinkContainingText("Acknowledgement");
+        webElementUtil.switchToNewTab();
+        webElementUtil.verifyPageTitle("Data Use Acknowledgement | Portal User Guide");
```
pmbohm: Can groovy environment variables be used in the comparison?
Is it worth (or possible) importing the en.js file so it could compare strings that are used from it?

I'm thinking these tests are all so brittle

bhasin85: The landing page link tests are validating links outside the portal. In PointTimeSeriesLatLonTest, we are using the text "Point timeseries" which can be read from en.js. The drawback will be that it will create a cross dependency of groovy and maven projects (which is not very clean and will increase the complexity). 

Also, I think it is good to keep the two projects independent. This way we have the advantage running the same tests against multiple environment without making config changes. Surely, it will add some extra maintenance. It is almost same like maintaining functional tests.

pmbohm:
I'm not the one to merge this pull request. I made comments and opened up a philosophical debate which may or may not actually exist.
Anyone else?

bhasin85:
Thanks @pmbohm  :clap:

anguss00:
Conflicts in here @bhasin85, and please could you squash the commits?

bhasin85:
New [PR](https://github.com/aodn/aodn-portal/pull/2306)

Issue #2279, opened 2016-08-26
ACORN layers not rendered properly


bhasin85:
**Steps to reproduce**

1. Select "IMOS - ACORN - Rottnest Shelf HF ocean radar site (Western Australia, Australia) - Real-time sea water velocity"
2. Go to step 2 and zoom in to about halfway on the zoom bar so you can see arrows
![1](https://cloud.githubusercontent.com/assets/2552378/17992746/6787d2d0-6b8d-11e6-9049-864bfdcd5807.png)

3. Change the To date to yesterday's date
![2](https://cloud.githubusercontent.com/assets/2552378/17992749/6d69cba4-6b8d-11e6-8bac-aa25690ee63d.png)

4. Change the To date to today's date
![3](https://cloud.githubusercontent.com/assets/2552378/17992750/72aed596-6b8d-11e6-98ef-88aebbec5812.png)

**What I expected**

Map image for step 2 and step 4 should be same.
"Displaying: 2016-08-26 00:30:00:000 UTC" within Temporal Extent should get updated

**What happened**

Different Map images
"Displaying: 2016-08-26 00:30:00:000 UTC" within Temporal Extent displays wrong date

kereid:
@bhasin85 I've just been looking through the bugs for this iteration and have noticed that the displayed time is actually what differs in this issue. The default is the last time on the day (when collection initially added), and then when you change the date then change it back it then shows the first time on that day. The display time for step 2 is 2:30 but the display time for step 4 is 00:30 therefore it is showing the images correctly based on that but absolutely agree this is still an issue as it is inconsistent and when the user picks a date then the time should either be the first or last and should match not only date but also time. Thought this was worth mentioning as it might affect the approach to resolution - well spotted though!

pmbohm:
See https://github.com/aodn/aodn-portal/pull/2304

PR #2304, opened 2016-09-01
Notes on #2279 - NOT MERGEABLE

pmbohm:
I know -  I know - it points to the bigger issue of why is there no TimeSeries filter type


pmbohm:
This is just notes on where I think/thought the issues lay.
Lots of todo's to read or ignore

PR #2307, opened 2016-09-08
Fixes #2286 -traps errors and disables timeSelect buttons when extent is not loaded

### LINKED ISSUE #2286: Clicking 'Move Time on Map' before times are loaded creates fatal error
## To Reporduce
1. Add any ncWMS collection, (such as acorn)
2. Click either 'Move Time on Map' arrow before times are loaded

## What happens
1. Uncaught TypeError: Cannot read property 'clone' of null error on console
2. The times never load 

## What should happen 
'Move Time on Map' arrows are inactive untill the times load

pmbohm:

### DIFF
--- a/web-app/css/general.css
+++ b/web-app/css/general.css
@@ -1311,3 +1311,11 @@ input:read-only { background: #eee; }
 .spatialSubsetControl {
     padding: 15px 5px 15px 10px;
 }
+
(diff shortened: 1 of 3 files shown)


bhasin85:
Not working for me :hammer:

PR #2308, opened 2016-09-08
Added Download handler for differt file types

bhasin85:
# For testing, update the metadata record.

**URL:** _http://geoserver-123.aodn.org.au/geoserver/ows?typeName=aodn_mhl_sst_map&propertyName=url_

typeName is layer
propertyName is the database column name with the urls

**Protocol:** _AODN:FILE-list_

**Name of the resource:** _CSV_

**Description:** _The "File List" is a WFS service that returns a list of OpenDAP URLs matching a query._

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -313,7 +313,8 @@ portal {
         [ 'handler': 'NetcdfSubsetServiceDownloadHandler', 'protocol': 'OGC:WPS--netcdf-subset-service'      ],
         [ 'handler': 'PythonDownloadHandler',              'protocol': 'OGC:WFS-1.0.0-http-get-capabilities' ],
         [ 'handler': 'BodaacDownloadHandler',              'protocol': 'IMOS:AGGREGATION--bodaac'            ],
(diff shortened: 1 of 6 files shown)


julian1:
Looks good to me although I'm probably not the best person to review!.

mhidas:
@bhasin85 When I filled out the [metadata record](https://catalogue-rc.aodn.org.au/geonetwork/srv/en/metadata.show?uuid=60160b01-8ffc-45ce-a6f4-ee70ce391ec6) as per your example above, I got an error when trying this download option. @smancini pointed out that you actually need to put in a complete WFS request in the URL field, then it works.

So
```
URL = http://geoserver-rc.aodn.org.au/geoserver/ows?typeName=aodn_nsw_oeh_bathy_map&propertyName=url
```
gives an error, but
```
URL = http://geoserver-rc.aodn.org.au/geoserver/ows?service=WFS&request=GetFeature&typeName=aodn_nsw_oeh_bathy_map&propertyName=url&outputFormat=csv
```
works.

Is that as intended?

For future reference, I've added the details for this download option to the [Wiki](https://github.com/aodn/aodn-portal/wiki/Configuring-downloads-in-metadata-records). Please check that what I've put there is correct.

PR #2316, opened 2016-09-13
Add simple controller to trigger reload of external config files

lwgordonimos:
This enables the external groovy config file(s) to be reloaded without restarting Tomcat.
Currently if Portal.groovy (or any other hypothetical external config file) is updated but the WAR contents are not, the new configuration will not be effective until Tomcat is restarted.

Example usage: (from localhost *only*): 

```bash
[/var/log/tomcat8]# wget -q http://localhost:8080/ExternalConfigReload -O/dev/null

[/var/log/tomcat8]# tail catalina.out 
13-Sep-2016 05:54:00.447 INFO [main] org.apache.coyote.AbstractProtocol.start Starting ProtocolHandler ["http-nio-8080"]
13-Sep-2016 05:54:00.463 INFO [main] org.apache.coyote.AbstractProtocol.start Starting ProtocolHandler ["ajp-nio-8009"]
13-Sep-2016 05:54:00.467 INFO [main] org.apache.catalina.startup.Catalina.start Server startup in 18839 ms
2016-09-13 05:54:48,871 INFO  [http-nio-8080-exec-2] [] au.org.emii.portal.ExternalConfigReloadController  - Reloading config from file '/etc/aodn-config/Portal.groovy'...
```

### DIFF
--- a/grails-app/controllers/au/org/emii/portal/ExternalConfigReloadController.groovy
+++ b/grails-app/controllers/au/org/emii/portal/ExternalConfigReloadController.groovy
@@ -14,9 +14,15 @@ class ExternalConfigReloadController {
         def locations = config.grails.config.locations
 
         locations.each { location ->
-            String configFileName = location.split("file:")[1]
-            log.info "Reloading config from file '${configFileName}'..."
-            config.merge(new ConfigSlurper().parse(new File(configFileName).text))
+            String configFileName = location.replaceFirst(~/^file:/, '')
+            File configFile = new File(configFileName)
+
+            if (configFile.exists()) {
+                log.info "Reloading config from file '${configFileName}'..."
+                config.merge(new ConfigSlurper().parse(configFile.text))
+            } else {
+                log.debug "File for config location '${location}' doesn't exist."
+            }
         }
         render(status: 200, text: 'OK')
     }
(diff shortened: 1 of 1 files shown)


jonescc:
I'm looking at this.

grails-app/controllers/au/org/emii/portal/ExternalConfigReloadController.groovy
```diff
@@ -0,0 +1,30 @@
...
+    def grailsApplication
+
+    def index() {
+        def config = grailsApplication.config
+        def locations = config.grails.config.locations
+
+        locations.each { location ->
+            String configFileName = location.split("file:")[1]
+            log.info "Reloading config from file '${configFileName}'..."
+            config.merge(new ConfigSlurper().parse(new File(configFileName).text))
```
jonescc: Perhaps check that file exists before slurping (running this locally I get Config.local.portal does not exist at this point)

jonescc:
Doesn't look like any of the logging config can be changed this way - probably read by the logging framework on startup.  Pity.  Otherwise looks good

grails-app/controllers/au/org/emii/portal/ExternalConfigReloadController.groovy
lwgordonimos: I was just about to update the PR for this exact reason, since I just opened it in IntelliJ (which seems to inject that location by itself).

jonescc:
PS. the log4j stuff can't actually be chnaged using chef at the moment (https://github.com/aodn/chef/blob/master/cookbooks/imos_webapps/templates/default/portal/Portal.groovy.erb#L157) so shouldn't prevent this PR from being merged when you're done.

lwgordonimos:
@jonescc please see latest commit for additional File.exists() check.

PR #2319, opened 2016-09-14
Updated FileListDownloadHandler

bhasin85:


jonescc:
When you perform a download the zip file is always named the same no matter what download option you select (if you have multiple) e.g. {layer_name}_source_files.zip - perhaps should be named after sanitised  download option e.g. {layer_name}_Dive_Reports.zip  @pblain - any comments?

pblain:
@bhasin85 , @jonescc - that's a good idea.  Users won't want to open each zip file to find out which is which.

bhasin85:
Refer to [PR](https://github.com/aodn/aodn-portal/pull/2322)

PR #2322, opened 2016-09-15
Download options update

bhasin85:
@jonescc  Ready for review :+1:


jonescc:
Looks good.  

@pblain, just wondering what should be happening with the estimated download size here.  The estimated download size of different options here can be different but we are only displaying one at the bottom -  at the moment I think any download handler can be used to display the estimated download size so now it could actually be the estimated download size of something other than the unsubsetted netcdf files.  Should @bhasin85 disable using the non BODAAC options for estimating download size or cater for different destimated download sizes somehow?

jonescc:
Spoke with @pblain - we should turn off download estimation for FileListDownload handler (canEstimateDownbload should be set to false) for this iteration.

We'll need to sort out how to handle different estimates of download size per download option before we can switch from using BODAAC to the file list download handler later down the track

PR #2326, opened 2016-09-15
Download options update - Selenium Test

bhasin85:


selenium/src/test/java/au/org/emii/portal/tests/DownloadOptionTest.java
```diff
@@ -0,0 +1,48 @@
...
+        });
+
+        WebElement toDate = webElementUtil.findElement(By.className("end-date"));
+        toDate.clear();
+        toDate.sendKeys(new String[]{"1992/03/24"});
+        webElementUtil.clickButtonWithText("Next");
+
+        // Step 3
+        webElementUtil.clickButtonWithText(" Download as… ");
+        WebElement propertyOption = webElementUtil.findElement(By.xpath("//span[contains(.,'Property')]"));
```
jonescc: I don't think we can add this test just yet - we shouldn't be setting up test metadata like this in systest - its going to get wiped next time we do a harvest

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -266,11 +266,13 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
     _utcExtentDateTimePickerConfiguration: function(name) {
 
         var that = this;
+        var seleniumCls = name.replace(' ', '-');
```
jonescc: Are we going to try and distinguis selenium test support classes somehow e.g. prefix with selenium-

bhasin85: I think it is a good idea. I will make the change.

web-app/js/portal/search/SearchFiltersPanel.js
```diff
@@ -54,6 +54,7 @@ Portal.search.SearchFiltersPanel = Ext.extend(Ext.Panel, {
                 {
                     facetName: facet.key,
                     title: OpenLayers.i18n(facet.name),
+                    id: OpenLayers.i18n(facet.name),
```
pmbohm: Its not a good idea to set id's

Issue #2318, opened 2016-09-14
Step 3 does not render if accessed too quickly


kereid:
App version: 4.25.7 Systest
            Build date: Mon Sep 12 13:50:54 AEST 2016
            Build number: 71
            Git ref: dd65fa212b67f6a081fed068f319ee89ddd49021

Steps to reproduce:

Add 3 relatively big data sets using control click. ie 

SRS Satellite Contributed Ocean Colour - SeaWIFS Chlorophyll Concentration in the Southern Ocean: Monthly, Johnson et al 2013

CARS 2009 - Australian weekly

NOAA - Australasian Surface Drifting Buoys

When taken to step 2 click forward to step 3

What should happen:

Details of all collections added are shown 

What does happen 

Big white space like was seen in #2009

kereid:
@bhasin85 This may be a browser issue as I am using chrome to test this round but usually use Firefox and can still see it in firefox (in systest and also in prod)

PR #2338, opened 2016-09-23
Fix for issue #2318

bhasin85:


anguss00:
Looks good from the code side @bhasin85, but how did you test this? I'm having trouble replicating this issue locally

Issue #2321, opened 2016-09-14
Download size estimated twice after performing download


jonescc:
### To Reproduce

Select IMOS - SRS Satellite - SST L3S - 01 day composite - night time on step 1
Select a couple of days worth of data on step 2
On Step 3 select download and download file

### What happens

Two estimate download size requests are sent to the backend along with the download request and the estimated download size is updated

### What I expect to happen

The subset hasn't changed - there's no reason to estimate the download size again even once.

anguss00:
Halting work on this as it requires refactoring of download estimation on step 3 to fix (pair with @jonescc). See https://github.com/aodn/backlog/issues/448

anguss00:
"Fixed" by https://github.com/aodn/aodn-portal/pull/2347

PR #2339, opened 2016-09-23
Remove calls to regen step 3 content on download requested, started a…

anguss00:
…nd failed
Fixes issue #2321. I can't see anything that needs to redraw this page after the download operation. Would be keen for @jonescc to lend an eye too


jonescc:
The generatecontent call was added in https://github.com/aodn/aodn-portal/commit/bdfe7bd8a9bf02182272d4b368086325883aa5a6

It was added to update the status of the download for downloads that took a while to start.  Can we change it to just update the download status rather than refreshing the whole panel including estimating download size again?

anguss00:
Closing as download estimator has been removed

Issue #2317, opened 2016-09-13
Date format error remains when it shouldn't for manual date entered


kereid:
App version: 4.25.7 - Systest
            Build date: Mon Sep 12 13:50:54 AEST 2016
            Build number: 71
            Git ref: dd65fa212b67f6a081fed068f319ee89ddd49021

Steps to reproduce:

Add any collection (this was found on IMOS - ANMN National Reference Stations - Maria and North Stradbroke Islands - Near real-time wave height)

Select a from date i.e. 2016/01/01 and a to date 2016/01/02 realise there is not data and manually type change dates to later change from date i.e.  2016/08/07 see the red error boundary (which is correct at this point as August is after January) then change to date to 2016/08/09.

What should happen:

Red error box turns off as dates are in correct order

What does happen:

Red error box remains and if user hovers over the box the error reads "The date in this field must be equal to or before 2016/01/02

If you proceed to step three the correct dates are applied.

PR #2346, opened 2016-09-29
Validate from and to date fields on value change after updating minmax

anguss00:
Fixes issue #2317.
I've renamed a few functions to avoid confusion.
Allow blank fields as this defaults to min and max of the spatial extent for collection.


jonescc:
Still getting a problem here.

To reproduce:

Select IMOS - SRS Satellite - SST L3S - 01 day composite - night time and go to step 2
When dates have loaded copy end date (for me it was 29/9/2016 ) to start date and hit tab

What happens:

End date is marked as an error and error text says Start date (30th September) must be before end date (29th September)

What  expect to happen

Pasted date (29th September) is not considered to be the 30th September and highlighted as an error

jonescc:
OK - so thats a problem with gridded data - fix works OK for non-gridded data

Issue #2342, opened 2016-09-27
Error message for no data much longer than required


kereid:
App version: 4.25.9
            Build date: Thu Sep 22 13:15:18 AEST 2016
            Build number: 73
            Git ref: fcd27bff088ce4cb22731561ed1de45c3d9c16ff

Steps to reproduce:
Select any SRS SATELLITE - OC MODIS collection and add to the Portal.
Draw a bounding box away outside the extreme range of co-ordinates (ie outside the box).
Go to step 3 and select a netCDF download.

What should happen:

(Ideally user gets indication no data to download which will eventually come with estimate enhancements)

But for the last 2 months the error has been a simple one liner of:
Process failed during execution /mnt/ebs/geoserver_wps_systest/temp/wps/e6f67903-c4e2-49c1-919b-6e47d3104d2b/out/srs_sst_l3s_1d_day_gridded_url.nc must be netcdf-3 or netcdf-4 file 

What does happen:
User gets still uninformative error but much longer (not sure that it should have changed?):

[...]

PR #2356, opened 2016-10-07
Error message for no data

bhasin85:
Fixes [#2342](https://github.com/aodn/aodn-portal/issues/2342) 
Related [PR](https://github.com/aodn/geoserver-build/pull/176)

### DIFF
--- a/web-app/js/portal/cart/DownloadPanel.js
+++ b/web-app/js/portal/cart/DownloadPanel.js
@@ -121,7 +121,43 @@ Portal.cart.DownloadPanel = Ext.extend(Ext.Panel, {
 
         this._loadMenuItemsFromHandlers(processedValues, collection);
 
-        return this._applyTemplate(tpl, processedValues);
+        if (this._spatialSubsetIntersect(collection)) {
+            return this._applyTemplate(tpl, processedValues);
+        } else {
+            return this._emptyMsg();
+        }
+    },
+
+    _spatialSubsetIntersect: function(collection) {
+
+        var intersect = true;
+        var filters = collection.getFilters();
+
+        if (filters) {
+            var params = filters.filter(function(filter) {
+                if (filter.isNcwmsParams || filter.type === 'geometrypropertytype') {
+                    return true;
+                }
+            })[0];
+
+            if (params && params.isNcwmsParams && params.latitudeRangeStart != undefined) {
+                var bounds = collection.getBounds();
+                var extent = new OpenLayers.Bounds(params.longitudeRangeStart, params.latitudeRangeStart, params.longitudeRangeEnd, params.latitudeRangeEnd);
+                intersect = bounds.containsBounds(extent, true, true);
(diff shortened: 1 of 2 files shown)


grails-app/controllers/au/org/emii/portal/WpsController.groovy
```diff
@@ -63,7 +64,7 @@ class WpsController extends HostVerifyingController {
                 job: [
                     uuid: params.uuid,
                     reportUrl: g.createLink(action: 'jobReport', absolute: true, params: params),
-                    createdTimestamp: new DateTime(String.valueOf(execResponse.Status.@creationTime)),
+                    createdTimestamp: new DateTime(String.valueOf(execResponse.Status.@creationTime)).withZone( DateTimeZone.forID( "Australia/Melbourne" )),
```
julian1: this commit, shoudln't be included here, if it's already in the other PR https://github.com/aodn/aodn-portal/pull/2356/commits/3501a7a8c05069ec7096bb0268ae353a27d27e7c

bhasin85: @julian1  I think I created the branch from issue_2332 instead of master and therefore all the changes got copied. Please review the changes.

web-app/js/portal/cart/DownloadPanel.js
```diff
@@ -121,7 +121,34 @@ Portal.cart.DownloadPanel = Ext.extend(Ext.Panel, {
...
 
-        return this._applyTemplate(tpl, processedValues);
+        if (this._spatialSubsetIntersect(collection)) {
+            return this._applyTemplate(tpl, processedValues);
+        } else {
+            return this._emptyMsg();
+        }
+    },
+
+    _spatialSubsetIntersect: function(collection) {
```
julian1: If I understand the code correctly, then the subsetintersect is only being applied for ncwms. I wonder if the method name should be changed to reflect that? 

Also, why is data being returned that is outside the bound of interest?

julian1: For me there are a couple of issues.
  - should the check be performed for all data sets and protocols - not only ncwms? If not then the naming should be changed to reflect that.
  - should the extent calculation be being done in a view class? For example if the predicate behavior was factored out then it would be available to show a pop-up at stage 2 for instance. Just an example, not saying we should do this. In any case I feel this is more model code, that view code.

julian1: @jonescc just pointed out that if someone uses the service directly eg. marvl, then they should also get a sensible error message, which they won't if they bypass the portal.

julian1:
Hey @bhasin85  I wasn't saying that this test needed to be done for all collections? It was just a question, since I don't know if we do.

bhasin85:
I think it was a good suggestion to do the similar test wherever possible. I have tried to cover the relevant collections I can think of.

giphy

PR #2358, opened 2016-10-07
Adjusted timezone 

### LINKED ISSUE #2332: GoGoDuck status page showing not in Australian time Zone?
Seeing in Prod and Systest

Steps to reproduce:

Create a small gogoduck job on status page look at time commenced.

What should happen:

It's in Australian time

What does happen:

It's showing as some other timezone

ie 1249pm Australian time job started shows as 2:49am on status page

This has only changed since I completed testing because I spoke to @anguss00 last Friday (16/09/2016) about a csv point job that I started at 817 and was still going at 10am and know I saw 0817 on the status page

bhasin85:
Fixes #2332

### DIFF
--- a/grails-app/views/wps/show.gsp
+++ b/grails-app/views/wps/show.gsp
@@ -20,7 +20,14 @@
 
                 <g:labelledContent labelCode="job.id.label" href="${job.downloadUrl}">${job.uuid}</g:labelledContent>
                 <g:labelledContent labelCode="job.createdTimestamp.label" if="${job.createdTimestamp}" >
-                  <joda:time value="${job.createdTimestamp}" />
+                    <label id="localTime"></label>
+                    <g:javascript>
+                       if('${job.createdTimestamp}')
+                       {
+                            var date = new Date('${job.createdTimestamp}');
+                            document.getElementById('localTime').innerHTML = date;
+                       }
+                    </g:javascript>
                 </g:labelledContent>
                 <g:labelledContent labelCode="job.status.label">
                   <g:message code="job.status.${job.status}" default="${job.status.toString()}" />


grails-app/controllers/au/org/emii/portal/WpsController.groovy
```diff
@@ -63,7 +64,7 @@ class WpsController extends HostVerifyingController {
                 job: [
                     uuid: params.uuid,
                     reportUrl: g.createLink(action: 'jobReport', absolute: true, params: params),
-                    createdTimestamp: new DateTime(String.valueOf(execResponse.Status.@creationTime)),
+                    createdTimestamp: new DateTime(String.valueOf(execResponse.Status.@creationTime)).withZone( DateTimeZone.forID( "Australia/Melbourne" )),
```
julian1: hard-coding the timezone doesn't seem right to me. wouldn't this be a problem for niwa if their using the portal?

bhasin85: @julian1  I agree it will be a problem. I have moved the timezone to the config file. :+1:

julian1: What happens if a portal-end user is outside australia/melbourne? will setting the timezone explicitly change any display behavior on the client side?

julian1: The server shouldn't be hard-coding the timezone for the client. In the past we always invested the effort to fix time/zone formatting issues - and without resorting to coding the behavior on the server. So there is already a precedent in terms of effort that's been invested in avoiding this kind of change. @jonescc any thoughts?

julian1: Just to note, javascript dateTimes native-representation is utc/zulu time. So, if the server always sends UTC, the client only needs to use toLocaleString() or similar to apply the client's locale to extract the correct timezone formatted value.
```
d = new Date()
Date 2016-10-13T23:21:57.532Z
d.toUTCString()
"Thu, 13 Oct 2016 23:21:57 GMT"
d.toLocaleString()
"14/10/2016, 10:21:57 am"
```

julian1:
Hi, what happens if the client is in WA, or SA or NT, or NZ, or anywhere outside NSW DST?

bhasin85:
@julian1  I have updated the logic to retrieve the Timezone from the browser and pass it to the server.

julian1:
@bhasin85 are you sure this is necessary? I don't understand why the server can't just send the time in UTC format. Then when it's displayed on the client you just call toLocalDateTime() in js to format it correctly according to the timezone of the user.

bhasin85:
@julian1  Please review the changes.

PR #2382, opened 2016-11-03
fixes #2378 - unified error messages in download panel

### LINKED ISSUE #2378: Out of bounds CSV at a point not shown on step 3
### To reproduce

In systest:

Select a gridded dataset such as ACORN 
Select Point timeseries
Set Point time series location to a location outside the bounding box of the collection
### What happens

lat/lon value out side of range is highlighted in red with maximum allowed value out of range error (good)
when I go to step 3 the spatial subset defaults to the non-point timeseries spatial subset which is confusing
### What I expect to happen

Get a point timseries location outside the spatial extent of the dataset message similar to what I would get for non-Point time series subsets outside the spatial extent

pmbohm:


bhasin85:
Nice Interface! @pmbohm 

giphy

PR #2387, opened 2016-11-10
Fix selenium tests

jonescc:
Fix various issues with selenium tests:

- only run a maximum 2 tests at a time
- wait for feature popup to be populated before validating lat/lon returned
- wait for new tabs to be displayed before trying to switch to them
- wait for search results to be loaded again after selecting a facet before trying to click the 'Add collection button'
- select a non-gridded collection for testing no subset selected warning
- set browser position and size to reflect requested resolution
- use package instead of classes to specify tests to run so that we don't have to update testng.xml everytime we add a new test
- make it easier to see what test each browserstack session relates to by using classname instead of browser/device etc as the name (browser/device etc info is is already reported elsewhere).
- log test run info only once and full test details for each test
- remove test which wasn't being run/is not required


selenium/src/test/java/au/org/emii/portal/PortalUtil.java
```diff
@@ -0,0 +1,17 @@
+package au.org.emii.portal;
```
bhasin85: @jonescc  Do you think we can move this under package _au.org.emii.portal.utils_

jonescc: No problem.

PR #2402, opened 2016-12-14
Fixes #2399 Expand/collapse of Step 1 items which have large amount of text in descriptions

### LINKED ISSUE #2399: Second click of grey button hides collection info on step 1
Steps to reproduce:

Add any collection, go back to step 1 and click the select button again for the same collection then go back to step 1 again.

What should happen:

Still see all the details like geographic extent and parameters etc.

What does happen:

Only see title and select button for collection

TBH I think this has been around for a fair while I just never realised why it happened.

pmbohm:


bhasin85:
@pmbohm  Somehow I am still able to reproduce the issue.

bhasin85:
@pmbohm  Looks Good. Browser caching issue.

PR #2403, opened 2016-12-14
Issue 2400 fixed

pmbohm:
urlListDownloadSubstitutions updated for development.
Tested with zip download with .../search?uuid=60160b01-8ffc-45ce-a6f4-ee70ce391ec6

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -178,9 +178,7 @@ knownServers = [
         type: 'GeoServer',
         csvDownloadFormat: 'csv-with-metadata-header',
         urlListDownloadSubstitutions: [
-            '^/mnt/imos-t3/': 'http://data.aodn.org.au/',
-            '^/mnt/opendap/2/SRS': 'http://thredds.aodn.org.au/thredds/fileServer/srs',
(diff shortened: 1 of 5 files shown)


lwgordonimos:
Reviewing/testing now...

lwgordonimos:
All looks to work as expected with random tests, but the only question/concern I would have due to the behaviour change:

Are we 100% certain that nothing is relying on previous assumptions? This assumption works fine for *this* particular collection, but is there anything else that may rely on the previous assumption, e.g.
```
file1.nc
file1.txt 
```
used to result in:
```
file1.nc
file1(1).txt
```
but now results in
```
file1.nc
file1.txt
```

As innocent and common sense as it appears, it does change the behaviour for existing unrelated files! Can we identify if any other collections have files with multiple extensions?

@mhidas Do you know of any to specifically test?

mhidas:
@lwgordonimos I don't know of any other collections that have this issue. 

If anything *was* relying on files to be renamed at download time, I would suggest this is far from ideal and should be re-thought. If we put in this fix we'll find out whether anything did rely on that, but I doubt very much there will be anything.

As far as I know, this is only affecting the names of the files delivered to the user. Nothing changes to files within our workflow. Right?

PR #2406, opened 2016-12-21
Issue 2395

pmbohm:
When a user loads a ncWMS layer and goes quickly to step 3, the page indicate that the Temporal extent filters are loading, and the download options are disabled during this time. This happens even on fast internet connections such as UTAS.

- Now the download panel (step 3) will reload when temporal filters finish loading, in this way the temporal filter should be correct before the user can choose a download option if the download option button is enabled that is.
- The user will see a message when the temporal extent matches the full extent of the chosen layer of the collection. This message is advisory and the download button is enabled

### DIFF
--- a/web-app/js/portal/cart/DownloadPanelItemTemplate.js
+++ b/web-app/js/portal/cart/DownloadPanelItemTemplate.js
@@ -57,10 +57,9 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
         if (values.errorMessage && values.errorMessage != "") {
             msg = values.errorMessage;
         }
-        else if (values.dataFilters == "") {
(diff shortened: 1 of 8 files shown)


web-app/js/portal/cart/DownloadPanelItemTemplate.js
```diff
@@ -54,13 +54,13 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
     _getUserMsg: function(values) {
         var msg = "";
 
+
```
anguss00: OCD going off on the extra whitespace

anguss00:
Aside from one dumb comment.. Looks good, performs perfectly

pmbohm:
@anguss00 spaces removed

PR #2407, opened 2016-12-22
Allow use of sharelinks without params

pmbohm:


pmbohm:
Supporting work for https://github.com/aodn/backlog/issues/390, creating a site map that will be indexed by google.
Google prefers urls without parameters so this pull request allows share links to be used like:
https://portal.aodn.org.au/search/uuid/aaad092c-c3af-42e6-87e0-bdaef945f522
The portal will still support (and advertise) share links like:
https://portal.aodn.org.au/search?uuid=aaad092c-c3af-42e6-87e0-bdaef945f522

PR #2415, opened 2017-01-24
fixes #2409 

### LINKED ISSUE #2409: Warning message still displayed after bounding box drawn
SYSTEST
App version: 4.27.7
           Build date: Tue Jan 17 09:35:54 AEDT 2017
            Build number: 89
            Git ref: 4bcc644805a93701ad3756bb4cc095903291f61a

Steps to reproduce:

Add any gridded collection do not apply any filters procesed to step 3 - see message "The full data collection will be downloaded. Please consider filtering the collection."

Go back to step 2 and draw a bounding box, go back to step 3

What should happen:

Message is gone (can see it will go if user changes date on step 2)

What does happen:

Although user can see there is spatial subset applied there is also the warning message still displayed

pmbohm:
Separate download panel message where temporal extent was not filtered by user

### DIFF
--- a/web-app/js/portal/cart/DownloadPanelItemTemplate.js
+++ b/web-app/js/portal/cart/DownloadPanelItemTemplate.js
@@ -57,10 +57,16 @@ Portal.cart.DownloadPanelItemTemplate = Ext.extend(Ext.XTemplate, {
         if (values.errorMessage && values.errorMessage != "") {
             msg = values.errorMessage;
         }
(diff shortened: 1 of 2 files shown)


pmbohm:
@bhasin85 Sure this can be discussed further. Keep in mind that we are encouraging people to subset more than just temporally and spatially. Also if a user uses neither temporal or spatial filter but filters with another variable, the result set is most of the time much much smaller, making any further helpful messages appear junky

bhasin85:
Confirmed the warning messages to display with @pblain .

PR #2418, opened 2017-01-27
Fixes #2410

### LINKED ISSUE #2410: Passive Acoustic collection causes havoc
SYSTEST
App version: 4.27.7
            Build date: Tue Jan 17 09:35:54 AEDT 2017
            Build number: 89
            Git ref: 4bcc644805a93701ad3756bb4cc095903291f61a
        -->
Steps to reproduce:

Add any collection i.e ANMN Current velocity time series draw a bounding box
Go back to step 1 and use keyword and type "passive" add passive acoustic collection, once on step 2 proceed to step three and then also back to step 1 and add any other collections

What should happen:

On Step 3 see all options for download, on step one add collections and be taken to step 2

What does happen:

There is a big blank white space on step 3 and step one when adding collections not take through to step 2 (Like holding control and clicking)

akashisama:


pmbohm:
Should you apply the new logic to `_checkPointIntersectsBounds` as well?

PR #2423, opened 2017-01-31
Dynamic sitemap from metadata server

pmbohm:


grails-app/views/robots.gsp
```diff
@@ -1,4 +1,4 @@
 <%@ page contentType="text/plain;charset=UTF-8" %>
 User-agent: *
 Allow: /
-Sitemap: https://portal.aodn.org.au/sitemap.xml
+Sitemap: https://portal.aodn.org.au/sitemap/sitemap.xml
```
bhasin85: How about adding conditions based on environment?

```
<g:if env="production">
    <%@ page contentType="text/plain;charset=UTF-8" %>
    User-agent: *
    Allow: /
    Sitemap: ${grailsApplication.config.grails.serverURL}sitemap/sitemap.xml
</g:if>
```

grails-app/conf/UrlMappings.groovy
```diff
@@ -1,10 +1,9 @@
-import grails.util.Environment
-
 class UrlMappings {
 
     static mappings = {
 
         "/proxy"(controller: "proxy", action: "index")
+        "/sitemap/**"(controller: "sitemap", action: "index")
```
bhasin85: This is not related to this PR, but do you think it will be better to do the redirection here? i.e. search/uuid/something to search?uuid=something&info=true

grails-app/views/robots.gsp
pmbohm: yep i meant to do that after our discussion didnt I :(

grails-app/conf/UrlMappings.groovy
pmbohm: Thats needed as we want to get to the index action yet we want the URL asking for sitemap.xml

bhasin85: Sorry @pmbohm  .... I mean is it possible to do the following redirection here instead of SearchController? If yes, then will it make more sense?

search/uuid/something to search?uuid=something&info=true

pmbohm: What your talking about was going to be changed in https://github.com/aodn/aodn-portal/pull/2407. But its merged!
I'll add a seperate pull request if a transparent redirect can be acheived

grails-app/views/robots.gsp
```diff
@@ -1,4 +1,4 @@
 <%@ page contentType="text/plain;charset=UTF-8" %>
 User-agent: *
 Allow: /
-Sitemap: https://portal.aodn.org.au/sitemap.xml
+Sitemap: ${grailsApplication.config.grails.serverURL.replaceFirst('/$',"")}/sitemap/sitemap.xml
```
bhasin85: @pmbohm  We can easily tell the robot that it should not visit any pages on the site by using the following. I think this is a better place than chef. 

```
    <%@ page contentType="text/plain;charset=UTF-8" %>
    User-agent: *
<g:if env="production">
    Allow: /
    Sitemap: ${grailsApplication.config.grails.serverURL}sitemap/sitemap.xml
</g:if>
<g:else>
    Disallow: /
</g:else>
```

pmbohm: I think the environment is always production except in an IDE

bhasin85: @pmbohm  You are right! :+1:

PR #2428, opened 2017-02-07
Create local PagingToolbar.js so that the refresh icon can be changed

### LINKED ISSUE #2422: Refresh button on step one has no purpose
Steps to reproduce:

On step one add any collections to map go back to step one select random facets then scroll down to bottom of results page and click refresh button

What happens:

User taken back to top of results page

What should happen:

Not sure, unsure of what this button is meant to do but what is the point of having a refresh button that only takes user to top of results page, not even back to 1st page of results if there are more than one page and user is on any other page but the first....

pmbohm:
fixes #2422


akashisama:
Face-to-face review suggestions were: 
-Add comment to PagingToolbar specifying that it is copied from Ext.PagingToolbar
- Don't set tool-tip text twice

PR #2438, opened 2017-03-01
Change step 1 search on top

pmbohm:
See aodn/backlog#536


pmbohm:
@akashisama I removed the requirement of non empty strings, so no error messages any more. Also the use of `togglecollapse` was wrong and created a bug which i discovered along the way

pmbohm:
@akashisama Fixed up my commit

PR #2443, opened 2017-03-07
Sitemap Url as text file

pmbohm:
The sitemap had the XML extension instead of TXT

### DIFF
--- a/grails-app/views/robots.gsp
+++ b/grails-app/views/robots.gsp
@@ -1,4 +1,4 @@
 <%@ page contentType="text/plain;charset=UTF-8" %>
 User-agent: *
 Allow: /
-Sitemap: ${grailsApplication.config.grails.serverURL.replaceFirst('/$',"")}/sitemap/sitemap.xml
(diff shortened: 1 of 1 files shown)


pmbohm:
From https://support.google.com/webmasters/answer/183668?hl=en
```
Guidelines for text file sitemaps:

Encode your file using UTF-8 encoding.
Your text file should contain nothing but the list of URLs.
**You can name the text file anything you wish, provided it has a .txt extension (for instance, sitemap.txt)**.
```

pmbohm:
https://github.com/aodn/backlog/issues/390

akashisama:
Should we avoid putting this change into production until we're sure we've got the urls are correctly setup for google to be happy? @pblain

pblain:
Yes, please revert.  Or Julian's suggestion of preventing Google from finding the sitemap is also OK. The main thing is to avoid being black flagged by Google. It takes a long time and a lot of work to recover once that happens.

PR #2444, opened 2017-03-07
Revert "Sitemap Url as text file"

jonescc:
Reverts aodn/aodn-portal#2443

Sounds like this needs to wait for input from @pblain.


julian1:
If we don't want Google to index - then why not remove all references to sitemap?  

Flipping the filename extension between txt and xml on an assumption that Google will index one and not the other is pre-supposing a lot of knowledge about how Google works.

PR #2446, opened 2017-03-13
Download shape file

bhasin85:
For testing search for record with UUID **c1344e70-480e-0993-e044-00144f7bc0f4** in portal

I have manually added shape file online resource for the following record.

http://catalogue-systest.aodn.org.au/geonetwork/srv/eng/metadata.show?uuid=c1344e70-480e-0993-e044-00144f7bc0f4

Transformation update for geonetwork record with UUID c1344e70-480e-0993-e044-00144f7bc0f4 for systest
Related [PR](https://github.com/aodn/schema-plugins/pull/39/files)

### DIFF
--- a/web-app/js/portal/cart/ShapeFileDownloadHandler.js
+++ b/web-app/js/portal/cart/ShapeFileDownloadHandler.js
@@ -1,0 +1,65 @@
+Ext.namespace('Portal.cart');
+
+Portal.cart.ShapeFileDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
+
+    getDownloadOptions: function(filters) {
+
+        var downloadOptions = [];
+
+        if (this._showDownloadOptions(filters)) {
+
+            downloadOptions.push({
+                textKey: 'downloadAsShapeFileLabel',
+                type: 'WFS',
+                handler: this._getUrlGeneratorFunction(),
+                handlerParams: {
+                    filenameFormat: '{0}_'+this.onlineResource.name+'.zip',
+                    downloadControllerArgs: {
+                        action: 'downloadShapeFilesForLayer'
+                    }
+                }
+            });
+        }
+
+        return downloadOptions;
+    },
+
+    _showDownloadOptions: function(filters) {
+
+        return this._resourceHrefNotEmpty()
+            && this._resourceNameNotEmpty()
+            && !Portal.filter.FilterUtils.hasFilter(filters, 'timeSeriesAtPoint');
+    },
+
+    _getUrlGeneratorFunction: function() {
+
+        var _this = this;
+
+        return function(collection) {
+
+            var builder = new Portal.filter.combiner.ShapeFileCqlBuilder({
+                filters: collection.getFilters()
+            });
+
+            return OpenLayers.Layer.WMS.buildGetFeatureRequestUrl(
+                _this._baseUrl(),
+                _this._layerName(),
+                _this._outputFormat(),
+                builder.buildCql()
+            );
+        };
+    },
+
+    _layerName: function() {
+        return Ext.urlDecode(this.onlineResource.href.split("?")[1]).typeName;
+    },
+
+    _outputFormat: function() {
+        return Ext.urlDecode(this.onlineResource.href.split("?")[1]).outputFormat;
+    },
+
+    _baseUrl: function() {
+        return this.onlineResource.href.split("?")[0];
+    }
+
+});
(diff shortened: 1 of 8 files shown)


bhasin85:
@pmbohm  Testing can also be done agains IMAS record **9d6b28d2-2d6b-4726-b603-ab03e3d15d34**

web-app/js/portal/cart/ShapeFileDownloadHandler.js
```diff
@@ -0,0 +1,65 @@
...
+Portal.cart.ShapeFileDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
+
+    getDownloadOptions: function(filters) {
+
+        var downloadOptions = [];
+
+        if (this._showDownloadOptions(filters)) {
+
+            downloadOptions.push({
+                textKey: this.onlineResource.name,
```
pmbohm: The label should come from en.js so that we get a uniform look in the drop down. 
See Example: https://github.com/aodn/aodn-portal/blob/42896f2cebb2b50b789fd207df703b46f1ea04ea/web-app/js/portal/cart/WfsDownloadHandler.js#L12

screen shot 2017-03-15 at 2 24 21 pm

web-app/js/portal/prototypes/OpenLayers.js
```diff
@@ -119,6 +119,7 @@ OpenLayers.Layer.WMS.buildGetFeatureRequestUrl = function(baseUrl, layerName, ou
     wfsUrl += '&outputFormat=' + outputFormat;
     wfsUrl += '&REQUEST=GetFeature';
     wfsUrl += '&VERSION=1.0.0';
+    wfsUrl += '&maxFeatures=50';
```
pmbohm: Is this enough?

web-app/js/portal/filter/combiner/ShapeFileCqlBuilder.js
```diff
@@ -0,0 +1,14 @@
+Ext.namespace('Portal.filter.combiner');
+
+Portal.filter.combiner.ShapeFileCqlBuilder = Ext.extend(Portal.filter.combiner.FilterCqlBuilder, {
+
+    buildCql: function() {
+
+        var cqlParts = this._visualisedFiltersWithValues().map(function(filter) {
+
+            return filter.getCql();
```
pmbohm: Try this with a spatial filter
I think it needs to be URL encoded?

web-app/js/portal/cart/ShapeFileDownloadHandler.js
bhasin85: Made the change as requested after discussing with @pblain

web-app/js/portal/prototypes/OpenLayers.js
bhasin85: Yes... this was needed parameter for the shapefile request

web-app/js/portal/filter/combiner/ShapeFileCqlBuilder.js
bhasin85: The url was getting encoded but it was failing because of a trailing slash. Should be fine now. :+1:

bhasin85:
@pmbohm  Ready for review round 2 :boxing_glove:

bhasin85:
@pmbohm  Removed the **zero width space**

pmbohm:
@bhasin85 Broke a test?

web-app/js/portal/prototypes/OpenLayers.js
julian1: why 50 as limit?

bhasin85: That is what the default geoserver value is for maxFeatures.

jonescc: Yes this is what it defaults to through the UI so that a user doesn't inadvertently ask to download the whole collection, but its not what we would want to use in the portal.   The maxfeatures parameter limits the number of records (features) returned in the response to the specified number i.e. it truncates the response.  We've just recently upped our server limit from 100000 records because that was too small so 50 records is going to drastically reduce this.

pmbohm: @bhasin85 Will you up this to 1,000,000,0000 please in another pull request?

jonescc: Better not to specify it here at all.  We are configuring this as a global setting in the server configuration.

bhasin85: Thanks @pmbohm  @jonescc  @julian1 

I have created another [PR](https://github.com/aodn/aodn-portal/pull/2450) based on your suggestions :+1:

PR #2449, opened 2017-03-16
Support proxy redirection and eliminate tile gutter

julian1:
Review but don't merge, since the gutter change will break squid caching (I think).

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -170,6 +170,16 @@ baselayerServer = [
     wmsVersion: '1.1.1'
 ]
 
+
+// overwritten by configuration management
+proxyRedirects = [
+    [
+        uri: 'http://mywms-server.aodn.org.au/geoserver',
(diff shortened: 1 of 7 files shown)


anguss00:
I'm a bit worried about groovy following the redirect in the external request that's being made, as its using Java HttpUrlConnection to make the external request in https://github.com/aodn/aodn-portal/blob/master/src/groovy/au/org/emii/portal/proxying/ExternalRequest.groovy#L35. AFAIK HttpUrlConnection doesn't automatically follow the redirect... or am I missing something?

julian1:
Could you explain in more detail. Note that this code doesn't change any of the existing proxying. It merely changes the textual url that we proxy to.

anguss00:
Yes sorry that makes sense. My comment above is coming from the perspective of providing a URL that has a redirect attached to it (that is, a url that chucks a 302). If URL redirection is being handled by changing the actual url that we proxy to then it should be fine.

PR #2452, opened 2017-03-17
Get rid of unneeded spaces in EPSG parameters of getfeatureinfo requests

akashisama:
(required change for geowebcache comparability)


pmbohm:
It turns out that this creates https://github.com/aodn/aodn-portal/issues/2461

PR #2453, opened 2017-03-20
Selenium test to make sure shapefile option is shown in download tab f…

akashisama:
…or relevant layers

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/step3/ShapeFileTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/step3/ShapeFileTest.java
@@ -1,0 +1,29 @@
+package au.org.emii.portal.tests.step3;
+
+import au.org.emii.portal.tests.BaseTest;
+import org.apache.log4j.Logger;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+public class ShapeFileTest extends BaseTest {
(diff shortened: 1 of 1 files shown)


selenium/src/test/java/au/org/emii/portal/tests/step3/ShapeFileTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.List;
+
+public class ShapeFileTest extends BaseTest {
+
+    private static Logger log = Logger.getLogger(ShapeFileTest.class.getName());
+
+    @Test
+    public void verifyShapeFile() {
+        getDriver().get(AODN_PORTAL_SEARCH_PAGE + "?uuid=9d6b28d2-2d6b-4726-b603-ab03e3d15d34");
```
anguss00: Hmm. May not be a biggie, but if this uuid dies for whatever reason, the test dies with it

anguss00: Dunno how we get around that though...

akashisama: Yeah there's no way around this for now. We can't find out via a selenium test which collections should have a shapefile, so we have to just pass it one manually.

anguss00: Yup. Absolute worst case, test breaking is an alarm bell for this record being in a bad state

PR #2456, opened 2017-03-22
Add GA tracking on free text search

anguss00:
Add events for click on magnifier and enter to search. 

Please comment on the event label wording if not appropriate @pblain


web-app/js/portal/search/FreeTextSearchPanel.js
```diff
@@ -23,14 +23,29 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
...
                         { xtype: 'spacer', width: 10 },
                         this.searchField = new Ext.form.TextField({
                             width: 250,
                             focusClass: "focused",
                             emptyText: OpenLayers.i18n('freeTextFilterSearchHelperMsg'),
                             fieldLabel: 'Name',
+                            listeners: {
+                                specialkey: function(field, e) {
+                                    if (e.getKey() == e.ENTER) {
+                                        trackFacetUsage(this.facetName, OpenLayers.i18n('facetTrackingFreeTextEnterLabel'));
```
akashisama: Is `facetName` accessible from here? When I replace `trackFacetUsage` with `console.log`, I get "undefined"

anguss00:
That's a good point @akashisama. I'll rework that part, because I agree it doesn't provide a useful comparison. These events were structured based around analysing the event flow in GA, where we'd be able to see the number of times the user has typed in the box and pressed enter vs the number of clicks on the magnifier. But I'm thinking this kind of behavioural analysis might be better served by hotjar rather than GA...

anguss00:
And on the second point, it may muddy the results for multiple clicks (unless we consider the event flow itself for an individual session). This is why I'm thinking hotjar is more suitable for this problem. What do you think @pblain?

web-app/js/portal/search/FreeTextSearchPanel.js
anguss00: Sorry about that... will fix!

pblain:
Just to be clear, the requirement is to count the number of times a user types in the box and clicks enter versus the number of times the user types in the box and then performs some other action. 

It would also be interesting to track clicks on the magnifying glass, noting that clicking on the magnifying glass is a particular case of "some other action". 

The other ideas are definitely interesting but my main concern, for now, is with the above.

pblain:
@anguss00 - I noticed, unrelated to this, that there is an event action named "freetextFilter".  Could you please change to "Keyword" so that it's consistent with the other action names?

I don't think the tracking stats in this PR belong under the "Facets" category.  We are tracking facet activity, but If you take a look in GA you'll see what I mean.  The Facets category is tracking, for each facet, what the users are searching on. Adding usability stats will be inconsistent with everything else under the category.

This is essentially a usability test, which may be short lived if we change the UI based on our findings.  For this reason I think we need a new Category for this type of experimental tracking. Perhaps name the category "UsabilityTests"; name the action "Keyword submit"; name the labels "Enter", "Magnifying glass" and "Other".

anguss00:
So at this stage, I'm not sure how we can accurately track the "Other" action category with our current implementation. The problem is, the blur (focus loss) event doesn't pass the event itself, making it impossible to identify where the event has been called from. This is going to result in lots of false positives that we just can't account for AFAIK. For instance, user enter text and clicks enter. Triggers the enter keydown event. Then clicks a collection in step 2. That second click causes a focus loss that I can't identify the source of (so can't filter out of GA stats). 
Another example, user clicks text field, enters value, then selects another facet. This will show up as a focus loss (i.e no enter action) BUT this isn't a usability test failure. 
Open to thoughts/opinions on this one, but I think the only reliable stats we can get are enter actions vs clicks on the magnifier.

anguss00:
Put up changes to GA wording

web-app/js/portal/search/FreeTextSearchPanel.js
```diff
@@ -67,20 +77,24 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
...
+            this.searchField.reset();
         }, this);
     },
 
     initComponent: function() {
         Portal.search.FreeTextSearchPanel.superclass.initComponent.apply(this, arguments);
     },
 
+    onStateSave: function(field, state) {
+        console.log(state);
```
akashisama: What's this for?

pblain:
Can we track the following?
(a) enter actions
(b) magnifying glass clicks
(c) keyword text box _gets_ focus events

This should be enough to give a reasonable picture of what's going on. 
(Hopefully a + b is somewhere in the vicinity of c)

web-app/js/portal/search/FreeTextSearchPanel.js
anguss00: My bad! Will fix that

anguss00:
Ok, changes up

akashisama:
Yay merged, now to wait for the stats to pour in.

anguss00:
Sweeet!!! All of the stats, all the time ;-)

PR #2462, opened 2017-03-29
New selenium test, check that the step 2 and 3 buttons begin disabled

akashisama:


selenium/src/test/java/au/org/emii/portal/tests/step1/StepsDisabledTest.java
```diff
@@ -0,0 +1,40 @@
...
+    private static Logger log = Logger.getLogger(ShapeFileTest.class.getName());
+
+    @Test
+    public void stepsDisabledTest() throws InterruptedException{
+        WebDriver driver = getDriver();
+        driver.get(AODN_PORTAL_SEARCH_PAGE);
+
+        List<WebElement> buttons = webElementUtil.findElements(By.className("viewPortButton"));
+
+        buttons.get(1).click();
```
jonescc: Would it be better to  match on button text here to make it clearer what button is being selected?  I had to go to the portal and use the debugger to work out exactly what button was being pressed.

akashisama:
Fair point, changes made. Feel free to squash the review changes commit into the other one if/when you merge .

PR #2458, opened 2017-03-22
Upgrade to JQuery3

pmbohm:
@bhasin85 I need a hand to fix a clash with Selinium!

PR #2463, opened 2017-03-31
Fixed various issues introduced by JQuery upgrade.

akashisama:
Some events weren't being run due to JQuery3 handling events differently.

JQuery3 Upgrade (see #2458) predictably caused a bunch of issues:
- Mouse over title text on the last level facet menu items wasn't working
- Mouse over title text on "pin" icon of getfeatureinfo wasn't working
- FeatureInfo hyperlink clicks were not being tracked
- Search results which were too long to fit in regular size couldn't be expanded

There were two events being added in jquery.js (lines 32 and 65 of the unrefactored file) which I couldn't for the life of me work out what they were supposed to do. I therefore removed them as a I strongly suspect they're remnants of long deceased features.

### DIFF
--- a/web-app/js/portal/jquery.js
+++ b/web-app/js/portal/jquery.js
@@ -1,7 +1,7 @@
 jQuery( document ).ready(function() {
 
-    jQuery(".resultsHeaderBackground:not(.facetedSearchBtn *)").on("click",
-        function(){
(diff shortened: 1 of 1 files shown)


akashisama:
This is waiting for a review. Won't be able to put the latest iteration into production without it.

pmbohm:
Nice one @akashisama

PR #2477, opened 2017-04-28
Add selenium test for clearing a facet using a checkbox

akashisama:


selenium/src/test/java/au/org/emii/portal/tests/step1/FacetTest.java
```diff
@@ -65,6 +66,68 @@ public void parameterPhysicalAtmosphereTest() throws InterruptedException {
...
+    public void untickFacetTest() throws InterruptedException {
+        log.info("Loading search page - Step 1");
+        WebDriver driver = getDriver();
+        driver.get(AODN_PORTAL_SEARCH_PAGE);
+
+
+        // grab the text of the original results so we can make sure it reverts back when we untick later
+        String originalResults = webElementUtil.findElement(By.className("faceted-search-results")).getText();
+
+        WebElement topPanel = webElementUtil.findElement(By.className("search-filter-panel"));
```
jonescc: This code is repeated here three times.  Would it be better to use a helper method to reduce code duplication?

akashisama: I feel like that would add extra clutter given the code would be unlikely to be used again outside of this method.

jonescc: Removing repeated code - reduces clutter.

PR #2485, opened 2017-05-16
Update slf4j version to be later than used by other dependences

gwalbran:
Currently the slf4j api version was 1.7.5 but slf4j-log4j12 impl was 1.5.8
which causes errors.


jachope:
Closing as not clear what error this PR is fixing

PR #2489, opened 2017-05-25
Clarify keyword search field behaviour with updated field text backgr…

ggalibert:
…ound.

### DIFF
--- a/web-app/js/portal/lang/en.js
+++ b/web-app/js/portal/lang/en.js
@@ -170,7 +170,7 @@ OpenLayers.Lang.en = OpenLayers.Util.extend(OpenLayers.Lang.en, {
     parameterFilter: 'Parameter',
     freetextFilter: 'Keyword',
     freeTextFilterSearchErrorMsg: "Search for free text within metadata records ",
-    freeTextFilterSearchHelperMsg: "Keyword Search.. ",
+    freeTextFilterSearchHelperMsg: "Type keyword(s) and hit Enter.. ",
     freeTextSearchClearButton: '<span class=\"fa fa-remove\"></span> Clear <b>\'${val}\'</b>',
     dateFilter: 'Date (UTC)',
     geoFilter: "Geographic Boundary",


pmbohm:
@ggalibert Why do you see the need to say `hit enter ..` ?

ggalibert:
@pmbohm ???

Because @pblain showed that 20% of the portal users don't know what to do and I have seen this myself during portal demonstrations to students.

pmbohm:
So 20% of people who use Google never find anything because they never press enter?

pmbohm:
Heres a screen shot of the search for www.carsales.com.au which includes a `keyword` free text search.
Enter a keyword and nothing happens until you press enter. Its the way it is in 2017

screen shot 2017-05-25 at 4 45 34 pm

What would be nicest is the results are retrieved as soon as the users finishes typing, but thats a feature and its been a while since we added one of those. :)

ggalibert:
Bad analogy @pmbohm , Google kept a "Search Google" button on their page.

Why do we need to discuss so hard for a modification that costs nothing! I've just done it for you.

Next time their is a portal demo, you should run it so that you can tell 20% of the students that they are idiots.

pmbohm:
@ggalibert Because its longer, wordier and as you propose it, leaves out the word search.
Its slower to spot the meaning

ggalibert:
When the key information is missing, it is shorter and hence clearer. Who would have thought?

Ok I'll close this as I can see I'm facing a wall of bricks. Well played guys... @pblain ?

pblain:
@ggalibert , @pmbohm - I can tell you one thing with certainty - NOBODY can predict how users will respond to a UI change. That's why we need to measure user behaviour rather than argue about it. We'll let it run as is a bit longer to get a better sample size - and then we'll test it with new text to see if the stats change one way or the other. Then we'll _KNOW_ who's right. Anyone want to place some bets?

I've done some reading on this type of thing so I have a suspicion about what will happen here. But it's still speculation and I don't want to inflame the argument. What we will do is test it with Google Analytics, find out which changes help and which don't, and converge on the best solution over time.

ggalibert:
Google analytics does not tell you everything though, since you agreed yourself that there are some situations when it's hard to interpret what happened in the user's head. How do you collect this feedback?

It is frustrating for POs because we are (more than devs) exposed to users via workshops and conferences and when we echo their feedback to you, we're not taken seriously.

pblain:
I take all feedback seriously.  I attended a workshop with Kath to see what happens.  I've also run usability tests. I also added HotJar so I can see recordings of real user activity on the site. This is all valuable and interesting - but it is anecdotal. Google Analytics is the only statistically valid way we have to optimise for usability. Everyone who is serious about usability is running AB tests - nearly always with Google Analytics. This is how all web based business, both large and small, who depend on satisfied users for there existence optimise their web sites. 

If the text change works we will see greater than 78% (or whatever it settles at) of users clicking enter after they go to the free text search. We will then optimise accordingly.

Issue #2481, opened 2017-05-11
Can't unexpand search result details


akashisama:
Steps to reproduce:
Open step 1. Scroll down and click on the title of the "The Australian Phytoplankton Database" result. The details will expand. Now click the title again.

What should happen:
The result goes back to how it was originally, unexpanded.

What does happen:
It remains expanded.

Must be related to https://github.com/aodn/aodn-portal/pull/2458

akashisama:
Sadly https://github.com/aodn/aodn-portal/pull/2491 doesn't seem to have consistently fixed the problem so I'm reopening

akashisama:
Still works consistently great on my local machine, AJAX problem? The data's already there though, no need for AJAX...

PR #2491, opened 2017-05-25
Fix #2484 and #2481 - phytoplankton search result

### LINKED ISSUE #2484: The Australian Phytoplankton Database (1844 - 2016)
## Steps to reproduce
Find Australian Phytoplankton Database (1844 - 2016) - abundance and biovolume on Step 1

## What happens
Can't see the date range because of the organisations

## What should happen
Can see the date range.

julian1:


akashisama:
Surprised it fixes #2481 as well, but luckily it does

PR #2492, opened 2017-05-29
Fixes the selenium tests to account for new order of date and organis…

akashisama:
…ation details in search results, which were changed for #2484.

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/step1/FacetOrganisationTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/step1/FacetOrganisationTest.java
@@ -84,7 +84,7 @@ public class FacetOrganisationTest extends BaseTest {
 
         List<WebElement> results = webElementUtil.findElements(By.className("resultsTextBody"));
         for (WebElement result : results) {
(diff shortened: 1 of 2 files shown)


selenium/src/test/java/au/org/emii/portal/tests/step1/FacetOrganisationTest.java
```diff
@@ -84,7 +84,7 @@ private void verifyOrganisation(String org) {
 
         List<WebElement> results = webElementUtil.findElements(By.className("resultsTextBody"));
         for (WebElement result : results) {
-            WebElement facetResult = result.findElement(By.xpath("div[2]/span[2]"));
+            WebElement facetResult = result.findElement(By.xpath("//span[contains(@class, 'fa-institution')]/../following-sibling::span"));
```
bhasin85: How about 

`div[3]/span[2]` 

instead of 

`//span[contains(@class, 'fa-institution')]/../following-sibling::span`?

akashisama: Sometimes there's no date so organisation won't be in the same numbered spot

Issue #2487, opened 2017-05-25
Step 1 search results Icons need tooltips to determine meaning


pmbohm:
### What happens
In Step 1 search results, the icons used are not enough to determine meaning
### What should happen
Tooltips should be added by adding `title` attributes. 
eg: `<span title="Platform" class="fa fa-fw fa-cog"></span>`

kereid:
If someone really wants to know they can go to search in the user guide in particular https://help.aodn.org.au/using-the-portal/select-data/

pmbohm:
@kereid Its not a game of chess. If you want to know what i mean by that you could read this entire book:
http://www.integratedbook.com/dont-make-me-think-revisited-3rd-edition/

pblain:
Great suggestion Phil.  Anyone interested in usability should, at a minimum, read that book. It's an easy read - which isn't surprising given the subject! There are some other great books on usability too. I recommend Oli Gardner's e-books and blog posts on the subject. Anyone really interested should also get up to speed on the latest research in human perception and cognition.

kereid:
I do agree that they aren't very informative as I logged the issue over 2 years ago see https://github.com/aodn/aodn-portal/issues/1548 so I think it's great it's come up again for discussion, that's why we put the symbol explanation into the user guide.  I guess there must have been a reason the bug was closed without enhancing and I just had a quick look in the backlog but couldn't find it if it's there, either way it will definitely get discussed at the next planning when we discuss the bugs.

pmbohm:
` "Ugly little symbols"` :( 

_oh the humanity_ :

pblain:
Tool tips don't solve usability issues. If the UI is hard to understand we need to change the UI so that it's  _easy_ to understand. Tool tips are great for users who want more detail and they can be very helpful. But if users can't understand what's going on without reading tool tips we have a big problem. I don't think we have such a problem here.  Take a date range such as 2007-2017. It's clearly a date range - even with no icon at all. Nobody needs to be told it's a date range.  Adding text that reads "Date range" distracts from what's important - the dates. I've tested users with this. They all know it's a date range.  

There are real usability problems with Step 2 however:  
* We still have metadata in the titles which makes them hard to read and compare. The metadata should be _consistently_ laid out in the search results - not inconsistently in the titles. 
* We have a _list_ of icons/metadata in the search results - which makes them hard to read and compare. The metadata should be positioned for easy reading and comparing - not in a list. 
[...]

PR #2495, opened 2017-06-07
Tooltip hover labels for faceted search result icons

pmbohm:


pmbohm:
Fixes #2487

akashisama:
I think it would be better to keep the tooltip names the same as the names used in the facets, to emphasise that they're the same thing. ie "Parameter", "Organisation", "Platform"

pmbohm:
I see it as an opportunity to give a long hand explanation of "Parameter" and "Platform" in particular. 

And i was thinking along the lines of this to help solve the connection to the facets https://github.com/aodn/backlog/issues/639

akashisama:
If we do add that feature then I think it will make even more sense to use the same labels for both.

PR #2497, opened 2017-06-07
Fixes #2488 by making the search icon functional

### LINKED ISSUE #2488: Step 1 keyword search usability
This is an issue that has been witnessed by me, Seb and Bene during the last TAMS practical, where the IMOS and the AODN portal were presented to 20 chinese students. At least 5 of them faced the same usability problem.

**What happens**
Looking for the hawksbill sea turtle data collection, I start typing "hawksbill sea turtle" in the keyword search and then I'm facing 3 problems: 
1. I don't have any hint telling me that hitting Enter would trigger the search.
2. I'm guessing I must click the on the magnifier logo. Unfortunately, that didn't work. 
3. Ok I must have missed something in the keyword search field so I go back in there but when doing so, I loose the text I had typed!

**What should happen**
Several improvements can be suggested for each problem mentioned above:
1. Change the initial background grey text of the keyword search field from "Keyword Search..." to "Keyword search and press Enter...". Or add a button Enter. Or add a tooltip somewhere.
2. Trigger the search when one clicks on the magnifier logo.
3. Do not clear the typed text when I leave the field and go back to it for editing/tuning the search for example.

pmbohm:
Fixes the issue without any extra clutter

### DIFF
--- a/web-app/js/portal/search/FreeTextSearchPanel.js
+++ b/web-app/js/portal/search/FreeTextSearchPanel.js
@@ -36,7 +36,7 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
                             tooltip: OpenLayers.i18n('searchFieldText'),
                             listeners: {
                                 click: function() {
-                                   this.onGo();
+                                   this.onSearchIconClick();
                                 },
                                 scope: this
                             }
@@ -76,7 +76,6 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
         this.searchField.on('focus', function() {
             trackUsabilityTest(OpenLayers.i18n('usabilityTestKeywordSubmitAction')
                 , OpenLayers.i18n('usabilityTestKeywordGotFocusLabel'));
-            this.searchField.reset();
         }, this);
     },
 
@@ -91,7 +90,6 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
         this.setCurrentFilter(currentVal);
         trackFacetUsage(OpenLayers.i18n("keyword"), currentVal);
         this.searcher.search();
-        this.searchField.reset();
     },
 
     setCurrentFilter: function(currentVal) {
@@ -101,7 +99,6 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
         }
         else {
             this.currentFilterContainer.collapse();
-            this.searchField.reset();
         }
     },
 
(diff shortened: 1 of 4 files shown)


akashisama:
Have we got a big enough sample from google analytics yet? @pblain said a sample size of 1000 or so visits was needed  think? Otherwise we can't be confident about what impact the change has.

pmbohm:
@akashisama Removed clearing of the search box on focus

web-app/js/portal/search/FreeTextSearchPanel.js
```diff
@@ -88,6 +85,7 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
 
     onGo: function() {
         var currentVal = this.searchField.getRawValue().toLowerCase();
+        console.log(currentVal);
```
akashisama: unneeded console log

web-app/js/portal/search/FreeTextSearchPanel.js
```diff
@@ -116,6 +114,14 @@ Portal.search.FreeTextSearchPanel = Ext.extend(Ext.Panel, {
         }
     },
 
+    onSearchSubmit: function() {
```
akashisama: This will catch both a hit on 'enter' and a hit on the magnifier yeah? It'd be better to keep them separate for better comparison with the current analytics.

pmbohm: They are seperate

web-app/js/portal/search/FreeTextSearchPanel.js
pmbohm: aghhh

web-app/js/portal/search/FreeTextSearchPanel.js
akashisama: ohh excellent. If the user is clicking enter, then search field won't be "dirty" and therefore it's a magnifying glass click.

pblain:
@akashisama , @pmbohm - as mentioned above we need about 1000 keyword search events to be statistically valid. We are currently at 580. We should be getting close to 1000 by the end of the month.

bhasin85:
@pmbohm @akashisama @pblain  Can we close/merge this one?

pmbohm:
No as far as i'm concerned this is the fix for #2488
I'll rebase it

pmbohm:
@pblain Ok to merge?

pblain:
@pmbohm - see https://github.com/aodn/aodn-portal/issues/2488#issuecomment-359663580

pmbohm:
@pblain This current pull request contains fixes offering a few features !
1) Clears out previous search text onFocus
1) Tooltips 
1) puts the search icon to the right of the box and makes it active as though it was Enter.

PR #2499, opened 2017-06-16
The results were taking too long to load sometimes (~5seconds)

akashisama:
so I've added a wait

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/step1/FacetUntickTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/step1/FacetUntickTest.java
@@ -24,6 +24,9 @@ public class FacetUntickTest extends BaseTest {
         WebDriver driver = getDriver();
         driver.get(AODN_PORTAL_SEARCH_PAGE);
 
+        WebDriverWait wait = new WebDriverWait(getDriver(), 30);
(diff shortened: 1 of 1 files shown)


selenium/src/test/java/au/org/emii/portal/tests/step1/FacetUntickTest.java
```diff
@@ -23,6 +24,8 @@ public void untickFacetTest() throws InterruptedException {
         log.info("Loading search page - Step 1");
         WebDriver driver = getDriver();
         driver.get(AODN_PORTAL_SEARCH_PAGE);
+        WebDriverWait wait = new WebDriverWait(getDriver(), 30);
+        wait.withTimeout(10, TimeUnit.SECONDS);
```
bhasin85:
```java
WebDriverWait wait = new WebDriverWait(getDriver(), 30);
```
is equivalent to 

```java
wait.withTimeout(30, TimeUnit.SECONDS);
```

I do not see any advantage in overriding the timeout.

akashisama: Whoops that method wasn't doing what I though it was. I've changed it to wait until there are results.

PR #2503, opened 2017-06-23
A potential fix for the issue of enabledFilters not working with geow…

akashisama:
…ebcachce


pmbohm:
@pblain @akashisama This forces one Geowebcache to map one to one WMS server which is not suitable for the AODN space. But I understand this is a RC hack to temporarily bypass hacking Geowebcache.

akashisama:
I've used this branch for the current sys-test hence the version bump. The reason I did this was because we're planning to revert this code in the next iteration so I didn't see any reason cluttering the commit history of master.

PR #2511, opened 2017-07-03
Allow testing for any download type rather than just shapefiles

akashisama:
Required some additional waits since the ncwms layers take ages to load their temporal extent


akashisama:
Bump

PR #2513, opened 2017-07-04
Missing temporal extent

bhasin85:


src/test/javascript/portal/cart/GogoduckDownloadHandlerSpec.js
```diff
@@ -125,5 +125,45 @@ describe('Portal.cart.GogoduckDownloadHandler', function () {
...
+
+            url = clickHandler(testCollection, testHandlerParams);
+
+            expect(url).toHaveParameterWithValue(
+                'jobParameters.subset',
+                'LATITUDE,20.0,42.0;' +
+                'LONGITUDE,-170.0,-160.0'
+            );
+
+            expect(url).not.toHaveParameterWithValue(
```
akashisama: Isn't this expect a bit superfluous? It will only fail when the previous expect already failed.

bhasin85: Maybe we can get rid of lines 154 to 158


```java
expect(url).toHaveParameterWithValue(
'jobParameters.subset',
'LATITUDE,20.0,42.0;' +
'LONGITUDE,-170.0,-160.0'
);
```

PR #2514, opened 2017-07-05
Possibly fixes #2507

### LINKED ISSUE #2507: Cannot un-expand search result details
This is the same as https://github.com/aodn/aodn-portal/issues/2481

I'm still seeing this in systest for both Chrome AND Firefox yet it was closed 15 days ago so assuming it would have been released into systest...

Steps to reproduce: 

On step 1 click the result of Australian Phytoplankton database to see all of the organisations that contributed. Click again.

What should happen:
Result text unexpands

What does happen:
Remains expanded - sometimes it unexpands and then straight away expands again.

akashisama:
When expanding a result, only set the originalheight when it is not already defined, meaning only on the first run. The original height should never change so this is cleaner than what we had before even if it doesn't fix the bug.

### DIFF
--- a/web-app/js/portal/jquery.js
+++ b/web-app/js/portal/jquery.js
@@ -4,10 +4,15 @@ jQuery( document ).ready(function() {
         function() {
             var resBody = jQuery(this).children('.facetedSearchResultBody');
             var fullHeight = resBody[0].scrollHeight;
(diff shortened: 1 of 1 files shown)


anguss00:
Good stuff, will merge when Trav is happy.

PR #2515, opened 2017-07-06
Fixes #2502

### LINKED ISSUE #2502: Some layers with diver platform have empty platform row in results
**Steps to reproduce:**

In the platform facet expand "Biological platform" and select "diver".

**What should happen:**

All results contain "Biological platform" in the platform row.

**What does happen:**

"Mapping of selected areas of Cockburn Sound, Western Australia", "Habitat Maps derived from Point Lookout Ecological Surveys (PLEA)" and "Ecophysiology of three habitat-forming seaweeds across a depth and latitudinal gradient in southeastern Australia" are missing the platform row.

akashisama:
Handle the case where metadata lists a non-recognized platform before any recognized platform

### DIFF
--- a/web-app/js/portal/search/FacetedSearchResultsDataView.js
+++ b/web-app/js/portal/search/FacetedSearchResultsDataView.js
@@ -186,21 +186,25 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
         return "";
     },
 
-    _getPlatformAsHtml: function(template, platform) {
+    _getPlatformAsHtml: function(template, platforms) {
 
         var label = this._buildLabel("fa-tags", OpenLayers.i18n('searchPlatformText'));
-
-        var broader = this.classificationStore.getBroaderTerms(platform, 1, 'Platform');
-        if (broader.length > 0) {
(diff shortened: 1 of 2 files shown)


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -186,21 +186,25 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
...
-            broader = broader.filter(function(item, pos) {
-                return !pos || item != broader[pos - 1];
-            });
-            return template.apply({
-                "label": label,
-                "value": broader.join(', ')
-            });
+        var broader = [];
+        var i = 0;
+        for(i=0;i<platforms.length;i++) {
```
anguss00: I reckon we can use Ext.each here instead of an iterator... saves a variable assignment

anguss00:
Also, when I select Platform-> Biological platform -> Diver, the platform row disappears from the search results which I don't think is the required behaviour

anguss00:
Scratch that, saw the changes at the same time as I commented!!

PR #2516, opened 2017-07-06
Send workspace and layername together rather than as separate parameters

akashisama:
for enabledFilters and uniqueValues requests


akashisama:
Fixed the tests

PR #2519, opened 2017-07-14
Reverse the proxy redirect logic since we have tilecache in metadata now

akashisama:


akashisama:
I've now reversed the tests as well because obviously I needed to reverse the tests as well.

PR #2525, opened 2017-08-01
allowAnyHost now works in dev environments.

pmbohm:
Sick of having to add to `knownServers` in `Config.groovy`
Fear no more! Filters in step 2 will magically populate in dev environments


pmbohm:
dooogh!! Cant be done as other meta information is required

PR #2526, opened 2017-08-03
Fixes #2524

### LINKED ISSUE #2524: Warning message when no temporal info exists
## Steps to reproduce
1. Select the bathymetry layer (i.e. the new layer with no time steps)
2. Go to step 2

## What happens
1. A warning message appears that reads "The dataset has no temporal information" 

## What should happen
1. No warning message
2. Info text that reads "Spatial subsetting only for this data collection"
3. Info text is styled black/grey not orange.

pmbohm:


anguss00:
Looks good @pmbohm. Just want to check with you RE testing

pmbohm:
RE ? Regular expression?

pmbohm:
@anguss00 
### Heres some senario's for testing
The 'real' layer we are testing is supposed to be in 'rc' but is broken at the moment

#### Cause portal to have no values in temporal extent

NcWMS.js:62
`var datesToProcess = [];// dateFilter['possibleValues'];`

The ncWMS collection has no values in temporal extent, yet the metadata says that there are temporal values (as it should)
 shouldHaveTemporalExtent will return true with properly configured collections apart from the new collection that caused all this `Port Phillip Bay Bathymetry`.

#### Cause portal to expect no Temporal extent
Change NcWmsPanel.js:510 to return false
```
    shouldHaveTemporalExtent: function() {
        var metadata = this.dataCollection.data.metadataRecord.data;
        // return (metadata.temporalExtentBegin.length > 0) && (metadata.temporalExtentEnd.length > 0);
        return false;
    },
```

PR #2529, opened 2017-08-08
Updated loading indicator test

bhasin85:


bhasin85:
@akashisama  Can we get this one merged?

akashisama:
Nah this code doesn't work. The line ```webElementUtil.findElement(By.className("fa-spinner"), 100, TimeUnit.MILLISECONDS);``` checks the existence of the spinner, but the spinner exists whether it's visible or not, so this will always pass even if we break the spinner. Will close.

PR #2530, opened 2017-08-10
Fix the un-expand issue for firefox

### LINKED ISSUE #2507: Cannot un-expand search result details

bhasin85:
Fixes #2507  

Tested on BrowserStack OS Windows 7 FireFox 54

### DIFF
--- a/web-app/css/general.css
+++ b/web-app/css/general.css
@@ -981,7 +981,6 @@ div.allwhite {
 }
 
 .resultsHeaderBackground .facetedSearchResultBody {
-    height: 100px;
     overflow-y: hidden;
 }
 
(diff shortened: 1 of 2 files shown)


pmbohm:
@bhasin85 many cheers for BrowserStack! thanks

PR #2531, opened 2017-08-10
Added grey text and tooltip

### LINKED ISSUE #2488: Step 1 keyword search usability

bhasin85:
Fixes #2488 

## Acceptance Criteria (B Test)
1. The default grey text reads: "Type keyword, press enter".
1. The text entered by the user is not cleared when re-entering the text box.
1. The default grey text is cleared when the user enters the text box.
1. The default grey text is re-instantiated if the box is cleared or left empty by the user.
1. Tool tip added that reads: "Type a keyword and press enter to narrow search results".
1. _No other changes_ from the A test.
1. Google Analytics tag added that indicates when the B test is released to prod (This bit will be done later)

Note: After we compare the current A-test results with this B-test we will test the impact of a "'Go" button.


bhasin85:
Thanks @jonescc  :+1:

PR #2536, opened 2017-08-24
Large result record can be unexpanded

bhasin85:


selenium/src/test/java/au/org/emii/portal/tests/step1/UnexpandLargeRecordTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+public class UnexpandLargeRecordTest extends BaseTest {
+
+    private static Logger log = Logger.getLogger(UnexpandLargeRecordTest.class.getName());
+
+    @Test
+    public void facetDateTest() {
```
akashisama: test function needs to be renamed

akashisama:
On Firefox the record go stale when expanded

PR #2549, opened 2017-08-30
Fixes #2537

### LINKED ISSUE #2537: If there is no spatial subset drawn on map there should be no co-ords displayed on step 2 or 3
Steps to reproduce:

1 Add any collection and draw a bounding box look at step 3 
2 Decide you would actually like to draw a 3 point polygon so change the option to polygon on step 2 look at step 3 
3 Go back to step 2 and draw a polygon, look at step 3
4 Go back to step 2 and change option to bounding box

What does happen
1 see 4 co-ordinates listed 
2 still see 4 co-ordinates listed but nothing displayed on map
3 see message about polygon
4 See co-ordinates in the NSEW boxes

What should happen
1 see 4 co-ordinates listed
2 see no co-ordinates listed - I don't know if it is applied to the download or not
3 see message about polygon
4 either don't see co-ordinates OR bounding box reappears

pmbohm:
Show geometry as a point list (if geometry is set) when selecting polygon Spatial Subset drawing

### DIFF
--- a/src/test/javascript/portal/prototypes/OpenLayersSpec.js
+++ b/src/test/javascript/portal/prototypes/OpenLayersSpec.js
@@ -200,7 +200,7 @@ describe('OpenLayers', function() {
         });
 
         it('set polygon spatial constraint control when style is POLYGON', function() {
-            map.doSetSpatialConstraint(Portal.ui.openlayers.SpatialConstraintType.POLYGON);
+            map.doSetSpatialConstraintStyle(Portal.ui.openlayers.SpatialConstraintType.POLYGON);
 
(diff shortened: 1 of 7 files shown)


pmbohm:
If a user changes between Polygon/Bounding box, all they have done is select what bounds will be drawn next. They have not chosen to clear the bounds. The existing bounds still apply and are displayed in the filter and Step 3, despite not appearing graphically on the map

There is no requirement for the bounds to be reset to default when swapping between Polygon/Bounding box

akashisama:
Then we need to make sure that the previously applied bounds continue to appear graphically  on the map. Users expect WYSIWYG.

jonescc:
Updated to always ensure last constraint is drawn until a new valid constraint is entered or the constraint is cleared.

@akashisama

Issue #2534, opened 2017-08-18
NetCDF Aggregation request slows portal systest


akashisama:
### To reproduce
Follow the instructions at https://github.com/aodn/geoserver-build/issues/240

I'm logging this as a separate bug because the portal shouldn't fall over because of a bug in another system.

### What should happen
An error message appears saying the aggregation failed, and then the portal continues to run smoothly like it was before.

### What actually happens
After the aggregation fails the portal becomes unusable, with proxied requests to any external service (such as catalogue-systest and geoserver-systest) constantly timing out. For instance, go back to step one and choose a facet, you will get a "Search is currently unavailable." error. This occurs not just for the user who has requested the aggregation, but any user running the portal via the impacted EB instance.

pmbohm:
Can this be fixed by any portal code changes? Is this manifestation occurring because the Systest EB stack is sharing memory? If so, this cant be a portal issue

pmbohm:
An issue can be triggered that shows weakness in our Portal proxy code. If a layer is loaded and the WMS is unresponsive the maximum connections can be reached in a browser session. 
#### To Trigger
 1) Use `geonetwork.url = "http://catalogue-systest.aodn.org.au/geonetwork"` portal.groovy in dev environment 
2) Load a gridded dataset and the URL's are like  `http://geoserver-systest-internal.aodn.org.au/geoserver/ncwms?LAYERS=blagh.....`
3) Try to use free text search on step one and notice the portal is unresponsive

PR #2550, opened 2017-08-31
Fixes 2534 by introducing a read timeout on openConnection in portal proxy

pmbohm:
fixes https://github.com/aodn/aodn-portal/issues/2534

### DIFF
--- a/src/groovy/au/org/emii/portal/proxying/ProxiedRequest.groovy
+++ b/src/groovy/au/org/emii/portal/proxying/ProxiedRequest.groovy
@@ -14,9 +14,9 @@ class ProxiedRequest extends ExternalRequest {
     def response
     def params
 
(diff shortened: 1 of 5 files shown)


pmbohm:
Downloads are not affected by this change, however downloads using the proxy are not the cause of the problems noticed in the bug

src/groovy/au/org/emii/portal/proxying/ProxiedRequest.groovy
```diff
@@ -14,9 +14,9 @@ class ProxiedRequest extends ExternalRequest {
     def response
     def params
 
-    ProxiedRequest(request, response, params, proxyRedirectService) {
+    ProxiedRequest(request, response, params, proxyRedirectService, grailsApplication) {
```
akashisama: Why require the whole config? Wouldn't it be simpler to just receive the timeout value by itself?

PR #2553, opened 2017-09-07
Test that previous button isn't visible in step 1,…

akashisama:
…including after going to step 2. Test Rail C4934

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/PreviousButtonTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/PreviousButtonTest.java
@@ -1,0 +1,35 @@
+package au.org.emii.portal.tests;
+
+import org.apache.log4j.Logger;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.testng.Assert;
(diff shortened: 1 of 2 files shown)


selenium/src/test/java/au/org/emii/portal/tests/PreviousButtonTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.apache.log4j.Logger;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class PreviousButtonTest extends BaseTest {
+    private static Logger log = Logger.getLogger(PreviousButtonTest.class.getName());
```
bhasin85: Maybe we can get rid of this line ad we are not using it.

selenium/src/test/java/au/org/emii/portal/tests/PreviousButtonTest.java
```diff
@@ -0,0 +1,35 @@
+package au.org.emii.portal.tests;
```
bhasin85: How about moving the class under package au.org.emii.portal.tests.step1?

PR #2555, opened 2017-09-07
Issue #2556 - Increase height of the download confirmation window - to prevent scrollbars from being displayed (with current content).  Will improve usability.

ccmoloney:
Increase height of download confirmation window - so that user doesn't have to scroll.


bhasin85:
Related issue #2556

PR #2558, opened 2017-09-13
fixes #2552

### LINKED ISSUE #2552: Proxy Time out
### Steps to reproduce:
Step 1. Use geonetwork.url = "http://catalogue-systest.aodn.org.au/geonetwork" portal.groovy in dev environment
Step 2. Load a gridded dataset and the URL's are like http://geoserver-systest-internal.aodn.org.au/geoserver/ncwms?LAYERS=blagh.....
Step 3. Check network log of your browser.

### What should happen
Step 2. The getmap requests time out because internal geoserver can't be accessed.
Step 3. The network log show that the requests sent back failed response codes.

### What does happen
Step 2. The getmap requests time out because internal geoserver can't be accessed.
Step 3. The network log shows the requests as having succeeded with 200 response codes.

Discovered during work on #2534. This bug could make future debugging difficult/confusing.

pmbohm:


bhasin85:
Related issue #2552

src/groovy/au/org/emii/portal/proxying/RequestProxyingController.groovy
```diff
@@ -44,10 +44,17 @@ abstract class RequestProxyingController extends HostVerifyingController {
...
         _setDownloadFilename(response, params)
-        _makeRequest(request, response, params, paramProcessor, streamProcessor)
+
+        try {
+            _makeRequest(request, response, params, paramProcessor, streamProcessor)
+        }
+        catch (IOException e) {
+            def msg = String.format("Could not reach %s %s", probeUrl, e.message)
+            log.error msg
+            render text: msg, contentType: "text/html", encoding: "UTF-8", status: HTTP_500_INTERNAL_SERVER_ERROR
```
bhasin85: I thought that the content type should be png

`Content-Type: image/png`

akashisama: Status should be a 504 Gateway Timeout: "The server was acting as a gateway or proxy and did not receive a timely response from the upstream server."

akashisama: I think contentType should be excluded, since we can't know what the proxied server would have returned if it was working.

pmbohm: contentType is text/html as we are returning a message, but yes 504 Gateway timeout sounds better

Issue #2542, opened 2017-08-28
Selenium browser update failing ExpandCollapseLargeRecordTest


bhasin85:
Test | OS | Browser | Duration | Finished
-- | -- | -- | -- | --
ExpandCollapseLargeRecordTest | Win 10 | Firefox 55.0 | 41s | 3d ago
ExpandCollapseLargeRecordTest | Win 10 | Firefox 54.0 | 45s | 3d ago
ExpandCollapseLargeRecordTest | Win 10 | Firefox 53.0 | 45s | 3d ago

bhasin85:
Able to reproduce.

akashisama:
Re-opening, latest fix fixes the functionality, but the selenium test is still failing on some browser/os combinations.

akashisama:
Sorry, no, correction, looking at the test run video again, the functionality _is_ failing still :(

PR #2562, opened 2017-09-14
Reworking of the test.

pmbohm:
The refactoring ensures we are definitely testing the correct elements should the order have been a problem.

I noticed the tests were always succeeding when i had lots of debug lines and queries within the test. This suggests that although we are waiting for animations to end, the style attributes were still not written into the page and we were testing them too early, hence the new wait.


pmbohm:
fixes #2542

bhasin85:
Smashed it :+1: 
giphy

PR #2571, opened 2017-10-19
Status urls from Batch/Geoserver GoGoDuck in popup

pmbohm:
Required for the new AWS batch GoGoDuck. Will break status urls returned from Geoserver GoGoDuck

### DIFF
--- a/grails-app/services/au/org/emii/portal/WpsAwsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsAwsService.groovy
@@ -46,6 +46,6 @@ class WpsAwsService extends AsyncDownloadService {
     def _getExecutionStatusUrl(registerResponse) {
         def xmlLink =  new XmlSlurper().parseText(registerResponse).@statusLocation[0].toString()
 
-        return xmlLink.replace(/format=xml/,"format=html")
+        return xmlLink.replace(/format=XML/,"format=HTML")
     }
 }
--- a/grails-app/services/au/org/emii/portal/WpsAwsService.groovy
+++ b/grails-app/services/au/org/emii/portal/WpsAwsService.groovy
@@ -1,0 +1,51 @@
+package au.org.emii.portal
+
+import grails.converters.JSON
+import groovy.xml.StreamingMarkupBuilder
+import groovyx.net.http.HTTPBuilder
+import groovyx.net.http.HttpResponseException
+
+import static groovyx.net.http.ContentType.XML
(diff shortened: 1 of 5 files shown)


pmbohm:
A cleanout is needed here starting with goggoduckV1 and therefore AsyncDownloadService.groovy, This is why this pull request starts to ignore AsyncDownloadService.groovy but many more changes are required to clean it all out.

bhasin85:
@pmbohm How about we update the PR to make it work with both the gogoduck's? That way we can get the PR merged.

pmbohm:
This pull requests handles `OGC:WPS--gogoduck` protocol WPS requests, with urls containing `https://wps.aodn.org.au` 
Standard Geoserver WPS GoGoDuck download requests, still handle sending emails.

jonescc:
Perhaps looking for geoserver in the url would be more indicative of whether we are using geoserver or AWS batch (all our geoserver instances include the geoserver context).   looking for the url above will only work with the prod wps endpoint which will make testing difficult.

jonescc:
I also wouldn't check the protocol or otherwise at some point we will have trouble deploying aws-wps netcdfoutput to test environments

pmbohm:
@jonescc Ok so now the new workflow for AWS batch will be used when the `jobType` is 'GoGoDuck' not 'NetcdfOutput', and the server does not contain 'geoserver'

jonescc:
I there some reason to check the protocol?   Otherwise we'll have to remember to change this when we come to migrate the netcdfoutput process.

pmbohm:
But we haven't migrated the netcdfoutput process! So we cant use the new code for it

grails-app/services/au/org/emii/portal/WpsAwsService.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    def onResponseSuccess = { resp, xmlReader ->
+        return new StreamingMarkupBuilder().bindNode(xmlReader).toString()
+    }
+
+    def _getExecutionStatusUrl(registerResponse) {
+        def xmlLink =  new XmlSlurper().parseText(registerResponse).@statusLocation[0].toString()
+
+        return xmlLink.replace(/format=xml/,"format=html")
```
jonescc: Returns a URL ending in format=XML not format=xml - so this doesn't work

jonescc:
Looks good - works for both old and new wps endpoints - just one issue to fixup.

PR #2580, opened 2017-11-22
Fix #2574 absolute image path GFI popup pointers

### LINKED ISSUE #2574: GFI large tables results cant be viewed
### Whats happening
On the new https://vmdp.deakin.edu.au/portal/search portal they are using tables content.ftl's (or no content.ftl) for GetFeatureInfo, resulting in large tables of results that cant be scrolled as there is no available horizontal scrolling.
Also happening in our portal with the new Animal Tracking QC layer http://portal-rc.aodn.org.au/search?uuid=0ede6b3d-8635-472f-b91c-56a758b4e091

### What should happen
Content should be accessible 

screen shot 2017-11-15 at 11 00 14 am

screen shot 2017-11-15 at 11 12 18 am

pmbohm:
This was affecting Deakin who have their portal with the context set as `/portal`
The problem can be seen on http://po.aodn.org.au/imos123/search
This fix works with contextless setups like production or through intellij

### DIFF
--- a/web-app/css/general.css
+++ b/web-app/css/general.css
@@ -534,7 +534,7 @@ div.olMap.x-item-disabled * {
 }
 
 .gx-popup-anc{
-    background:transparent url(/images/popupPointer.png) no-repeat 0 0;
+    background:transparent url(../images/popupPointer.png) no-repeat 0 0;
     position:absolute;
     left:45px;
     z-index:2;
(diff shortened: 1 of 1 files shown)


pmbohm:
@pblain This will fix Deakins problem, although they have fixed it themselves in the interim.

pblain:
@pmbohm Great. I'll let them know.

Issue #2554, opened 2017-09-07
BrowserStack Daily Usage Quota Reached


bhasin85:
Hi Ankit Bhasin,

100% of your organization's (IMOS) Automate testing time has been used. Your limit will be reset at 7th Sep, 2017 00:00 hrs UTC. To ensure our customers have the optimum user experience, BrowserStack has a Fair Usage policy through generous daily testing time limits. As your usage is extensive, we recommend you upgrade your Automate plan.

If you have any queries, reply to this email, and we will be in touch.

Happy testing
The BrowserStack Team
www.browserstack.com

bhasin85:
If we cannot increase our free quota, we can reduce the number of os/browser/version which will bring down the usage.

PR #2581, opened 2017-11-28
Only test firefox version 47.0. Only test two versions of chrome.

### LINKED ISSUE #2545: Selenium browser update failing PointTimeSeriesLatLonTest
Test | OS | Browser | Duration | Finished
-- | -- | -- | -- | --
PointTimeSeriesLatLonTest | Win 10 | Firefox 54.0 | 1m 57s | 3d ago
PointTimeSeriesLatLonTest | Win 10 | Firefox 55.0 | 1m 48s | 3d ago
PointTimeSeriesLatLonTest | El Capitan | Firefox 55.0 | 1m 18s | 3d ago
PointTimeSeriesLatLonTest | El Capitan | Firefox 53.0 | 1m 17s | 3d ago
PointTimeSeriesLatLonTest | El Capitan | Firefox 54.0 | 1m 18s | 3d ago
PointTimeSeriesLatLonTest | Win 10 | Firefox 53.0 | 1m 54s | 3d ago

akashisama:
Fixes #2545 and #2554.

Firefox started using a completely new driver in version 48.0 and that driver doesn't fully support the WebDriver standard yet, which selenium uses. Therefore we should hold off on testing the latest versions of firefox until they've completed their implementation. The implementation progress is being tracked at https://bugzilla.mozilla.org/show_bug.cgi?id=721859

I've also gotten rid of the tests of chrome version 58.0 since the difference between it and 59.0 is insignificant and the extra tests were pushing us towards our daily limit.

### DIFF
--- a/selenium/src/test/resources/testng.xml
+++ b/selenium/src/test/resources/testng.xml
@@ -7,46 +7,10 @@
 
     <!-- Reference https://www.browserstack.com/automate/java-->
     <!--Firefox Windows 10 Configurations-->
-    <test name="Firefox Latest on Windows 10">
+    <test name="Firefox 47.0 on Windows 10">
         <parameter name="browser" value="Firefox"/>
         <!--Note: If browser_version capability is not set, the test will run on the latest version of the browser set by browser capability.-->
-        <parameter name="browser_version" value=""/>
-        <parameter name="os" value="WINDOWS"/>
-        <parameter name="os_version" value="10"/>
-        <parameter name="resolution" value="1024x768"/>
-        <parameter name="device" value=""/>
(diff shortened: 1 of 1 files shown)


anguss00:
Hmm why were we testing two versions of Chrome?

akashisama:
Actually we were testing three versions, the latest, 59.0 and 58.0, so we'll be testing two now. In theory testing multiple versions allows us to see when a change in the browser breaks the test rather than a change we made. The reason we were testing three was because we could and more is always better until proven otherwise.

anguss00:
Gotcha, makes perfect sense!

PR #2583, opened 2017-12-06
Don't test loading indicator on firefox…

akashisama:
… since the firefox testing driver is usually too slow to catch the indicator in time.

The loading indicator only appears for about a second. The firefox driver takes longer than a second to check it. Therefore it only passes about 15% of the time, even though the indicator is working.


anguss00:
I, for one, welcome our new Opera Overlords

PR #2586, opened 2017-12-20
Fix for MOTD on home page without ExtJS

pmbohm:
### To test 
Add an accessible url to portal's config.groovy `motdUrl` Eg: https://s3-ap-southeast-2.amazonaws.com/static.emii.org.au/portalMotd/index.txt

The plain text file should contain two plain text lines. The portal will show a modal popup on the front page.
See https://github.com/aodn/static.emii.org.au_site/pull/13


akashisama:
Seems good to me, but doesn't it require a change to the config file as well, to point motdUrl at the right place?

pmbohm:
@akashisama True this is only part of the picture. Next time we want to show a message now we can via config unlike currently.
There are still missing peices to the puzzle,  eg showing and hiding at particular times, so its probably a manual process to handle motd for now

PR #2596, opened 2018-01-17
Zero data check

pmbohm:
Checks the current maplayer WMS endpoint with a getFeatureRequest.
Stops the download button being available when there is filtering/subsetting that wont return results

### DIFF
--- a/application.properties
+++ b/application.properties
@@ -1,6 +1,6 @@
 #Grails Metadata file
-#Wed Dec 20 16:49:04 AEDT 2017
-app.build.date=20/12/2017 16\:49
+#Tue Nov 14 15:05:31 AEDT 2017
+app.build.date=14/11/2017 15\:05
 app.build.number=UNKNOWN
 app.build.scmRevision=UNKNOWN
 app.grails.version=2.4.4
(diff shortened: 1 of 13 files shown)


pmbohm:
Sorry i broke NcWmsPanelSpec.js by creating SpatialSubsetintersectTester.js

TODO _applyFilterValuesToCollection in NcWmsPanel.js is not called when using the point timeseries options and so there is no check done in this case

gsatimos:
I'm picking this one up.

gsatimos:
The functionality to display a warning on step 2 when no data is selected was reverted with https://github.com/aodn/aodn-portal/pull/2596/commits/434b987a28f624541588522dbd8c3a8526efba9b
and merged with this PR. 

To reproduce this bug 
1. Switch to commit https://github.com/aodn/aodn-portal/commit/5e14212a0834b612ec0698d7815818d3222059db (e.g. create a new branch here)
2. run the portal locally
3. Step 1: pick a NcWMS collection e.g." IMOS - ACORN - Rottnest Shelf HF ocean radar site (Western Australia, Australia) - Delayed mode sea water velocity"
4. On Step 2, draw a bounding box that doesn't include any data
5. In the left hand panel, dismiss the collection
6. Navigate back to step 1 by clicking the step 1 tab
7. Select the same collection
8. On step 2, now draw a bounding box around the data (i.e. a box so you would expect some data)

You should find that a new bounding box is drawn without the initial bounding box being dismissed. 
There will also be a console error.

Issue #762, opened 2013-12-17
Bounding Box Filter Is Not Applied To All Layers


tojofo:
**Steps To Repeat**
Open a 1-2-3 portal
Add a minimum of two layers
Draw a bounding box on the map
Go to the download screen

**What I Expect To Happen**
The filter applied for all layers to contain the bounding box just applied

**What Actually Happened**
The bounding box is only applied to the layer that was active at the time of drawing

PR #2598, opened 2018-01-18
updated labels

gsatimos:
Minor UI changes as part of [backlog issue #762](https://github.com/aodn/backlog/issues/762)

Peter and Phil had a discussion and decided that some facet labels on Step 2 would change so:
- 'Spatial Subset' -> 'Spatial'
- 'Temporal Extent' -> 'Temporal'
- 'Filters' -> 'Others'

I investigated whether changing the size of these labels was feasible but there aren't any good css selectors to make use of to target these (without risking the style change leaking out into other parts of the page/site)


pmbohm:
@pblain  as discussed the word 'Filters' no longer appears

PR #2604, opened 2018-01-23
Empty download check fix

pmbohm:
fixes https://github.com/aodn/backlog/issues/774

### DIFF
--- a/web-app/js/portal/filter/ui/FilterGroupPanel.js
+++ b/web-app/js/portal/filter/ui/FilterGroupPanel.js
@@ -140,27 +140,37 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Ext.Container, {
     },
 
     _handleGetFeatureRequestResults: function(results) {
-        var res = Ext.util.JSON.decode(results.responseText);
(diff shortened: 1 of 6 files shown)


web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -158,8 +157,7 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Ext.Container, {
             this.dataCollection.getFilters(),
             this.dataCollection.layerSelectionModel.selectedLayer.url,
             this.dataCollection.layerSelectionModel.selectedLayer.wmsName.split('#')[0],
-            "application/json",
-            true
```
akashisama: this "true" is needed to make some non-gridded layers work, eg anmn

akashisama: re face to face discussion, 'true' still works for me on argo with spatial extent

pmbohm: ok then. I'll add it

PR #2606, opened 2018-01-30
Handling depth filters

pmbohm:
For https://github.com/aodn/backlog/issues/718
Reads the z-axis information from the getMetadata request and creates a depth/elevation filter if required. 
The selected z-axis is applied to the map by adding "ELEVATION=[a value]" to getMap requests.

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -59,11 +59,11 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
 
     _addZAxisControls: function() {
 
-        this.zAxisPickerTitle= OpenLayers.i18n('zAxisLabel');
+        this.zAxisPickerTitle = OpenLayers.i18n('zAxisLabel');
 
         this.zAxisPickerStore = new Ext.data.ArrayStore({
             fields: [
-                {name: this.zAxisPickerTitle, type: "integer" }
+                {name: this.zAxisPickerTitle, type: "integer"}
             ],
             expandData: true
         });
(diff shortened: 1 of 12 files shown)


ggalibert:
To add to @anguss00 first point and for clarity, my concern is that the [getmetadata](http://geoserver-123.aodn.org.au/geoserver/ncwms?layerName=csiro_cars_monthly_url%2FTEMP&REQUEST=GetMetadata&item=layerDetails) returns ```"zaxis":{"values":[0,-50,-200,-1000,-5000],"positive":false,"units":"m"}``` while in the NetCDF file we have:
```
...
	short DEPTH(DEPTH) ;
		DEPTH:axis = "Z" ;
		DEPTH:long_name = "depth" ;
		DEPTH:positive = "down" ;
		DEPTH:reference_datum = "sea surface" ;
		DEPTH:standard_name = "depth" ;
		DEPTH:units = "m" ;
		DEPTH:valid_max = 12000s ;
		DEPTH:valid_min = 0s ;
...
 DEPTH = 0, 50, 200, 1000, 5000 ;
```
How did the getmetadata end up giving negative values? Is this because of the attribute ```"positive":false```? If so, then this attribute may need to be taken into account so that you can handle values that are of a consistent sign with what is stored in the dataset.

pmbohm:
@anguss00 review changes pushed

web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -169,11 +168,17 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
...
-            return String.format("{0}-{1}",
-                this.zAxisFromPicker.getValue(),
-                this.zAxisToPicker.getValue()
-            );
+
+        if (this._isZAxisFilterAvailable()) {
+
+            if (this.layer.extraLayerInfo) {
+
+                var multiplier =  (this.layer.extraLayerInfo.zaxis.positive === false) ? -1 : 1;
```
anguss00: Sorry to be that guy, but there's an extra space here after the =

anguss00:
One annoying comment, looks great aside from that

anguss00:
Can you squash the commits?... lol jks

PR #2608, opened 2018-02-08
Fixes #2607

### LINKED ISSUE #2607: missing JS function error
### Problem:
JS console error occurs
This error only is seen when there are major configuration problems, unless this occurs in other cases?
```
Uncaught TypeError: this._createErrorMessageContainer is not a function
    at constructor._addErrorMessage (http://127.0.0.1:9090/js/portal/filter/ui/FilterGroupPanel.js:79:38)
```
### Cause:
 _createErrorMessageContainer is missing

### To reproduce (one way)
Remove or mis-identify an ncWMS server as a Geoserver in `Config.groovy` KnownServers array. This will cause the filters request to fail and call _onFiltersLoadFailure. With this fix there will be a warning message box appear

pmbohm:
See #2607 to test


jonescc:
Looks good to me

PR #2609, opened 2018-02-13
added selenium test: Indicate Result Added To Map

gsatimos:
this is a test for [this case case in TestRail](https://aodn.testrail.com/index.php?/cases/view/81)


akashisama:
Could you add an assertion before clicking the button to make sure it's not already showing as selected?

(also: what is wrong with bloody travis I wonder? It failed the build, but obviously this change can't have broken an integration test)

gsatimos:
good idea - I'll add the extra assertion. We'll see how travis behaves after I push the change.

akashisama:
Assertion messages are back to front

PR #2610, opened 2018-02-15
Refactor NcWmsPanel.js Removed zaxisPickerContainer into separate files

pmbohm:


anguss00:
We may need to grab the commit from https://github.com/aodn/aodn-portal/pull/2613, cause we're missing the map reset functionality on step 2

pmbohm:
@anguss00 too many branches in this forest

PR #2613, opened 2018-02-20
fixes #2612

### LINKED ISSUE #2612: Reset vertical subset should revert to topmost z axis slice
From
https://github.com/aodn/issues/issues/168

### Steps to reproduce:
Add cars layer, go to step 2, select a From depth other than zero, click reset.

### Observed behaviour:
depth slice shown on map does not reset to zero point (topmost depth slice)

### Expected behaviour:
Reset shows zero point (topmost depth slice).

pmbohm:


anguss00:
Looks good on step 2, however when I go to step 3 the subset is not reset and maintains the previously selected subset

PR #2623, opened 2018-03-27
Selenium logos for each record

gsatimos:
See discussion here about how https://github.com/aodn/internal-discussions/issues/405 checking logos are greyscale has not been implemented.

This test is for https://aodn.testrail.com/index.php?/cases/view/14858 and checks that either the width or height for each image is as large as the max-width and max-height enforced by the style (and that the styles are all correct for the img elements)

I've gone ahead and updated TestRail to note that the selenium test has been added


anguss00:
Sweet! If you could please squash those commits together we're ready to rock

PR #2624, opened 2018-03-27
Depth service now from Geoserver

pmbohm:
For aodn/backlog#748


pmbohm:
This will require the depth_service.url config change for testing in systest. https://github.com/aodn/chef-private/pull/2763

pmbohm:
@lwgordonimos bump ?

PR #2625, opened 2018-03-28
Add support for sourcing filters and filter values from core geoserver

jonescc:


jonescc:
Needs some more testing

PR #2636, opened 2018-04-18
Sometimes the info container isn't ready for a message to be added…

akashisama:
… so wait till it is.

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -171,9 +171,21 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
             this.loadingInfo.removeClass(style);
         }, this);
 
-        this.loadingInfo.addClass(cls);
-        this.loadingInfo.update(msg);
-        this.loadingInfo.show();
+        var info = this.loadingInfo;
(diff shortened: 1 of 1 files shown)


web-app/js/portal/details/NcWmsPanel.js
```diff
@@ -171,9 +171,21 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
             this.loadingInfo.removeClass(style);
         }, this);
 
-        this.loadingInfo.addClass(cls);
-        this.loadingInfo.update(msg);
-        this.loadingInfo.show();
+        var info = this.loadingInfo;
+        if(info.deferLayout) {
```
jonescc: Is this an internal property?  Couldn't find a description for it and it looks like its used for internal laying out purposes.  I've seen rendered used for this type of check in the past which is a public api property.  Would this work instead - would make it easier to find out what the property does for someone looking at this in future (and me)?

akashisama: Full disclosure, I'm working mostly off trial and error here. rendered seems to be always true. But the public method getEl() seems to always be undefined when we need to wait, so I've changed it to check that.

PR #2638, opened 2018-04-23
fixes #2630 validates ncwms datetimepickers before adding values to filters

### LINKED ISSUE #2630: Bounding box overrides invalid date check for gridded data collections
### Steps to reproduce:

1.  Add any gridded data collection ie SSTAARS or satellite 

1.  On step 2 select dates that there is no data for by manually typing them (SSTAARS any date in 2015, satellite any date in 1991)

1.  Then draw a bounding box on the map

1.  Proceed to step 3 and try to download

### What should happen:

When user gets to step 3 the download button should be greyed out as there is no data available for the selected dates, or if such occurrence the download button is not greyed out the download should fail and indicate no data.

### What does happen:

1.  Updating Bbox loads bad dates
1.  Therefore download can be started and user gets a succeed download but it is empty. 

**Note csv at a point will give a failed download

### LINKED ISSUE #2631: NcWMS temporal filter doesn't correctly reset
Steps to reproduce:
Select satellite collection, select a date beyond the data range (i.e 1991) for the to and from fields. Move to step 3. Move back to step 2 and select clear subset.

Observed behaviour: 
Both fields reset, but the From field is highlighted red and references the previously set To value.

pmbohm:
Also seems to fix #2631

### DIFF
--- a/web-app/js/portal/details/NcWmsPanel.js
+++ b/web-app/js/portal/details/NcWmsPanel.js
@@ -88,11 +88,12 @@ Portal.details.NcWmsPanel = Ext.extend(Ext.Container, {
             if (layer.extraLayerInfo.zaxis.positive == true) {
                 title = OpenLayers.i18n('zAxisLabelPositiveUp');
                 this.multiplier = 1;
-            } else {
+            }
+            else {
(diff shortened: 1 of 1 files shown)


gsatimos:
looks good. Confirmed fix for both #2630 and #2631

PR #2645, opened 2018-05-24
Tracking GFI now clearer. fixes #2643

### LINKED ISSUE #2643: Spurious Google Analytics Event
## Steps to reproduce
Go to google analytics
Select BEHAVIOR
Select Events
Select Top Events 
Select (Event Category) DataCollections

## What happens
There's an event action named "Feature Information"

## What should happen
Depends what the event is.  Is it a download? If so it needs renaming. Is it a GFI  request? If so it should be under the "FeatureInfo" event category. Let me know what's going on with this and we can decide what to do about it.

analytics 5

pmbohm:


web-app/js/portal/utils/trackUsage.js
```diff
@@ -6,6 +6,9 @@ function trackUsage(category, action, label, collection, value) {
             dimension1: collection
         });
     }
+    else {
+        //console.log(String.format("ga('send', 'event', category={0}, action={1}, label={2}, value={3})",category, action, label, collection, value));
```
anguss00: A wild console log appears

pmbohm: @anguss00 I am now officially sick or re-writing this line every time that there is some work regarding tracking. I imagine if this line was available to be uncommented this bug would not even have come into existence.

anguss00: Fairo

PR #2646, opened 2018-05-24
Fixes #2644 copying current time between layers

### LINKED ISSUE #2644: Date/Time filter on step 2 only applied to currently selected layer of a collection
**What to do:**
1. In step 1 select CARS dataset
2. In step 2, `Subset` tab, select a different date to be displayed than the default one, say `2009-08-28 00:00:00:000 UTC `.
screenshot from 2018-05-18 14-54-04

3. In the `Layer` tab select a different layer than the default one, say `csiro_cars_weekly_url/PSAL`
4. Click on the WMS layer to bring the getFeature window.

**What happens:**
The getFeature window shows that the date is still set to its default value `Sat, 26 Dec 2009 00:00:00 GMT`
screenshot from 2018-05-18 14-54-58
If I go back to the default layer `csiro_cars_weekly_url/TEMP` the date/time hasn't changed from what we previously set in the `Subset` tab.

**What should happen:**
When a date/time as been selected for display in the `Subset` tab, it should apply to all layers in the `Layer` tab and not only the currently selected one. It is important that one should be able to switch from layer to layer for the same date/time.

pmbohm:


akashisama:
Fixes the feature date but not the map date.

To see what I mean:
1. In step 1 select CARS dataset
1. In step 2, Subset tab, set the 'To' date to 2009-06-26
1. In the Layer tab select 'csiro_cars_weekly_url/PSAL_anomaly'
<<< the map you see is from 2009-12-26
To confirm this:
1.  set the 'To' date to 2009-12-26, nothing changes cause you're already seeing that map
1. set the 'To' date to 2009-06-26, it looks completely different cause it's a different map (the map which should have appeared in the first place when you set the layer to'csiro_cars_weekly_url/PSAL_anomaly')

PR #2649, opened 2018-05-30
Handle numColorBands in style legend

anguss00:
Fixes https://github.com/aodn/chef-private/issues/2811 (actually portal bug)

### DIFF
--- a/src/test/javascript/portal/details/StylePanelSpec.js
+++ b/src/test/javascript/portal/details/StylePanelSpec.js
@@ -101,6 +101,38 @@ describe("Portal.details.StylePanel", function() {
             expect(getParameterByNameFromUrlString(urlString, "VERSION")).toEqual("1.1.0");
         });
 
+        it("should include numColorBands if it is defined", function() {
+
+            var layer = {
+                extraLayerInfo: {numColorBands: '24'},
+                url: "",
+                params: {}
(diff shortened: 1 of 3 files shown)


src/test/javascript/portal/details/StylePanelSpec.js
```diff
@@ -101,6 +101,38 @@ describe("Portal.details.StylePanel", function() {
             expect(getParameterByNameFromUrlString(urlString, "VERSION")).toEqual("1.1.0");
         });
 
+        it("should include numColorBands if it is defined", function() {
+
+            var layer = {
+                extraLayerInfo: {numColorBands: '24'},
```
ggalibert: Why is it hard coded to '24' here and line 112? numColorBands can be anything between 2 and 253 I think.

anguss00: Just for the purposes of Jasmine tests. The actual value is determined from the config

pmbohm:
Commented also on aodn/chef-private#2811

@angus @ggalibert It seems the functionality we are after is over-ridding the default number of bands that Thredds is capable of serving, and the legend should correspond to the map.

Should this be added to get map requests as well?
Should getFeatureInfo return the actual value or the banded value?

PR #2654, opened 2018-06-06
Fix PointTimeSeriesLatLonTest selenium test to match current functionality…

akashisama:
Evidently with the older depth service the latitude and longitude given in the FeatureInfo were whatever latitude and longitude were sent in the request. But with the new service the lat and lon represent the location of the depth data, which ends up being slightly different to the requested location.  So I've given the test a little leeway by checking if latitude and longitude are within 0.1 of their expected values. I also added a check to make sure a depth/elevation is actually being shown.

### DIFF
--- a/selenium/src/test/java/au/org/emii/portal/tests/PointTimeSeriesLatLonTest.java
+++ b/selenium/src/test/java/au/org/emii/portal/tests/PointTimeSeriesLatLonTest.java
@@ -21,9 +21,11 @@ public class PointTimeSeriesLatLonTest extends BaseTest {
         webElementUtil.clickButtonWithTitle("Add this collection");
 
         log.info("Validating that point time series lat/lon is not populated by default");
-        WebElement latitude1 = webElementUtil.findElement(By.xpath("//label[text()='Latitude']/following-sibling::input"));
-        Assert.assertNotNull(latitude1);
-        Assert.assertTrue("enter".equalsIgnoreCase(latitude1.getAttribute("value")));
+        WebElement latitudeTextBox = webElementUtil.findElement(By.xpath("//label[text()='Latitude']/following-sibling::input"));
+        Assert.assertNotNull(latitudeTextBox);
+        Assert.assertTrue("enter".equalsIgnoreCase(latitudeTextBox.getAttribute("value")));
+		//click on it so the people watching at home can see (BrowserStack recording)
+        latitudeTextBox.click();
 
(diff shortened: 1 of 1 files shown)


selenium/src/test/java/au/org/emii/portal/tests/PointTimeSeriesLatLonTest.java
```diff
@@ -46,17 +48,28 @@ public void validateLatLon(int x, int y) {
...
 
-        WebElement latitude2 = webElementUtil.findElement(By.xpath("//label[text()='Latitude']/following-sibling::input"));
-        Assert.assertNotNull(latitude2);
-        String lat = latitude2.getAttribute("value");
-        Assert.assertTrue(popupHtml.contains(lat));
+        WebElement latitudeTextBox = webElementUtil.findElement(By.xpath("//label[text()='Latitude']/following-sibling::input"));
+        Assert.assertNotNull(latitudeTextBox);
+        double lat =  Double.parseDouble(latitudeTextBox.getAttribute("value"));
+        double featureInfoLatitude = Double.parseDouble(parts[1]);
+        double latDifference = lat - featureInfoLatitude;
```
pmbohm: Does this difference need to be absolute?

akashisama: No, because it was checking for both negative and positive difference:
```latDifference > -0.1 && latDifference < 0.1```

But converting it to absolute is way cleaner so I've added a commit to do that instead.

File: selenium/src/test/java/au/org/emii/portal/tests/PointTimeSeriesLatLonTest.java [hunk 2]
```diff
@@ -46,17 +48,28 @@ public void validateLatLon(int x, int y) {
...
+        double featureInfoLatitude = Double.parseDouble(parts[1]);
+        double latDifference = Math.abs(lat - featureInfoLatitude);
+        Assert.assertTrue(latDifference < 0.1);
+
+        WebElement longitudeTextBox = webElementUtil.findElement(By.xpath("//label[text()='Longitude']/following-sibling::input"));
+        Assert.assertNotNull(longitudeTextBox);
+        double lon =  Double.parseDouble(longitudeTextBox.getAttribute("value"));
+        double featureInfoLongitude = Double.parseDouble(parts[3]);
+        double lonDifference = Math.abs(lon - featureInfoLongitude);
+        Assert.assertTrue(lonDifference < 0.1);
```
pmbohm: I can read it now

PR #2657, opened 2018-06-20
Selenium replaces Testrail C268

pmbohm:


pmbohm:
@kereid Do i need to do more to replace  Testrail C268 ?

PR #2670, opened 2018-07-17
Update "List of URLs" download option label to "List of HTTP URLs" to…

ggalibert:
… make the used protocol clearer and in preparation for a potential future support for a "List of OPeNDAP URLs". Also reflected this change in the retrieved file name.

Note that I didn't change the content of `downloadUrlListAction` in `web-app/js/portal/lang/en.js` since @jonescc believes this is used by google analytics and it would then be problematic.

Finally, not sure whether I should also update the "List of URLs" entries to "List of HTTP URLs" in `src/test/javascript/portal/cart/DownloadPanelSpec.js`.

### DIFF
--- a/web-app/js/portal/cart/BodaacDownloadHandler.js
+++ b/web-app/js/portal/cart/BodaacDownloadHandler.js
@@ -31,7 +31,7 @@ Portal.cart.BodaacDownloadHandler = Ext.extend(Portal.cart.DownloadHandler, {
                 handler: this._getUrlGeneratorFunction(),
                 handlerParams: {
                     downloadLabel: OpenLayers.i18n('downloadUrlListAction'),
-                    filenameFormat: '{0}_URLs.txt',
+                    filenameFormat: '{0}_HTTP_URLs.txt',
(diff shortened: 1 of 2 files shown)


jonescc:
The instances of "List of URLs" in src/test/javascript/portal/cart/DownloadPanelSpec.js refer to the google analytics label which hasn't changed, so no need to change that.

The build is failing, but its also failing in master and isn't related to these changes.  

@pblain, the code changes look fine, assuming we shouldn't change the label used for List of URL's in google analytics as this is cosmetic change.  

Not sure who needs to sign off on a label change.

pblain:
@jonescc - Correct - we definitely don't want to change the GA label. Changing it would screw up the stats. Once you choose GA event naming conventions you're stuck with them, which is why we need to get them right in the first place. We haven't always got them right unfortunately.

ggalibert:
Then if that's OK can someone merge this PR?

lbesnard:
@ccmoloney this is old but still relevant. Can someone look at this and merge if it's not breaking the portal?

PR #2682, opened 2018-07-26
Fixes #2664 bounding box

### LINKED ISSUE #2664: Bounding box not removed if drawn when gridded collection loading
Steps to reproduce:

Add any gridded collection to the map , when taken to step 2 draw a bounding box while collection is loading. Wait until collection finishes loading and then click reset bounding box.

What should happen:

Bounding constraints are reset and bounding box display removed from map

What does happen:

Bounding constraints are cleared but bounding box remains displayed on the map

craigrose:
Fixes #2664
Disables drawing of bounds until filter values are fully loaded.  Shows a 'wait' cursor when disabled.


jonescc:
Resolves original problem just fine.

PR #2692, opened 2018-08-08
DO NOT MERGE!!! removes proxyredirects

gsatimos:
This PR is to remove the proxy redirects from the portal in regard to this issue https://github.com/aodn/backlog/issues/773

This has not yet been tested with an actual geowebcache version 1.13.0 (or higher).


jonescc:
Changes look good to me by themself, but should be smoke tested in a stack with the upgraded geowebcache before merging.

akashisama:
I was just playing with this against 1.13.2. Functionally it seems to work perfectly. But a failed GetFeature request is being sent each time a new filter is  added... I'm not sure what this request is supposed to be doing.

The request looks like this: ```https://tilecache-systest.aodn.org.au/geowebcache/service/wms?typeName=imos:argo_profile_map&SERVICE=WFS&outputFormat=application/json&REQUEST=GetFeature&VERSION=1.0.0&CQL_FILTER=data_centre_name%20LIKE%20'CSIRO'&maxFeatures=1```

And the response that should be sent back (and is sent back if you send the request to a geoserver rather than a geowebcache) is:
```
{"type":"FeatureCollection","totalFeatures":150499,"features":[{"type":"Feature","id":"argo_profile_map.fid--566aaed5_16554ef7057_77f1","geometry":{"type":"Point","coordinates":[61.939,-25.111]},"geometry_name":"position","properties":{"platform_number":"1901119","juld":"2008-07-29T19:26:41.999Z","data_centre_name":"CSIRO","profile_processing_type":"Delayed","oxygen_sensor":false,"url":"IMOS/Argo/dac/csiro/1901119/profiles/D1901119_001.nc"}}],"crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:EPSG::4326"}}}
```

To replicate:
1. Select the ARGO Profiles layer
1. Open developer tools and watch network connections
1. Select a data centre from the "Data centre name" dropdown

What happens:
A GetFeature request is sent to geowebcache and geowebcache returns a 503 Service Unavailable status.

gsatimos:
@akashisama I belive @jonescc ran into this issue as well but I can't remember all the details about it. Will check with him on tues

jonescc:
No I think I missed this in my testing.  I think this is a check added to determine whether filters selected any data so that any appropriate error message can be shown on step 3.  @pmbohm?

jonescc:
Anyway, looks like this request used to work with geowebcache 1.11 but doesn't in 1.13.2.  Its a WFS request so looks like a grey area to me in terms of whether geowebcache should handle it, but we could probably switch to using a GetFeatureInfo instead - looks like it supports maxFeatures -http://docs.geoserver.org/stable/en/user/services/wms/reference.html#wms-getfeatureinfo.

jonescc:
Hmm x and y required. Other options are using DescribeLayer to find the actual wfs endpoint or get it fixed in geowebcache.

jonescc:
Scrap that, doesn't work with 1.11 either - geowebcache requires a layer name to be specified to be able to work out where the request should be routed.  If we want to remove proxy redirects then we need to work out another way of checking this.  DescribeLayer and then GetFeature perhaps if we can't use GetFeatureInfo.

pmbohm:
_"to determine whether filters selected any data so that an appropriate error message can be shown on step 3"_
Yes there was a recent bug and fix https://github.com/aodn/aodn-portal/pull/2702. Current behaviour is therefore to check against the map layer on parameters available in the map layer every time a filter is applied.

A getFeatureInfo request is possible, but will it get through Geowebcache?

lwgordonimos:
In any solution, it shouldn't be trying to sending non-WMS requests to Geowebcache.

jonescc:
Yes GetFeatureInfo is a wms request where a layer is specified

gsatimos:
@jonescc not sure who was PRing this atm but from the recent discussion it looks like it's not ready to merge?

jonescc:
May as well include the fix for the getfeature issue before we merge

jonescc:
Actually, fix can be made separate to this PR as long as its done next iteration.  I'll raise an issue for it.

PR #2705, opened 2018-08-16
Add 'AODN-Portal' as the user agent for all proxied requests

pmbohm:


pmbohm:
for https://github.com/aodn/backlog/issues/902

pmbohm:
also see https://github.com/aodn/geoserver-build/pull/280

lwgordonimos:
Could we add the version to the user agent as well? This seems to be the convention, and could be handy info?

e.g. `AodnPortal/v4.38.10`

pmbohm:
Its possible
I think the convention of giving a version is so that the web service can adjust is response depending on the user-agent IE: chrome V 67.0 vs MS Internet Explorer v1

Issue #2684, opened 2018-07-31
Tab key-press missing a box in date selection


ocehugo:
Reproduction:

```
1. Step-1 search for ```radar delayed capricorn velocity```
2. Select/go step 2
3. The date box "From" says 2007-10-15 23:30 UTC.
4. Select the date box "From" and type 2013-02-01.
5. Press tab
6. #2683 happen but also the "hour" box is not selected and the jump is towards the "To" date box.
``` 

Expectation:

The "tab key event" in step 3. above should stop at the "hour" box. Instead, the current behaviour is ignoring the hour box and jump directly to the "To" date box right below.

Although fewer users would want to clip anything in the "hour" timescales, the expected behaviour in web forms navigation is to use Tab keys for next toolboxes.
I keep pressing tab two times (to jump to the date and ignore hours) but end up having to press shift-tab to go back because the "hour box" is ignored.

pblain:
@ocehugo - the hour box is a dropdown - no typing allowed. I don't think it would make sense to tab into it if you can't type. What do you think?

ggalibert:
@pblain it would be great to be able to tab into it because although we cannot type in the hour box, we can scroll through hours via arrow keys (see radar collections for example).

pblain:
@ggalibert , @ocehugo - OK, I see what you mean. Fair enough then.

gsatimos:
looks like what @ocehugo mentions above is the cause of the issue; ajax request to get `getFilterValues` (retrieves a list of available times available for the given date) doesn't return in time fast enough for the field to become re-enabled and receive the tab. working on this issue now.

pmbohm:
Looking at the interface it seems if the time component is unavailable to be selected then it is unavailable to be selected. 
The current tab functionality works well if the time pickers are allowed to load. 
In the case of a power user (@ocehugo ) quickly moving through the form via tab and going to far, they (a power user) also know the way to go back. Win win.
I don't think this is a bug. I personally like that the time dropdown can be entered via tab and navigated via arrow keys and wouldn't want to see this lost.
Additionally we cant use arrow selection in the date components because that currently opens the date picker.

gsatimos:
closing a PR I had open for this as a work-in-progress. It's only a partial solution but may be useful when this issue gets visited again: https://github.com/aodn/aodn-portal/pull/2708

PR #2708, opened 2018-08-20
WIP do not merge: date field tab moved to time field

gsatimos:
@pmbohm would like your assistance with this when you're available. This PR is regarding https://github.com/aodn/aodn-portal/issues/2684

The intention is to get the focus to move to the time field after a date field is selected. Currently I have it working for the case where the date entered has 1 or more times available. However work out a way to solve in the case where the date has no available times.


pmbohm:
My comments on this are at https://github.com/aodn/aodn-portal/issues/2684

gsatimos:
this remains a partial fix for issue https://github.com/aodn/aodn-portal/issues/2684

PR #2713, opened 2018-08-29
Adjust spatial filter display on Step 3.  Fixes #2704

### LINKED ISSUE #2704: inconsistent display of spatial filter on step 3
This issue tracker is only for AODN Portal issues.

### Steps to reproduce the issue
1. select a collection on step 1 e.g. `IMOS - SRS Satellite - SST L3S - 01 day composite - night time`
2. draw bounding box on step 2 and move to step 3 (observe the format of the spatial filter displayed to the user)
3. go back to step 2 and select `Point Timeseries` then move to step 3 (observer the format of the spatial filter)

### Expected behaviour
Ideally, the format of the coordinates should be the same

### Actual behaviour
Bounding box sample: `Spatial: 116.807W -16.744S 119.619E -14.371N`
Timeseries sample: `Timeseries at Lat/Lon: -20.4355,115.9277`

Note that the bounding box unnecessarily uses `N/S` and `E/W` which is redundant since this is already explicit because coordinates work with `+/-` to denote this.

Understandably, the `N/S E/W` notation might be useful for the user in which case they should _both_ use it (and be consistent)

ccmoloney:
Updated display of subset parameters selected on Step 3 to be the following format:

Spatial: From Lat/Lon -43.639, 132.715 to Lat/Lon -35.377, 136.582
Temporal: 2009-Jan-01-00:00-UTC to 2009-Jan-08-00:00-UTC
Depth: 0m to 5000m

### DIFF
--- a/grails-app/views/_js_includes.gsp
+++ b/grails-app/views/_js_includes.gsp
@@ -99,6 +99,7 @@
     <script type="text/javascript" src="${resource(dir: 'js/portal/filter', file: 'GeometryFilter.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/filter', file: 'NumberFilter.js')}"></script>
     <script type="text/javascript" src="${resource(dir: 'js/portal/filter', file: 'StringFilter.js')}"></script>
+    <script type="text/javascript" src="${resource(dir: 'js/portal/filter', file: 'StringDepthFilter.js')}"></script>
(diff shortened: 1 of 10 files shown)


ccmoloney:
Fix for #2704

gsatimos:
@ccmoloney I ran the jasmine tests locally to verify found 2 to be failing:
`Portal.cart.NcWmsInjector getDataFilterEntry indicates bounds properly created.`
`Portal.filter.Filter classFor returns expected types for simple examples (not exhaustive).`

Can you try running them again locally on yours?

ccmoloney:
@gsatimos Should be good to go now.

PR #2716, opened 2018-08-30
Ala improvements

pmbohm:


pmbohm:
[ala.xml.txt](https://github.com/aodn/aodn-portal/files/2346863/ala.xml.txt)
demo ALA metadata.

Issue #2706, opened 2018-08-17
Both Pan control and bounding box selected


kereid:
This issue tracker is only for AODN Portal issues.

### Steps to reproduce the issue
1. Add any collection.
2. When taken to step 2 draw a bounding box and then click the pan control button. 
3. Remove the collection and go back to step 1
4. Re-add the collection

### Expected behaviour
After re-adding and user taken back to step 2 the pan control should still be active

### Actual behaviour
Both bottons are active

This is in Systest

ccmoloney:
@kereid Should the bounding box remain when you remove the collection?

kereid:
@ccmoloney yes the bounding box should remain when you remove the collection as the bounding box is applied to anything on the map so "in real life" if someone removed the collection they may have more on the map that the bounding box should still be applied to

ccmoloney:
Should be resolved by: https://github.com/aodn/aodn-portal/pull/2718

PR #2718, opened 2018-09-04
Fix for #2706 - pan & bounding box controls both active.

ccmoloney:


web-app/js/portal/ui/openlayers/SpatialConstraintMap.js
```diff
@@ -1,6 +1,9 @@
 OpenLayers.SpatialConstraintMap = OpenLayers.Class(OpenLayers.Map, {
 
     initialize: function(div, options) {
+        if(div) {
```
pmbohm: This is a bit weird here. I'm not sure why `initialise` is taking two arguments? Only `options` or null are supplied
```
OpenLayers.SpatialConstraintMap = OpenLayers.Class(OpenLayers.Map, {

    initialize: function(options) {

        if(options) {
            this.navigationControl = options.navigationControl;
        }
        ......
```

web-app/js/portal/ui/openlayers/SpatialConstraintMap.js
```diff
@@ -105,12 +108,32 @@ OpenLayers.SpatialConstraintMap = OpenLayers.Class(OpenLayers.Map, {
...
+            if(this.navigationControl.panControl.active) {
+                this.spatialConstraintControl.deactivate();
+            } else {
+                this.spatialConstraintControl.activate();
+            }
+        }
+    },
+
+    disableControl: function() {
+        console.log("disableControl");
```
pmbohm: console.log

web-app/js/portal/ui/openlayers/SpatialConstraintMap.js
ccmoloney: Yeah - looks weird, but it's old weird stuff - not new weird stuff  :)

pmbohm:
BTW Your commit is called `Fix Jasmine tests.` Maybe you meant to include it as another commit and squashed it. Git log will show this which is a bit cryptic.
Also you should comment:
Fixes #2706 and then it will link to the issue itself and self close when merged

web-app/js/portal/ui/openlayers/SpatialConstraintMap.js
pmbohm: Yep old and now time to fix

ccmoloney:
Fixed commit message.

web-app/js/portal/ui/openlayers/SpatialConstraintMap.js
ccmoloney: Nope - it has two arguments because it is over-riding the base class initialize method.

pmbohm: Erk. Because of this, these lines aren't needed
```
if(div) {
            this.navigationControl = div.navigationControl;
}
```
 `this.navigationControl` gets created by the `OpenLayers.Map.prototype.initialize.apply(this, arguments);`

ccmoloney: it is needed - because that's how i put a handle to the navigation control in here - so that i can check it's state in the enableControl & disableControl methods.

ccmoloney:
Phil's commit resolves this.

PR #2721, opened 2018-09-05
Google Analyitics for layerSwitcher layers

pmbohm:
For 
https://github.com/aodn/backlog/issues/919

Tracks data and baseLayer visibility switching

### DIFF
--- a/src/test/javascript/portal/details/LayerControlPanelSpec.js
+++ b/src/test/javascript/portal/details/LayerControlPanelSpec.js
@@ -70,7 +70,7 @@ describe("Portal.details.LayerControlPanel", function() {
 
             layerSelectComponent.fireEvent('change', layerSelectComponent, layerSelectComponent.items[0]);
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'changeLayerTrackingAction',
+                'change layer',
                 'Data Collection Title',
                 'first layer'
             );
@@ -121,7 +121,7 @@ describe("Portal.details.LayerControlPanel", function() {
             layerControlPanel._visibilityButtonChecked(null, true);
 
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'layerControlTrackingActionVisibility',
+                'visibility (overlay)',
                 'on',
                 'Data Collection Title'
             );
@@ -131,7 +131,7 @@ describe("Portal.details.LayerControlPanel", function() {
             layerControlPanel._visibilityButtonChecked(null, false);
 
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'layerControlTrackingActionVisibility',
+                'visibility (overlay)',
                 'off',
                 'Data Collection Title'
             );
--- a/web-app/js/portal/ui/openlayers/layer/LayerSwitcher.js
+++ b/web-app/js/portal/ui/openlayers/layer/LayerSwitcher.js
@@ -1,0 +1,26 @@
+Ext.namespace('Portal.ui.openlayers');
+
+Portal.ui.openlayers.LayerSwitcher = Ext.extend(OpenLayers.Control.LayerSwitcher, {
+
+    onButtonClick: function(evt) {
+
+        var button = evt.buttonElement;
+
+        if (button._layer) {
+
+            var layer = this.map.getLayer(button._layer);
+
+            // tracking baselayer being switched "on" or the dataLayers chosen visibility
+            var intendedVisibility = (button.type == "radio") ? true : !layer.visibility;
+
+            var state = (intendedVisibility) ? "on" : "off";
+            var type = (layer.isBaseLayer) ? "baseLayer" : "dataLayer";
+
+            var action = String.format("{0} ({1})", OpenLayers.i18n('layerControlTrackingActionVisibility'), type);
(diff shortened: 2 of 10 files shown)


src/test/javascript/portal/details/LayerControlPanelSpec.js
```diff
@@ -70,7 +70,7 @@ describe("Portal.details.LayerControlPanel", function() {
 
             layerSelectComponent.fireEvent('change', layerSelectComponent, layerSelectComponent.items[0]);
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'changeLayerTrackingAction',
+                'change layer',
                 'Data Collection Title',
```
jonescc: Isn't this in the wrong order?

src/test/javascript/portal/details/LayerControlPanelSpec.js
```diff
@@ -121,19 +121,17 @@ describe("Portal.details.LayerControlPanel", function() {
             layerControlPanel._visibilityButtonChecked(null, true);
 
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'layerControlTrackingActionVisibility',
-                'on',
-                'Data Collection Title'
+                'visibility (overlay)',
+                'on - Data Collection Title'
```
jonescc: Should this be in a different format to the other trackLayerControlUsage events (and different to what we've recorded previously)?

src/test/javascript/portal/details/LayerControlPanelSpec.js
```diff
@@ -121,19 +121,17 @@ describe("Portal.details.LayerControlPanel", function() {
...
 
         it('sends google analytics tracking when unchecked', function() {
             layerControlPanel._visibilityButtonChecked(null, false);
 
             expect(window.trackLayerControlUsage).toHaveBeenCalledWith(
-                'layerControlTrackingActionVisibility',
-                'off',
-                'Data Collection Title'
+                'visibility (overlay)',
+                'off - Data Collection Title'
```
jonescc: And here.

src/test/javascript/portal/details/LayerControlPanelSpec.js
pmbohm: Yes after a discussion with @anguss00 we captured for the `visibility` action the amount of times that "on" or "off" was selected. Really what we need now is to track when a user changes the visibility of a layer, what sort of layer it was therefore what part of the GUI was used.
So now we will have three new actions for visibility - baselayer/datalayer/overlay, and then for the label prefix the action state (on or off) with the layername. So in the case of the new datalayer we can see which datalayer was selected. 
Additionally this pull request only tracks that a baselayer was turned on, not that another baselayer was turned off.

src/test/javascript/portal/details/LayerControlPanelSpec.js
pmbohm: `action` then `label`
The action is now a string not i18n

web-app/js/portal/ui/openlayers/layer/LayerSwitcher.js
```diff
@@ -0,0 +1,32 @@
+Ext.namespace('Portal.ui.openlayers');
+
+Portal.ui.openlayers.LayerSwitcher = Ext.extend(OpenLayers.Control.LayerSwitcher, {
+
+    constructor: function(options) {
+        OpenLayers.Control.LayerSwitcher.prototype.constructor.apply(this, arguments);
+        this.layerStates = [];
```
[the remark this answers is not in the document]
pmbohm: cool i'll give it a go. The OpenLayers.Control.LayerSwitcher had no constructor but i assume then its from the base class

web-app/js/portal/utils/trackUsage.js
```diff
@@ -7,7 +7,7 @@ function trackUsage(category, action, label, collection, value) {
         });
     }
     else {
-        //console.log(String.format("ga('send', 'event', category={0}, action={1}, label={2}, value={3})",category, action, label, collection, value));
+        console.log(String.format("ga('send', 'event', category={0}, action={1}, label={2}, value={3})",category, action, label, collection, value));
```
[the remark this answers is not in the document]
pmbohm: nope :(

src/test/javascript/portal/details/LayerControlPanelSpec.js
jonescc: Yes I can see that, but the parameters expected for trackLayerControlUsage are now action, label, collection - these parameters don't match that.  See https://github.com/aodn/aodn-portal/blob/26a62c3a0379398908fc6d01cdf29871e16e6cf7/web-app/js/portal/utils/trackUsage.js#L69

src/test/javascript/portal/details/LayerControlPanelSpec.js
jonescc: Yes, but I don't understand why we have to bundle the collection title in the label parameter rather than using the separate collection parameter.

src/test/javascript/portal/details/LayerControlPanelSpec.js
pmbohm: Ok so i now can see on the GA page how to see the collection dimension
Back to the drawing board

pmbohm:
@jonescc New push

Issue #2711, opened 2018-08-24
GetFeature requests for empty download failing with geowebcache 1.13.2


jonescc:
### Steps to reproduce the issue
1. Pick argo collection and go to step 2
2. Select a filter - e.g. platform number

### What happens
Get an error in the console:

    GET https://portal-systest.aodn.org.au/proxy?url=https%3A%2F%2Ftilecache-systest.aodn.org.au%2Fgeowebcache%2Fservice%2Fwms%3FtypeName%3Dimos%3Aargo_profile_map%26SERVICE%3DWFS%26outputFormat%3Dapplication%2Fjson%26REQUEST%3DGetFeature%26VERSION%3D1.0.0%26CQL_FILTER%3Dplatform_number%2520LIKE%2520%271900047%27%26maxFeatures%3D1&_dc=1535071783251 503 (Service Unavailable)

### What I expect to happen

No errors in the console

jonescc:
Refer https://github.com/aodn/sumologic/pull/16 for more info.

jonescc:
So discussing this with @craigrose, I think the best/most correct option is to use any WFS download service defined for the collection to determine whether there is anything to download - i.e. use the WFS download url and set maxFeatures to 1.  This should return a result in a reasonable amount of time otherwise we have a problem with the wfs download which should have been picked up by nagios.  
Tried this with some ordered layers which timeout when asking for all records and they worked OK.

jonescc:
The risk with this option is that a WFS request may take longer to return a result than it should, there are a lot of filter changes and the server is tied up trying to service the is there any data requests instead of drawing tiles or servicing actual download requests.  GetFeatureInfo can be used to identify whether geoserver would be able to find any features to draw for the filter criteria and is a likely to be a more performant option.

jonescc:
Fixed by https://github.com/aodn/aodn-portal/pull/2724

PR #2724, opened 2018-09-06
Using GetFeatureInfo (instead of GetFeature) to check if any data wil…

craigrose:
…l be returned by a set of filters and displaying a message if there is none.

### DIFF
--- a/web-app/js/portal/filter/ui/FilterGroupPanel.js
+++ b/web-app/js/portal/filter/ui/FilterGroupPanel.js
@@ -4,7 +4,7 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Portal.filter.ui.GroupPanel, {
 
     _filtersLoaded: function(filters) {
         var filterPanels = [];
-        var filterService  = new Portal.filter.FilterService();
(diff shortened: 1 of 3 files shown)


craigrose:
fixes https://github.com/aodn/aodn-portal/issues/2711

web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -63,7 +63,10 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Portal.filter.ui.GroupPanel, {
     _handleGetFeatureRequestResults: function(results) {
         if (results.status == 200) {
             var res = Ext.util.JSON.decode(results.responseText);
-            this.dataCollection.totalFilteredFeatures = (res && res.totalFeatures >= 0) ? res.totalFeatures: undefined;
+            if (res && res.totalFeatures == "unknown") {
+                res.totalFeatures = res.features.length;
```
jonescc: We've just lost the returned totalFeatures.  Why not use a temporary variable instead of modifying the response.

PR #2728, opened 2018-09-27
Let geoserver know what services we are talking to

jonescc:
Its used by geoserver to determine what parsers to use and its the only way we can tell geoserver to use different parsers for the WMS palette request parameter and the ncwms palette request parameter


jonescc:
And besides, its consistent with what we have to do for every other request.

PR #2738, opened 2018-10-15
Proof of concept for retrieving filter config from a url

akashisama:
for https://github.com/aodn/backlog/issues/945#issuecomment-429200369

I haven't added tests for the new functionality yet since it's just proof of concept. I figure the functionality might change.

I've changed the config to show how it's supposed to be configured, but that probably shouldn't be merged into master, since geoserver-123 doesn't seem to work as coregeoserver for retrieving filter values (should we enable wps?)

I made sure to keep instances of CoreGeoserverServer generic (in the sense that each can accept requests for any server). But I wonder if we should actually just refactor to make each instance of a WmsServer attach to a specific server, rather than having server sent as a parameter to each function... cause the thing is, the way the code is currently set up,  a new WmsServer is created for each server, so a particular instance is never going to receive requests for more than one server...

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -172,22 +172,27 @@ baselayerServer = [
 proxyConnectTimeout = 2000
 
 
+filtering {
+    baseUrl = "https://raw.githubusercontent.com/aodn/filter-config/test"
+}
+
 // This array should be populated from chef config
 knownServers = [
     [
         uri: 'http://geoserver-123.aodn.org.au/geoserver/wms',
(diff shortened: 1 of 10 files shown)


jonescc:
Would be good if we could merge it.  Best not to have long running feature branches if we don't need to. 
Would need to configure the proof of concept stack (sandbox?) to have wps enabled.

+1 for refactoring into a separate server - that would have been my review comment.

jonescc:
OK, refactored and added unit tests.  For review by someone else.

PR #2739, opened 2018-10-17
Java8 unit test fixes

ccmoloney:


jonescc:
Looks good.  Might be time to update travis to use openjdk 8 to build the project.  That way we know the pull request has done its job.

https://github.com/aodn/aodn-portal/blob/master/.travis.yml#L6

PR #2741, opened 2018-10-23
Use wms layer name to lookup filters in layer filter store

jonescc:


web-app/js/portal/data/DataCollection.js
```diff
@@ -46,15 +46,23 @@ Portal.data.DataCollection = function() {
...
 
         return {
             server: layer.url,
             serverType: serverType,
-            layer: layerName
+            layer: filterLayer
         };
     };
 
+    constructor.prototype._getFilterLayer = function(serverType, layer) {
```
pmbohm: This is actually the layerName not the layer

jonescc: Renamed to filterLookupName

jonescc:
Good to go @pmbohm

PR #2742, opened 2018-11-01
Working getFeature check on user filter selection

pmbohm:
Geowebcache, we would inform the user the filter selection they had just made would result in an empty download.  Geowebcache does not pass through the getFeatureInfo checks and therefore broke the check. 

This changes the getFeatureInfo check previously being made on the WMS Layer into a GetFeature count check for Geoserver servers.

The implied feature request in aodn/backlog#49 would have required extensive changes to the filters to have all filters reload available options in response to a user applying a filter. complications would have arisen also from users adding a global spatial filter after layer filters are already applied.

Fixes aodn/backlog#49

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -184,8 +184,16 @@ knownServers = [
         type: 'GeoServer',
         csvDownloadFormat: 'csv-with-metadata-header',
         urlListDownloadSubstitutions: [
-            '^': 'http://data.aodn.org.au/'
+                '^': 'http://data.aodn.org.au/'
         ]
+    ],
+    [
+        uri: 'http://geoserver-rc.aodn.org.au/geoserver/wms',
+        wmsVersion: '1.1.1',
+        type: 'GeoserverCore',
+        csvDownloadFormat: 'csv-with-metadata-header',
+        filtersDir: "imos-geoserver",
+        wpsUrl: 'http://geoserver-rc.aodn.org.au/geoserver/wps'
     ],
     [
         uri: 'http://nonprod.marine.ga.gov.au/geoserver/wms',
--- a/test/unit/au/org/emii/portal/wms/CoreGeoserverServerTests.groovy
+++ b/test/unit/au/org/emii/portal/wms/CoreGeoserverServerTests.groovy
@@ -98,18 +98,9 @@ class CoreGeoserverServerTests extends GrailsUnitTestCase {
     }
 
     void testLookup() {
-        def describeLayerCalledCount = 0
+        def result = coreGeoserverServer.getLayerInfo("http://geoserver-rc.aodn.org.au/geoserver/wms", "imos:argo_profile_map")
+        def expected = ["owsType": "WFS", "wfsUrl": "http://geoserver-rc.aodn.org.au/geoserver/wfs?", "typeName": "imos:argo_profile_map"]
 
-        coreGeoserverServer.metaClass._describeLayer = { server, layer -> describeLayerCalledCount++ ; return validDescribeLayerResponse }
-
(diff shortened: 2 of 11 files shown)


pmbohm:
Tested by declaring geoserver-rc.aodn.org.au below to be a type 'GeoserverCore' , 'GeoserverFilterConfig' and 'GeoServer'

```
development {
        geonetwork.url = "http://catalogue-rc.aodn.org.au/geonetwork"
...

knownServers = [
[
        uri: 'http://geoserver-rc.aodn.org.au/geoserver/wms',
        wmsVersion: '1.1.1',
        type: 'GeoserverCore',
        csvDownloadFormat: 'csv-with-metadata-header',
        filtersDir: "imos-geoserver",
        wpsUrl: 'http://geoserver-rc.aodn.org.au/geoserver/wps'
    ],
```

test/unit/au/org/emii/portal/wms/CoreGeoserverServerTests.groovy
```diff
@@ -101,14 +102,14 @@ class CoreGeoserverServerTests extends GrailsUnitTestCase {
 
         coreGeoserverServer.metaClass._describeLayer = { server, layer -> describeLayerCalledCount++ ; return validDescribeLayerResponse }
 
-        def result = coreGeoserverServer.getLayerInfo("https://geoserver.aodn.org.au/geoserver/wms", "imos:argo_profile_map")
+/*        def result = coreGeoserverServer.getLayerInfo("https://geoserver.aodn.org.au/geoserver/wms", "imos:argo_profile_map")
```
pmbohm: what happened to https://geoserver.aodn.org.au/geoserver/wms

craigrose:
@pmbohm I've run this branch up locally but can't seem to see any change in behaviour - thinking I must be doing something wrong.

grails-app/conf/Config.groovy
```diff
@@ -184,9 +184,17 @@ knownServers = [
...
-            '^': 'http://data.aodn.org.au/'
+                '^': 'http://data.aodn.org.au/'
         ]
     ],
+    [
+        uri: 'http://geoserver-rc.aodn.org.au/geoserver/wms',
+        wmsVersion: '1.1.1',
+        type: 'GeoserverCore',
+        csvDownloadFormat: 'csv-with-metadata-header',
+        filtersDir: "imos-geoserver",
```
jonescc: Not required for GeoserverCore

jonescc:
Looks like the application/json info_format isn't supported by tilecache.  Only text/html (used by the Feature Info popup), text/plain and application/vnd.ogc.gml (gml 2) are supported.  Tilecache also requires the SRS parameter to be passed (get a null pointer exception otherwise which is the error we're getting in production).   So the only option to get this to work in prod (where we use tilecache) using GetFeatureInfo  is to use text/plain or gml 2 which we've had problems with in the past (GA).  Using describelayer to find the the associated wfs service is a reasonable alternative.

web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -83,20 +80,21 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Portal.filter.ui.GroupPanel, {
         }
     },
 
-    // uses the WMS map layer
-    _getFeatureUrlGeneratorFunction: function() {
+    _getFeatureParams: function() {
 
-        var builder = new Portal.filter.combiner.MapCqlBuilder({
+        var builder = new Portal.filter.combiner.DataDownloadCqlBuilder({
```
jonescc: Will have issues with download only filters?

pmbohm: its getting `_filtersWithValues`

gsatimos:
just grabbed this one off the board and now seeing that there's a bit of history here. @pmbohm can you talk me through this fix when you have a sec?

PR #2749, opened 2018-11-15
String filters now multi select

pmbohm:
This is for a "small task" on Jacs board


pmbohm:
For https://github.com/aodn/backlog/issues/20

PR #2756, opened 2018-12-06
JS matching knownServers done without HTTP protocol

pmbohm:
Fixes aodn/issues#349

### DIFF
--- a/src/test/javascript/portal/data/ServerSpec.js
+++ b/src/test/javascript/portal/data/ServerSpec.js
@@ -2,6 +2,8 @@ describe('Portal.data.Server', function() {
     describe('getInfo', function() {
         var server1Info;
         var server2Info;
+        var server3Info;
+        var server4Info;
 
         beforeEach(function() {
             Ext.namespace('Portal.app.appConfig');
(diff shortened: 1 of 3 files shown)


src/test/javascript/portal/data/ServerSpec.js
```diff
@@ -26,7 +39,15 @@ describe('Portal.data.Server', function() {
...
-        it('returns undefined for unknown server', function() {
+        it('not blocked by protocol ', function() {
+            expect(Portal.data.Server.getInfo('http://maybesecureserver3')).not.toBe(Portal.data.Server.UNKNOWN);
+        });
+
+        it('is blocked', function() {
+            expect(Portal.data.Server.getInfo('http://rogueServer')).toBe(Portal.data.Server.UNKNOWN);
+        });
+
+        it('returns server regardless of HTTP protocol', function() {
```
jonescc: Not sure we're testing this here.  The result is exactly the same as for the above two test cases.

pmbohm: Yes agreed. Ive removed the duplication now.

PR #2757, opened 2018-12-10
GA checkFeatureCount

pmbohm:
A solution for testing if users will attempt an empty download for all WCS layers. 
The only lightweight option is the same as ncWMS checking.

The WCS 'GetCoverage' request is the only other option which is essentially downloading the data and seeing if its empty.


web-app/js/portal/filter/ui/FilterGroupPanel.js
```diff
@@ -46,6 +46,34 @@ Portal.filter.ui.FilterGroupPanel = Ext.extend(Portal.filter.ui.GroupPanel, {
...
+            success: this._handleGetFeatureRequestMethod
+        });
+    },
+
+    _handleGetFeatureRequestMethod: function(results) {
+
+        (JSON.parse(results.responseText).owsType == "WCS") ?  this.getWcsFeatureCount(): this.getWfsFeatureCount();
+    },
+
+    getWcsFeatureCount: function() {
```
jonescc: Were not really getting counts here - would it be clearer if we made these methods return booleans/renamed them to reflect what they are actually doing - checking if there is any data for requested filters

pmbohm: True. Changes now pushed @jonescc

jonescc:
featuresAvailable is much clearer!

PR #2769, opened 2018-12-20
adds cookie for detecting failed downloads

gsatimos:
- see comment regarding this: https://github.com/aodn/issues/issues/355

### DIFF
--- a/web-app/js/jquery/jquery.fileDownload.js
+++ b/web-app/js/jquery/jquery.fileDownload.js
@@ -327,6 +327,12 @@ $.extend({
         setTimeout(checkFileDownloadComplete, settings.checkInterval);
 
 
+        function deleteCookie(name, path, domain) {
+            var cookieData = name + "=; path=" + path + "; expires=" + new Date(0).toUTCString() + ";";
+            if (domain) cookieData += " domain=" + domain + ";";
+            document.cookie = cookieData;
+        }
+
         function checkFileDownloadComplete() {
             //has the cookie been written due to a file download occuring?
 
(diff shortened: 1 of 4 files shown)


web-app/js/jquery/jquery.fileDownload.js
```diff
@@ -327,6 +327,12 @@ $.extend({
         setTimeout(checkFileDownloadComplete, settings.checkInterval);
 
 
+        function deleteCookie(name, path, domain) {
```
pmbohm: There is a new `jQuery File Download Plugin v1.4.5` does it do what you want?
It seems a little wrong to mod this plugin

gsatimos: I had a look at it and it doesn't do what we want. Personally, and this might just be my opinion but if we're adding code directly to our repo in this manner instead of pulling it in from it's own source then we're basically saying 'this is now part of our code' and therefore it can be editable just like the rest of our code. Otherwise we should just be managing it properly as a third-party dependency.

lwgordonimos: Not necessarily. "Vendoring" third party libraries into your codebase is very common, but doesn't automatically make them fair game for forking. It just makes upgrading jquery a nightmare in future! There has to be a way to achieve this without modifying 3rd party libs.

gsatimos: um, not really sure I know what the answer is to that. Maybe submit a PR and get the feature added to the library?

akashisama: Would it be possible to do something like the overrides we have in the prototypes folder? https://github.com/aodn/aodn-portal/tree/master/web-app/js/portal/prototypes

gsatimos: thanks @akashisama I'll see if that would work

ccmoloney:
The solution George has implemented solves the issue within the framework of the current codebase.  It might not be ideal - but we are operating within the constraints of the code as it stands.  If we feel strongly that this is a section of the system that needs to be re-designed/re-implemented then we should raise a backlog item to address it as an exercise in tackling technical debt.  For now the user-facing issue has been rectified.

PR #2779, opened 2019-01-16
Fixes #2776 by testing ncWMS layers using the set map spatial extent

### LINKED ISSUE #2776: No data to download prevention not working for bounding box on gridded collections
This is in production - may be some remnant of "the disaster day"? Noted when a user email came through.

Steps to reproduce:

Add a gridded collection to the map such as https://portal-systest.aodn.org.au/search?uuid=d7a14921-8f3f-4522-9a54-e7d1df969c8a on step 3 zoom out to where there is no data displayed and draw a bounding box. Also select the Point at timeseries box for comparison and click where bounding box is.

What should happen:

Get an error message indicating there is no data to download so that user cannot even try 

What does happen:

There is no error and the user can set off the download only for the download to fail with not a very good message in the email.

Point time series in the same spot does have an error message though...

pmbohm:


jonescc:
No unit test to pick this up?  No wonder we broke it in the first place.

jonescc:
Suggest we add a unit test.  Making sure we don't break things is what unit tests are good for.

PR #2780, opened 2019-01-24
Using Geowebcache for static

pmbohm:


pmbohm:
Not to be merged until http://geoserver-static.aodn.org.au/geoserver/gwc/service/wms is alive
Part of aodn/backlog#1035

pmbohm:
@lwgordonimos New push

lwgordonimos:
I've tweaked this one a bit to work with the "integrated caching" feature. The config works fine in my stack, and is able to be overriden in Portal.groovy effectively but I'm not 100% on whether this is the best place to inject this "tiled=true" parameter(which is required for GWC integration). Open to suggestions as to whether this is better placed elsewhere.

akashisama:
What made you change to using the integrated caching rather than the internal gwc directly?

lwgordonimos:
This is only the default values in the base config, because I was thinking how to make the transition smooth. This means that it will work regardless of the GWC situation, but when we want to use it in a real environment, we can point to GWC directly.

It makes this change compatible with both GWC enabled and non-cached geoserver-statics, so is trying to be conservative

lwgordonimos:
i.e, if we configure the URL as `http://geoserver-static.aodn.org.au/geoserver/wms` it will work in pretty much any scenario (albeit slowly if totally uncached), but if we set it to `http://geoserver-static.aodn.org.au/geoserver/gwc/service/wms` it will explode if the Geoserver doesn't have GWC enabled.

akashisama:
The defaultParams seems like a good place to put "tiled=true" to me. Tested it out against current non-gwced static geoserver and it worked as expected so it should be safe to put into production.

Only annoying thing is that with this extra parameter added, the tile requests currently in the squid cache will no long be relevant. Would be good to make sure the requests with "tiled=true" are cached before putting this into prod (assuming this gets in before the swap over to using static-gwc).

akashisama:
The reasons the preview images aren't initially in the cache, and are therefore slow to load first time are:
1. The requests are for jpeg files, which we don't seed. I see no reason why we shouldn't make these request png to match with the requests for the main map.
1. Some of the previews show first or second level zoom which we don't seed, but they'd add almost no extra time to the seeding since the top two zoom levels have barely any tiles.

lwgordonimos:
Yeah I'd started to look at the JPG thing on Friday and was thinking the same. They should just be PNG.
 Sounds reasonable with seeding. It seems to take about 30 mins (ish) to seed static with the previous;y configured zoom levels, so a few more minutes would be fine since it's so rarely needed.

lwgordonimos:
@akashisama I've updated the PR to have the minimaps requested as PNG format. The zoom level stuff we can deal with in the seeder config.

PR #2782, opened 2019-02-11
layer switcher control open by default

gsatimos:
https://github.com/aodn/backlog/issues/972


gsatimos:
closing this PR as there's no point having a one-liner PR around since 12 Feb.

PR #2784, opened 2019-02-19
Server timeout for downloads message updated

### LINKED ISSUE #2783: Error message good but not useful
This was seen in Systest twice in a row then the download worked.
download error

Steps to reproduce (maybe)

Add Marvl3 Aus shelf data to the map draw a bounding box - proceed to step 3 and select Download as Temperature observations csv. 

What does happen:

Download failed to start and got the error: an error occurred while downloading this file

What should happen:

Download fails to start and get an informative error - either with information that helps the user understand why or information the user can send to the AODN helpdesk if they so choose and then we can use the information to determine which happened

Note: this addition of the message is a great addition far better than the download not starting and just timing out!

pmbohm:
Fixes #2783

### DIFF
--- a/web-app/js/jquery/jquery.fileDownload.js
+++ b/web-app/js/jquery/jquery.fileDownload.js
@@ -346,11 +346,12 @@ $.extend({
             //check if cookie is set to indicate failure
             if (settings.cookieFailedName !== undefined) {
                 if (document.cookie.toLowerCase().indexOf(settings.cookieFailedName.toLowerCase()) > -1) {
-                    internalCallbacks.onFail('an error occurred while downloading this file', fileUrl);
+
+                    internalCallbacks.onFail(OpenLayers.i18n("timeoutError"), fileUrl);
                     deleteCookie(settings.cookieFailedName, settings.cookiePath, settings.cookieDomain);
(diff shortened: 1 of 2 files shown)


akashisama:
Sadly the message isn't accurate. Adding extra filtering can actually make the problem worse. To reproduce:
(make sure you use a browser set to ask where to save files so it doesn't try to actually download the file)
1. Add the layer "MARVL3 - Australian shelf temperature data atlas"
1. Add a BBOX around the whole of Australia
1. Go to step 3
1. Click "download as" - > Temperature Observations csv
1. The download dialogue comes up instantly (click cancel once you get there, it's a big download)
1. Go back to step two and add a BBOX around Tasmania.
1. Go to step 3 
1. Click "download as" - > Temperature Observations csv
1. The download request times out!

pmbohm:
So this indicates a solution needs to occur much further up the food chain. 
https://github.com/aodn/issues/issues/355 is the parent issue.

Do all our WFS download requests need to be asynchronous WPS requests now due to increased data size of our data sets?

akashisama:
I've altered the error message to provide as much detail as we can really confirm about what went wrong.

PR #2800, opened 2019-07-23
Depth service layer config - now world_depth_data

pmbohm:


pmbohm:
For aodn/backlog#1046

PR #2811, opened 2019-08-26
report is written to a temp file as it is built instead of a big string in memory

craigrose:
For https://github.com/aodn/issues/issues/488

### DIFF
--- a/grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
+++ b/grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
@@ -92,7 +92,7 @@ class BulkDownloadService {
             log.debug "Added $bytesCopied Bytes"
 
             if (!isReportFile) { report.addSuccessfulFileEntry url, filenameToUse, bytesCopied }
-        } catch (Throwable e) {
+        } catch (Exception e) {
 
             log.warn "Error adding file to download archive. URL: '$url'"
             log.debug "Caused by:", e
--- a/grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
+++ b/grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
@@ -72,9 +72,9 @@ class BulkDownloadService {
         def filenameToUse
         def isReportFile = false
 
-        if (url == report.ZIP_FILE_NAME) {
+        if (url == report.ZIPPED_REPORT_FILENAME) {
             url = report.tempFile.toURI()
-            filenameToUse = report.ZIP_FILE_NAME
+            filenameToUse = report.ZIPPED_REPORT_FILENAME
             isReportFile = true
         } else {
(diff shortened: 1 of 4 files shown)


src/groovy/au/org/emii/portal/DownloadReport.groovy
```diff
@@ -6,9 +6,12 @@ import java.text.DateFormat
 
 class DownloadReport {
 
+    static final def REPORT_TEMP_FILE_SUFFIX = "download_report"
+    static final def ZIP_FILE_NAME = "download_report.txt"
```
akashisama: this variable name seems confusing to me. Once you understand the code it's clear it means "filename to give the report file in zip" but initially it seems like it must mean "name of the zip file"

test/unit/au/org/emii/portal/DownloadReportTests.groovy
```diff
@@ -61,37 +61,21 @@ class DownloadReportTests extends GrailsUnitTestCase {
...
+        def entry =  """\
             --[ #1 ]------------------------------------
             URL:                 url
             Filename in archive: filename
             Result:              went well
-            """,
-            report.reportBody
-    }
-
-    void testGetText() {
```
akashisama: how come you removed this test?

grails-app/services/au/org/emii/portal/BulkDownloadService.groovy
```diff
@@ -81,9 +91,8 @@ class BulkDownloadService {
 
             log.debug "Added $bytesCopied Bytes"
 
-            report.addSuccessfulFileEntry url, filenameToUse, bytesCopied
-        }
-        catch (Exception e) {
+            if (!isReportFile) { report.addSuccessfulFileEntry url, filenameToUse, bytesCopied }
+        } catch (Throwable e) {
```
akashisama: I'm not sure catching Throwable is justified here. Presumably this is to catch OutOfMemoryError but if that happens isn't the whole webapp going to be basically non-functional? I think we should allow the error to propagate to the top level asap rather than catching it here and treating it the same as if it had just been a faulty internet connection.

craigrose: No that catch is to catch failures when adding files to the zip so the correct entry can be generated for the report.  I changed it from Exception to Throwable because Throwable is the base class and hence will catch all problems.

test/unit/au/org/emii/portal/DownloadReportTests.groovy
craigrose: This was a test for the text getter that no longer exists.  We now get a file instead.

craigrose: I've added a test for getTempFile - tests that the file exists and is readable

PR #2815, opened 2019-09-10
prototype portal integration with geonetwork 3 #1259

sachitrajbhandari:
https://github.com/aodn/backlog/issues/1259

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -86,19 +86,23 @@ featureToggles {
 enabledFacets = [
     [
         name: 'parameterFilter',
+//        key: 'Measured parameter'
         key: 'parameterCategories'
     ],
     [
         name: 'organisationFilter',
+//        key: 'Organisation',
         key: 'orgUnitCategories',
         collapsedByDefault: false
     ],
     [
         name: 'platformFilter',
+//        key: 'Platform'
         key: 'platformCategories'
     ],
     [
         name: 'temporalResolutionFilter',
+//        key: 'Temporal Resolution',
         key: 'temporalResolutionCategories',
         collapsedByDefault: true
     ],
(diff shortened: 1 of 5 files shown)


jonescc:
Looks good.  As suggested by Alex on Friday, ideally we would allow GeoNetwork 3 behaviour to be set in configuration to make it easy to deploy (we wouldn't have to coordinate cloud-deploy/config changes to get this included).   This would also make it possible to merge this now without breaking GeoNetwork 2 support we are currently using.

grails-app/conf/Config.groovy
```diff
@@ -144,6 +148,7 @@ environments {
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "http://${localhostAddress}:8883/geonetwork"
         gogoduck.filenamePrepend = "IMOS_aggregation"
+        geonetwork.version = 3
```
ccmoloney: Do we want to be a bit more specific with the version number in case we have to cater for changes in a subsequent 3.x version?

web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -221,12 +236,20 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
...
         var label = this._buildLabel("fa-tags", OpenLayers.i18n('searchPlatformText'));
-        var broaderPlatforms = this._getBroaderTerms(platforms, 1,'platformCategories');
+
+        var termType = ''
+        if (Portal.app.appConfig.geonetwork.version === 3) {
+            termType = 'platformCategories'
+        } else {
+            termType = 'Platform'
+        }
+
```
ccmoloney: Could we abstract these if(version === x) { something = something } type bits into a CatalogueTermLookup class or something?  It could have getPlatformCategoryTerm(version), getOrganisationTerm(version), getParameterCategoryTerm(version) methods

PR #2816, opened 2019-09-12
Inject jenkins UID as BUILDER_UID

anguss00:
#https://github.com/aodn/issues/issues/511
Requires https://github.com/aodn/jenkins-shared-library/pull/1

To be merged after all Jenkinsfiles have been updated https://github.com/aodn/cloud-deploy/pull/512

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,12 +1,11 @@
 pipeline {
     agent none
-
     stages {
         stage('container') {
             agent {
                 dockerfile {
                     args '-v ${HOME}/.m2:/home/builder/.m2 -v ${HOME}/.grails:/home/builder/.grails'
-                    additionalBuildArgs '--build-arg BUILDER_UID=${JENKINS_UID:-9999}'
+                    additionalBuildArgs '--build-arg BUILDER_UID=$(id -u)'
                 }
             }
             stages {


Jenkinsfile
```diff
@@ -1,12 +1,14 @@
 pipeline {
     agent none
-
+    environment {
+        UID = buildHelper.getUid()
```
lwgordonimos: This method won’t work with agents, since there’s no point getting the master UID when the container is then run on an agent. It *must* be the uid where the container is run no matter what.

anguss00: I have no idea how to solve this then

lwgordonimos: Actually, It might be ok with the templates we’re using, since the uid is fixed at 498, so assuming that is consistent this would be fine , but maybe with a todo to note that it wouldn’t work outside the specific templates.

anguss00: All good, @akashisama found the solution, just putting up now

lwgordonimos: Actually it might not need to be dynamic... that templates are fixed at 498, so we might get away with setting that as the default value rather than trying to calculate it.

anguss00: You can run ``additionalBuildArgs '--build-arg BUILDER_UID=$(id -u)'`` and it will dynamically set as appropriate

PR #2820, opened 2019-10-10
Fix jenkins git credentials for release

akashisama:


anguss00:
It may need the git config for user and email set in bumpversion to avoid that first-run config error we had

PR #2837, opened 2020-01-29
Implement category based logo assignment mechanism

APaulic:
Implements a logo assignment mechanism that is based on the use of categories to determine which logo will be associated with a given fascet.

Backlog Issue: https://github.com/aodn/backlog/issues/1464

### DIFF
--- a/web-app/js/portal/search/FacetedSearchResultsDataView.js
+++ b/web-app/js/portal/search/FacetedSearchResultsDataView.js
@@ -61,7 +61,7 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
                     return (this.isRecActive(values.uuid)) ? OpenLayers.i18n('collectionExistsMsg') : OpenLayers.i18n("addDataCollectionMsg");
                 },
                 getIconUrl: function(values) {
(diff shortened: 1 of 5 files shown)


web-app/js/portal/search/FacetedSearchResultsDataView.js
```diff
@@ -61,7 +61,7 @@ Portal.search.FacetedSearchResultsDataView = Ext.extend(Ext.DataView, {
                     return (this.isRecActive(values.uuid)) ? OpenLayers.i18n('collectionExistsMsg') : OpenLayers.i18n("addDataCollectionMsg");
                 },
                 getIconUrl: function(values) {
-                    return Ext.ux.Ajax.constructProxyUrl(this.getGeonetworkImageUrl(values.iconSourceUuid));
```
pmbohm: I would keep the proxy. It means we keep everything coming from https.

pmbohm:
@APaulic Apart from the proxy issue, it looks good

web-app/js/portal/search/FacetedSearchResultsDataView.js
APaulic: Thanks @pmbohm, I added the imageBucket URL to allowed hosts and returned the use of a proxy. Tested and working from my end :+1:

PR #2839, opened 2020-02-05
Move geonetwork image bucket config to global

pmbohm:


lwgordonimos:
For reference, if `geonetwork.imageBucket` is not explicitly set, this following line causes an exception which prevents the subsequent allowed hosted from being configured:

https://github.com/aodn/aodn-portal/commit/bb6a2f1d73bcf91347abc5da524f20edd776af99#diff-8830ff4bbb6846531e5e5d21b14004ceR47

We noticed this because systest wasn't allowing the base layers through the proxy.

PR #2840, opened 2020-02-06
Use data layer for WFS getFilters

pmbohm:


pmbohm:
Required for https://github.com/aodn/backlog/issues/1287

PR #2854, opened 2020-05-05
Fixes issues#2850 by making geoserver-static calls https

pmbohm:


pmbohm:
Fixes aodn/aodn-portal/issues/2850

PR #2865, opened 2020-05-26
Infinite scroll

pmbohm:
For https://github.com/aodn/backlog/issues/38


jonescc:
Some anomalies with display of extents in step 1?  Not sure if this is caused by these changes 

image

pmbohm:
A fix (for a very old sleepr bug)  is now incorporated into the branch

web-app/js/portal/search/SearchBodyPanel.js
```diff
@@ -31,33 +32,100 @@ Portal.search.SearchBodyPanel = Ext.extend(Ext.Panel, {
...
+            });
+        }, this);
     },
 
     _onResultsStoreLoad: function() {
         if (this.resultsStore.getTotalCount() == 0) {
             this._displayNoResultsAlert();
         }
+        var that = this;
+        setTimeout(function () {
```
jonescc, line 48: If I'm using the scroll bar I can quite easily beat this timeout and then can't scroll anymore (until I go backwards a bit after 2 seconds have elapsed that is).  timeouts are bad

pmbohm: I struggled without a timeout for quite a while. The issue is when the mouse is scrolled moderately fast, the mouse proximity to the bottom is within the trigger range even when new results load. So three or four paged requests can be made one after another before the mouse position is out of the trigger range.
I reason that someone racing through the results is not really using the Portal properly. Additionally scanning/ scrolling though the results  is so boring that filtering would become attractive.
The timeout could be reduced but the requests to Geoserver were taking a bit over a second at worst, so i settled on 2 seconds.

jonescc: Happy to merge if @kereid / @ccmoloney are OK with this. It works fine otherwise.

ccmoloney: I'm ok with it.

PR #2872, opened 2020-06-19
Previous minimaps now cleaned up

pmbohm:


pmbohm:
This was a bug I think i laid in wait years ago although Blame shows 'Burgess'

PR #2891, opened 2021-04-28
Share link optionally uses the 'q' search service for GN3

craigrose:


grails-app/conf/Config.groovy
```diff
@@ -167,6 +167,7 @@ environments {
         gogoduck.url = "http://${localhostAddress}:8300/go-go-duck"
         geonetwork.url = "https://catalogue-imos.aodn.org.au/geonetwork"
         geonetwork.version = 2
+        geonetwork.searchService = "xml.search.summary"
```
jonescc, line 170: Version and url don't match -  catalogue-imos is version 3 now

grails-app/conf/Config.groovy
```diff
@@ -167,6 +167,7 @@ environments {
...
         gogoduck.filenamePrepend = "IMOS_aggregation"
 
         if (geonetwork.version == 3) {
```
jonescc, line 173: I'd just set the search service according to the version and not require anyt cloud deply config until we work out its necessary.

jonescc: Note this would have to go at the end of the file after reading cloud-deploy config

jonescc (approved):

PR #2901, opened 2021-11-04
UI login logout

nspool:
This PR contains this initial version of the User Registration workflow. The feature toggle `cognitoAuthentication` must first be enabled.

- User registration workflow
- Login/logout functionality
- User profile page/form
- Password reset function
- An authentication status widget showing the email of the current user
- Guest Login function so that the user doesn't need to be logged in to download data
- A hook in the Download handler that will display the login modal


web-app/js/aws-cognito/_config.js
```diff
@@ -0,0 +1,8 @@
+window._config = {
+    region: "ap-southeast-2",
+    cognito: {
+      UserPoolId: "ap-southeast-2_HKMh9Piwp",
+      ClientId: "ttspre6jds4fa5sjt979mtjrb", // nsp-cog01
+    },
```
ndhogan, line 6: @nspool - should we use placeholder values for _UserPoolId_ and _ClientId_?

ndhogan (changes requested):

web-app/js/aws-cognito/_config.js
nspool: How about just setting them to null?

ndhogan (dismissed):
:+1:

akashisama:
When running this up via cloudformation with the feature toggle not enabled, I get the following errors in the javascript console:
``` javascript
Uncaught Error: Both UserPoolId and ClientId are required.
    at new t (amazon-cognito-identity.min.js:136)
    at UserAuthentication (UserAuthentication.js:2)
    at Auth.js:1
```
``` javascript
Uncaught Error: Both UserPoolId and ClientId are required.
    at new t (amazon-cognito-identity.min.js:136)
    at UserAuthentication (UserAuthentication.js:2)
    at portal-all.js?v=4.42.67-dev:1
```

``` javascript
Uncaught ReferenceError: Portal is not defined
    at search:106
```

nspool:
It appears `portal-all.js` is still including all of the auth JS files. What is included in this file is not controlled by the `g:if` directives but by the function `CollatePortalJavaScriptSource`. Looks like there is still some work to implement this feature toggle as intented.

ndhogan (approved):

PR #2903, opened 2021-11-10
Implement User Feedback

nspool:
- Add a user delete workflow
- Add confirmation dialog to signout
- Refactored the utils.js into UserAuthentication.js to improve encapsulation and make it easier to port into the next portal
- Move Australia to top of country dropdown
- Update industry options on signup and profile page
- Add Data Use Acknowledgement link
- Allow skipping signup confirmation code and go to directly to download
- Read  AWS Cognito UserPoolID and ClientID from environment variables
- Remove placeholder text


grails-app/views/auth/_changePassword.gsp
```diff
@@ -9,7 +9,7 @@
                 <input autocomplete="off" type="password" title="ERROR" placeholder="Current Password" id="changePasswordCurrent" name="changePasswordCurrent" pattern="^[\S]+.*[\S]+$" minlength="8" maxlength="20" required>
             </div>
             <div>
-                <label for="signUpPassword">New Password (Required)</label><div style="float:right"><small>(8-20 characters, containing lower case, upper case, symbol and number)</small></div>
+                <label for="signUpPassword">New Password (Required)</label><div style="float:right"><small>(At least 8 characters)</small></small></div>
```
ndhogan, line 12: Missed this last time sorry - should probably say ``` for="_changePasswordNew"``` - likewise for _current password_ and _confirm new password_

ndhogan (changes requested):
Just a small bug with the delete function and a couple of suggested changes.

Nice work with the "pendingConfirmation" thing

Not really part of this job, but I noticed that User Profile the disabled view of "I agree to being contacted" has the checkbox below the text (at least in my browser - see below)... I think it should reflect the edit view (if it's a quick fix)?

user registration concept - quickfix_

ndhogan (approved):
Neat solution to the AWS keys

PR #2909, opened 2022-02-03
upgrade docker to latest ubuntu

digorgonzola:
The maven build seems to complete fine! Artifact here: https://build.aodn.org.au/job/aodn-portal_build/job/docker_ubuntu20/lastSuccessfulBuild/artifact/aodn-portal-0.0.0-production.war

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:20.04
 
 ARG BUILDER_UID=9999
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


craigrose:
do we actually want latest?

Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:latest
```
akashisama, line 1: should this one be locked to 20.04 given it's using the system python? Very possible that the locked python library versions won't be compatible with newer version of python

digorgonzola: I'll defer to your better judgment!

akashisama (approved):

PR #2913, opened 2022-02-28
portal in docker with debug

craigrose:
This puts portal in a docker container with port 5005 for attaching a debugger.
Also removed a redundant dependency from pom.
See readme.


craigrose:
@sqbaillie This is ready to review again

sqbaillie (approved):

PR #2922, opened 2022-05-17
Revert "remove gn version switching"

gsatimos:
Reverts aodn/aodn-portal#2921

@craigrose I deployed to edge and I think there may be one small hiccup. The text about the collections don't seem to be displaying. 

image

I don't know if maybe this _has_ to be merged for it to work? https://github.com/aodn/cloud-deploy/pull/1153/files

Thing is that you would think that it would just ignore that param completely anyway so not sure if that's it or not.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -92,6 +92,35 @@ auth {
 enabledFacets = [
     [
         name: 'parameterFilter',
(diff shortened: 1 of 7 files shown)


craigrose:
@gsatimos Thanks. I missed that - it's definitely something I've done in aodn-portal.

PR #2925, opened 2022-06-16
squidle - add sample config

lbesnard:


lbesnard:
@bpasquer I guess we could close this right?

bpasquer:
yes that's been dealt with

PR #2939, opened 2023-02-15
Step 3: Remove NetCDF from unsubsetted download link text

nspool:


evacougnon:
@nspool please DO NOT merge for now

I wasn't aware of the `downloadFileListAction: 'Un-subsetted Non-NetCDFs'` option, which may actually fit the purpose. 
Let me have a chat with @atkinsn , as both options exist there might be a reason behind.

FYI @aodn/data-team

mhidas:
I don't think it's helpful to label the download option based on what it _isn't_. I'm not sure what Portal users would make of "Un-subsetted non-netCDFs", but surely "Un-subsetted files", or better still something like "Original files" is at least as informative?

In terms of Portal functionality, is there any difference between `downloadFileListAction` and `downloadUnsubsettedNetCdfAction`?

vfisaac:
Will close the PR and wait for feedback from Nat

atkinsn:
I think consensus amongst the POs is more relevant in this case, my only contribution would be in regards to this change making sense in situations where there are just NetCDFs (p.s. agree with Marty's comment above, about not labeling what it 'isn't'). 

My only comment would be that in situations where there are NetCDF files, so there are the other download options, would it be confusing for users to be provided with options for both "NetCDF" and "Un-subsetted files" when in these cases they are both NetCDFs, and currently they both say "NetCDF" - see example screenshot. Is it worth creating another protocol for "other file" types??

image

bpasquer:
> Is it worth creating another protocol for "other file" types??

Actually I think @evacougnon was wondering what the protocol `downloadFileListAction` was and if anyone had ever heard of it, cause it seems that it is exactly what we are looking for here.

leonardolaiolo:
>Actually I think @evacougnon was wondering what the protocol `downloadFileListAction` was and if anyone had ever heard of it, cause it seems that it is exactly what we are looking for here.

I cannot remember to have read about it anywhere.

Overall, I like Eva's idea of changing to a more generic `Un-subsetted files`. On the other hand, if we think some users could get confused, we could add complexity to it and add more specific options such as `Un-subsetted CSV files`, although I believe this is not necessary.

evacougnon:
It will be good to be able to use the `downloadFileListAction` protocol, but I can't find where to call it.

@nspool , do you know where/how the `downloadFileListAction` protocol is used? For `downloadUnsubsettedNetCdfAction` we call it using `IMOS:AGGREGATION--bodaac` in geonetwork. How can we access the `downloadFileListAction`?
Once I know how it's used and if I can set the `downloadFileListAction`, then we only have this protocol to update.

nspool:
@evacougnon it appears to be the handler for the `AODN:FILE-list` protocol.

evacougnon:
I tried it yesterday and didn't get what I expected:

Screenshot from 2023-02-16 14-54-14

I expect to see 2 options for download:
- List of URLs
- Un-subsetted Non-NetCDFs

evacougnon:
@nspool I think I got it (part of it maybe!), this protocol doesn't include the list of URL download option. Comparing the corresponding handlers, `web-app/js/portal/cart/FileListDownloadHandler.js` doesn't include what `web-app/js/portal/cart/BodaacDownloadHandler.js` has:
```
downloadOptions.push({
                textKey: 'downloadAsUrlsLabel',
                handler: this._getUrlGeneratorFunction(),
                handlerParams: {
                    downloadLabel: OpenLayers.i18n('downloadUrlListAction'),
                    filenameFormat: '{0}_URLs.txt',
                    downloadControllerArgs: {
                        action: 'urlListForLayer',
                        urlFieldName: this._urlFieldName()
```

